PR #53, opened 2026-08-12
lexer: Add hex, binary, and octal integer literals

### LINKED ISSUE #9: lexer: Add hex, binary, and octal integer literals
## Context

Silk has decimal integer literals only. The lexer scans a run of decimal digits and
never looks at a prefix (`packages/compiler/src/Lexer.ts:322-339`). Bit work, byte
masks, and character codes are therefore hard to read. A user who writes a mask must
convert it to decimal by hand.

## Current behavior

The lexer accepts one or more ASCII digits as a decimal integer literal. It does not
recognize a base prefix. This code does not lex as one number:

```silk
const mask: u32 = 0xff
```

The lexer produces `DecimalInteger(0)` and then the identifier `xff`. The parser then
reports an error, because two expressions sit next to each other.

The specification states the decimal rule only: "A decimal integer literal SHALL contain
one or more ASCII digits" (`openspec/specs/bootstrap-lexer/spec.md:14-15`). It excludes
no other base. This ticket is therefore additive.

The value path already works. Integer literals keep their exact magnitude until a type
selects them (`openspec/specs/bootstrap-integer-scalars/spec.md:39`). A hexadecimal
literal needs a new scan rule only, not a new value rule.

## Requirements

1. The lexer must accept `0x` and `0X` followed by one or more hexadecimal digits.
2. The lexer must accept `0b` and `0B` followed by one or more binary digits.
3. The lexer must accept `0o` and `0O` followed by one or more octal digits.
4. Each prefixed literal must produce the existing `DecimalInteger` token kind.
5. The lexer must keep the exact magnitude of the literal, as it does for decimal.
6. A prefix with no digit after it must produce one lexical diagnostic.
7. A prefixed literal must not accept a fraction part or an exponent part.
8. The change must not add a new token kind.

## Example

```silk
const upperMask: u32 = 0xff00
const lowByte: u8 = 0b1010_0000
const permission: u16 = 0o644

pub fn high(value: u32) -> u32 {
  return value & 0xffff0000
}
```

## Out of scope

- A digit separator inside a literal. Ticket 06 covers the `_` separator.
- A hexadecimal float literal.
- A base prefix for a float literal.
- A source-form record of the base. The token slice keeps the exact bytes already.

## Implementation note

The scan must stay a longest-token scan. The lexer reaches the number branch with the
byte `0`. The branch must look at the next byte before it consumes the digit run. If the
next byte is not a base letter, the branch must fall back to the decimal rule, because
`0` alone is a valid decimal literal and `0.5` is a valid float.

A prefix letter is also an identifier-start byte (`packages/compiler/src/Lexer.ts:24-25`).
The literal `0xff` must therefore become one token, not a number token and an identifier
token. The token kind stays `DecimalInteger`, so the value conversion must read the
prefix from the token slice and select the base. A conversion that assumes base 10 gives
a wrong value in silence.

## Dependencies

None.

## Acceptance criteria

- [ ] The lexer produces one integer token for `0xff`, `0b1010`, and `0o777`.
- [ ] A lexer test shows that `0` and `0.5` keep their current token kinds.
- [ ] A lexer test shows that `0x` alone produces one lexical diagnostic.
- [ ] The value of `0xff` equals the value of `255` in a compiled program.
- [ ] A literal outside the range of its selected type produces the existing diagnostic.
- [ ] `openspec/specs/bootstrap-lexer/spec.md:14-15` gets an amendment for the three
      new base prefixes.

julia-script:
Closes #9

The number branch now looks at the byte after a leading `0` before it consumes a digit
run. `0x`/`0X`, `0b`/`0B`, and `0o`/`0O` each commit to their base and scan only that
base's digits, so `0xff` is one token rather than `0` followed by the identifier `xff`.
The token kind stays `DecimalInteger` and the scan stays a longest-token scan: a leading
`0` that no base letter follows falls back to the decimal rule, and a prefixed literal
never absorbs a fraction or exponent part.

Because the kind is shared, every conversion from a literal token to a value now reads
the base from the token's own slice through a new internal `IntegerLiteral` actor
(`packages/compiler/src/internal/IntegerLiteral.ts`), used by the lexer's digit scan,
`Elaboration.analyzeInteger`, and both `DeclarationIndex` conversions. The previous
`unsignedMagnitude` helper assumed base ten and `BigInt(\`-${digits}\`)` threw on a
negative prefixed constant.

A base prefix with no digit after it becomes one `Invalid` token carrying the new
`LEX0004` lexical diagnostic, which keeps the existing lossless-coverage and
invalid-byte recovery guarantees.

### Acceptance criteria

- [x] The lexer produces one integer token for `0xff`, `0b1010`, and `0o777`.
- [x] A lexer test shows that `0` and `0.5` keep their current token kinds.
- [x] A lexer test shows that `0x` alone produces one lexical diagnostic.
- [x] The value of `0xff` equals the value of `255` in a compiled program.
- [x] A literal outside the range of its selected type produces the existing diagnostic.
- [x] `openspec/specs/bootstrap-lexer/spec.md:14-15` gets an amendment for the three new base prefixes.

The spec amendment restates the decimal sentence at those lines as the unprefixed rule
and adds a `Requirement: Integer literals select a base from their prefix` section with
five scenarios.

Tests: baseline 3 failures, after 3 failures, +11 new tests

`packages/compiler/test/IntegerBaseAcceptance.test.ts` checks three-engine agreement
(MIR evaluation, direct Wasm, native clang) for hexadecimal, binary, and octal literals
in both `const` declarations and expressions. `test/OsFileSystem.test.ts > lowers the OS
directory-list provider runner` timed out once in the full parallel run on a machine
under heavy concurrent load and passes in isolation; it is not counted as a failure.

---

### DIFF
--- a/openspec/specs/bootstrap-lexer/spec.md
+++ b/openspec/specs/bootstrap-lexer/spec.md
@@ -11,9 +11,10 @@ The lexer SHALL recognize ASCII whitespace, `//` line comments, `///` documentat
 distinct token kind, the keywords `pub`, `fn`, `return`, `let`, `move`, and the provisional
 `import`, ASCII identifiers, decimal integer literals, `(`, `)`, `{`, `}`, `:`, `,`, `=`, `.`,
 `-`, `->`, and end-of-file. An identifier SHALL begin with an ASCII letter or underscore and
-continue with ASCII letters, digits, or underscores. A decimal integer literal SHALL contain one
-or more ASCII digits. A `-` immediately followed by `>` SHALL remain one arrow token; any other
-`-` SHALL be one minus token.
+continue with ASCII letters, digits, or underscores. An integer literal written without a base
+prefix SHALL contain one or more ASCII digits and SHALL be read in base ten; the base prefixes are
+specified by "Integer literals select a base from their prefix". A `-` immediately followed by `>`
+SHALL remain one arrow token; any other `-` SHALL be one minus token.
 
 #### Scenario: Lex the first parser fixture
 
@@ -262,3 +263,39 @@ multiline literal.
 
 - **WHEN** a multiline literal body contains `\"\"\"` followed later by an unescaped `"""`
 - **THEN** the escaped quotes remain literal content and the later unescaped triple quote closes the token
+
+### Requirement: Integer literals select a base from their prefix
+
+An integer literal SHALL accept the base prefixes `0x` and `0X` for base sixteen, `0b` and `0B` for
+base two, and `0o` and `0O` for base eight, each followed by one or more digits of its selected
+base. A prefixed literal SHALL be one integer literal token of the same kind as an unprefixed
+literal, SHALL retain its exact magnitude, and SHALL NOT accept a fraction part or an exponent
+part. A leading `0` that no base letter follows SHALL keep the decimal reading, so `0` alone and
+`0.5` retain their existing token kinds. A base prefix that no digit of its base follows SHALL be
+one invalid token producing exactly one lexical diagnostic. Every conversion of an integer literal
+to a value SHALL read the base from the token's own source slice rather than assume base ten.
+
+#### Scenario: Lex every base prefix
+
+- **WHEN** the source bytes spell `0xff 0b1010 0o777`
+- **THEN** the stream contains exactly three integer literal tokens whose slices are `0xff`, `0b1010`, and `0o777`
+
+#### Scenario: Preserve the unprefixed readings
+
+- **WHEN** the source bytes spell `0 0.5`
+- **THEN** the stream contains one integer literal token `0` and one floating-point literal token `0.5`
+
+#### Scenario: Reject a prefix without digits
+
+- **WHEN** the source bytes spell `0x` with no following hexadecimal digit
+- **THEN** the lexer emits one invalid token covering the prefix and exactly one lexical diagnostic
+
+#### Scenario: Keep a prefixed literal free of float parts
+
+- **WHEN** the source bytes spell `0xff.5`
+- **THEN** the integer literal token ends after `0xff` and the remaining bytes are tokenized independently
+
+#### Scenario: Convert a prefixed literal in its own base
+
+- **WHEN** a compiled program returns `0xff` where another returns `255`
+- **THEN** both programs produce the same value, and a prefixed literal outside its selected type's range produces the existing out-of-range diagnostic
--- a/packages/compiler/src/DeclarationIndex.ts
+++ b/packages/compiler/src/DeclarationIndex.ts
@@ -1,6 +1,7 @@
 import * as Option from 'effect/Option'
 import * as Diagnostic from './Diagnostic.js'
 import * as Intrinsic from './Intrinsic.js'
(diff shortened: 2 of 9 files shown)


julia-script:
Orchestrator note (run-s1k9) — verification result and two things reviewers should see that the description does not call out.

**Verified:** `Closes #9` present, all six acceptance criteria ticked, base is `main`, test delta 3 → 3 matches the run baseline (the 3 pre-existing `test/Instances.test.ts` golden-MIR failures). I re-ran `test/OsFileSystem.test.ts` on this branch in isolation under low load: **10/10 pass**, confirming the reported timeout was contention from concurrent agent worktrees on a 4-CPU box (load average peaked at 12–22; that file needs ~25s of test time against a 15s timeout). Not a regression from this change.

**Undocumented behavior change:** `0b"bytes"` previously lexed as `0` followed by a `ByteStringLiteral`; it now lexes as an empty-binary-prefix `LEX0004` diagnostic followed by a `TextLiteral`. Nothing in the repository uses that spelling and the old reading was two adjacent expressions, so no valid program changes — but it is a lexer behavior change worth a reviewer's eye, and it is only mentioned in the agent's internal report, not in the PR body.

**Scope beyond the issue's letter:** this PR also fixes two pre-existing bugs it uncovered while centralizing base selection — `Elaboration.unsignedMagnitude` hardcoded base 10, and `DeclarationIndex.constantLiteral` built `BigInt("-0xff")`, which throws. Both are genuinely required for prefixed literals to work end to end, but they are edits outside the files #9 names.

---

julia-script:
Orchestrator note (run-s1k9) — CI attribution for this PR, and a repo-level finding that applies to every open PR right now.

**Neither red check on this PR is caused by it.** Both are failing on `main` itself:

- **`validate`** — reproduced locally on a clean checkout of `49bfeb4` with no PR changes applied: `@silk-effect/compiler-cli` fails 3 tests in 2 files (`FormatCommand.test.ts > maps read and write failures to exit class two without corrupting source`; `FormatWorkflow.test.ts > reports storage failures, continues in order, and gives exit class two precedence`; `FormatWorkflow.test.ts > reports an atomic commit failure without replacing the original file`).
- **`webcontainer-browser`** — fails on main's own push run (job 94147127235, `head_branch: main`, `event: push`) at step 6, `pnpm exec playwright install --with-deps chromium`:
  ```
  [ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "playwright" not found
  ```
  Playwright is not resolvable at the workspace root, so the step dies in about half a second, before any browser test executes.

**Scope of the problem:** every one of the last 15 pushes to `main` — back to 2026-08-09 — shows `conclusion: failure` for both the CI and Release workflows. `main` has been continuously red for several days, so a green CI check is not currently available as a merge gate for this or any other PR. The usable signal is a local test-delta comparison against the pre-existing failure set, which for this PR is 3 → 3 in `@silk-effect/compiler` (plus the 3 pre-existing `compiler-cli` failures above, untouched).

This is worth fixing independently of the backlog work — the playwright resolution failure in particular looks like a one-line workspace/dependency issue, and it is masking whatever the `webcontainer-browser` job is actually meant to verify.

---

PR #55, opened 2026-08-12
stdlib: Add the namespace field to option, result, and vector

### LINKED ISSUE #33: stdlib: Add the namespace field to option, result, and vector
## Context

The standard library ships 27 modules in `packages/compiler/stdlib/manifest.json`.
Exactly 3 of them have no `namespace` field: `silk/option`, `silk/result`, and
`silk/vector`. No requirement governs which modules declare a namespace, so this is
an oversight. This is the cheapest fix in the standard library.

## Current behavior

The compiler builds the namespace map from the `namespace` field and the aliases, then
removes the undefined values (`packages/compiler/src/Stdlib.ts:38-43`). Name resolution
seeds the module scope from the same list and skips the undefined values
(`packages/compiler/src/NameResolution.ts:160-189`).

A module with no manifest namespace is never auto-injected and never scope-seeded.
Therefore `Vector.make(...)` and `Result.succeed(...)` do not resolve. Every caller must
write the selective import form. All 31 stdlib imports use that form today.

```silk
import silk.vector { Vector, make }

// Vector.make() does not resolve today.
```

## Requirements

1. The manifest must give `silk/option` the namespace `Option`.
2. The manifest must give `silk/result` the namespace `Result`.
3. The manifest must give `silk/vector` the namespace `Vector`.
4. The `silk/option` entry must keep its `None` and `Some` aliases.
5. Qualified use such as `Vector.make()` must resolve with no import statement.
6. The selective import form must continue to work.

## Example

```silk
// After this change, no import statement is necessary.
let numbers = Vector.make<i32>()
let value = Option.some<i32>(1)
```

## Out of scope

- Changes to the name resolution algorithm.
- Changes to any other manifest entry.
- Rewriting the existing stdlib import statements.

## Dependencies

None.

## Acceptance criteria

- [ ] A test shows that `Vector.make()` resolves with no import statement.
- [ ] A test shows that `Option.some(...)` and `Result.succeed(...)` resolve with no import statement.
- [ ] A test shows that `import silk.vector { Vector, make }` still compiles.
- [ ] A check shows that all 27 manifest entries declare a namespace.

## Implementation note

This change unblocks the `Option` and `Result` combinator work. A combinator such as
`Option.map(...)` cannot resolve until `silk/option` has a namespace.

julia-script:
Closes #33

`silk/option`, `silk/result`, and `silk/vector` were the only three shipped modules without a `namespace` field, so they were never auto-injected into user scope and never scope-seeded. `Vector.make(...)` and `Result.succeed(...)` could not resolve without a selective import.

This declares `Option`, `Result`, and `Vector` in `packages/compiler/stdlib/manifest.json` (keeping `silk/option`'s `None` and `Some` aliases) and regenerates `packages/compiler/src/Stdlib.generated.ts`. No name-resolution code changed.

## Acceptance criteria

- [x] A test shows that `Vector.make()` resolves with no import statement.
- [x] A test shows that `Option.some(...)` and `Result.succeed(...)` resolve with no import statement.
- [x] A test shows that `import silk.vector { Vector, make }` still compiles.
- [x] A check shows that all 27 manifest entries declare a namespace.

New tests live in `packages/compiler/test/StdlibNamespaceAcceptance.test.ts` (a no-import program naming all three namespaces, agreeing on the evaluator, direct WebAssembly, and a native clang binary; plus the selective import form on the evaluator and WebAssembly) and in `packages/compiler/test/StdlibResolution.test.ts` (every manifest entry declares a namespace, `Option`/`Result`/`Vector` map to their modules, and `silk/option` keeps its aliases).

Note on the run: the box was shared with concurrent worktrees during testing, and `test/OsFileSystem.test.ts > lowers the OS directory-list provider runner` intermittently exceeded its 15s timeout under load. It passes in isolation (10/10) on an idle machine. `test:native-acceptance` passes 1/1.

---


julia-script:
Orchestrator review (run-s1k9) — independent reviewer pass on this P1 PR, plus CI attribution.

**Verdict: approve, nothing blocking.** Verified mechanically first: `Closes #33`, 4/4 acceptance boxes ticked, base `main`, test delta 3 → 3 against the run baseline. Branch has since been rebased onto `49bfeb4` (clean, no conflicts) so it sits on current `main`.

**Correctness.** The manifest gains exactly the three required `namespace` fields (`manifest.json:101` `Option` with its `None`/`Some` aliases retained, `:117` `Result`, `:158` `Vector`), and all 27 entries now declare one. `Stdlib.generated.ts` matches what `scripts/generate-stdlib.mjs:52-57` would emit — field order, indentation, and quoting all conform, and the three added lines sit in the generator-mandated position, so the file is regenerated rather than hand-edited.

**Collisions** — the risk I specifically asked the reviewer to hunt, since auto-injection changes global scope for every program. No real collision:
- Seeding dedupes by `bindingTarget` (`NameResolution.ts:341-352`), so a selective `import silk.option { Option, Some, None }` resolves to the same target and dedupes instead of raising `bindingConflict`. `None`/`Some` were already seeded before this PR — aliases were never gated on `namespace`.
- No `.silk` file outside the three stdlib modules declares `Option`, `Result`, or `Vector`, and the embedded-Silk fixtures that declare `struct Vector<T>` (`DeclarationIndex.test.ts:966,994,1005,1017`, `Parser.test.ts:381`) never write `Vector.` or import `silk.vector`. Closure loading is demand-driven with no unconditional prelude, so those closures never seed the binding.
- The `Vector.make()` shape is an existing exercised pattern, not a new one: `Layout` is already both a struct and a module namespace and is used that way inside `vector.silk`.

**Worth knowing, not blocking** (pre-existing mechanism this change widens): `Stdlib.requiredModules` matches `\bNamespace\.` against raw source *text*, comments and string literals included. A user file that merely mentions `Result.` in a comment now pulls `silk/result` into the closure, and a project declaring its own top-level `Option`/`Result`/`Vector` would then get a `bindingConflict` rather than shadowing. `Option` and `Result` are much likelier user-chosen names than `Layout` or `RawBuffer`. Nothing in the repo trips it today.

**CI attribution — both red jobs are pre-existing on `main`, not caused by this PR.** I reproduced `validate`'s failure on a clean checkout of `49bfeb4` with no PR changes applied: `@silk-effect/compiler-cli` fails 3 tests in 2 files — `test/FormatCommand.test.ts > maps read and write failures to exit class two without corrupting source`, and `test/FormatWorkflow.test.ts > reports storage failures, continues in order, and gives exit class two precedence` and `> reports an atomic commit failure without replacing the original file`. That is `main`'s state, independent of #33. `webcontainer-browser` is being checked separately.

This also exposed a gap in my own baseline, which I've corrected for later workers: I had baselined only `@silk-effect/compiler`, while CI's `validate` runs every package. The full-repo pre-existing failure set is 3 in `@silk-effect/compiler` (`Instances.test.ts`) **plus** these 3 in `@silk-effect/compiler-cli`.

Two optional nits from the review, neither blocking: `StdlibResolution.test.ts:40-43` asserts no entry lacks a namespace but never pins the count at 27, so an entry silently vanishing from the manifest would still pass; and the qualified test's `settled` drops its argument rather than destructuring the `Result` union the way its selective counterpart does.

---

julia-script:
Closing the loop on the `webcontainer-browser` check I said I was verifying separately: it is **also failing on `main` itself**, not caused by this PR. Main's own push run for `49bfeb4` (job 94147127235) fails at step 6, `pnpm exec playwright install --with-deps chromium`, with `[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "playwright" not found` — the step dies before any browser test runs.

More broadly, the last 15 pushes to `main` (back to 2026-08-09) all show `conclusion: failure` for both the CI and Release workflows, so no green CI baseline exists to gate against right now. Full detail in [this note on #53](https://github.com/julia-script/silk-effect/pull/53#issuecomment-5268292276). Both red checks here are inherited from `main`; this PR's own signal is the 3 → 3 local test delta.

---

julia-script:
⚠️ Orchestrator adversarial review (run-s1k9) — **this PR introduces a regression.** I confirmed it by compiling on both branches, so this supersedes the "non-blocking wart" note in my earlier review comment. A fix is being dispatched to this branch.

## A comment or a string literal now changes name resolution

Program:

```silk
// TODO: replace with Result.succeed once effect rows land
pub struct Result<A, F> { value: A }
```

Measured with `Analysis.ofSourceRealized`, identical source on both branches:

| program | clean `main` (`49bfeb4`) | this branch |
|---|---|---|
| comment mentioning `Result.` + own `struct Result` | `[]` | **`SEM0016`** |
| same, comment deleted | `[]` | `[]` |
| `"Result.succeed"` in a **string literal** + own `struct Result` | `[]` | **`SEM0016`** |

Deleting the comment makes the program compile. So does deleting the string. That is a comment affecting name resolution — and note the string-literal case, which the review that found this did not predict.

**Mechanism:** `Stdlib.ts:53-64` matches `\bResult\.` against **raw source bytes**, comments and string literals included → `ModuleClosure.ts:426-428` pulls `silk/result` into the closure → `NameResolution.ts:160-188` seeds `Result` as a `LocalDeclaration` → the user's own `Result` is pushed at `:190-198` → `:345-360` sees two distinct `bindingTarget`s (`declaration:silk/result.Result` vs `declaration:root.Result`) and raises `SEM0016`.

**Why it is new:** the text-matching mechanism pre-exists for 24 namespaces, but `silk/option`, `silk/result`, and `silk/vector` had no `namespace` field, so they were unreachable this way. This PR makes three of the most likely user-chosen type names in the language reachable.

**Cross-module contagion (worse, same cause):** `Analysis.ts:151-155` builds one union closure across all roots and seeds project-wide, so file A's comment `// Vector.make` can make unrelated file B's `pub struct Vector<T>` conflict — with B containing no trigger text at all. That is the LSP-facing shape of the same bug.

## Fix being applied

Making the closure scan lexical rather than raw-byte, so `Namespace.` occurrences inside comments and string literals no longer pull a module into the closure, plus regression tests for both the comment and string cases.

## A design question for you, deliberately not decided by an agent

Even with lexical matching, a program that *genuinely* writes `Result.succeed(...)` **and** declares its own top-level `Result` will get `SEM0016` rather than having the user's declaration shadow the seeded namespace. Which is intended — conflict, or user-declaration-wins shadowing? Nothing in the repo pins either behavior today, and the answer changes what `NameResolution.ts:345-360` should do for every namespace, not just these three. I am not having a worker invent that rule.

---

julia-script:
✅ Orchestrator round 2 (run-s1k9) — **regression fixed and independently re-verified; adversarial re-review of the fix is CLEAN.**

I re-ran the same three probes from my previous comment against commit `43009b1`:

| program | clean `main` | pre-fix | post-fix |
|---|---|---|---|
| comment mentioning `Result.` + own `struct Result` | `[]` | `SEM0016` | **`[]`** |
| no comment | `[]` | `[]` | `[]` |
| `"Result.succeed"` in a string literal | `[]` | `SEM0016` | **`[]`** |

**The fix is better than what I specified.** Rather than re-lexing or hand-rolling a comment/string scanner, it reuses the token stream that already exists: `requiredModules` is called from `ModuleClosure.loadProject:426`, where the module is parsed and `SyntaxFile` retains trivia-preserving tokens — so they are free at that point, and a benchmarked hot path is not doubled. It keeps the **adjacency** requirement (namespace token immediately followed by `Dot`) to reproduce `\bNamespace\.` exactly, so `Namespace .member` still stays out of the closure, and accepts `DotDot` because the old regex matched `Result..`.

A second adversarial pass targeted the two ways this class of fix fails, and found neither:
- **No closure shrinkage.** The old regex's adjacency requirement was confirmed (`Result .x` never matched), and token adjacency equals byte adjacency because `syntax.tokens` is trivia-preserving. All shipped stdlib sources were re-scanned for comment-only dotted mentions: 4 exist, 3 self-referential, and the one cross-module case (`os_filesystem.silk` → `Path.`) already has an explicit `import silk.filesystem`. No closure changes.
- **No closure growth.** `foo.Result.bar` matched before and after. Every arbitrary-text token kind is skipped — `Token.ts:4-76` has no block comments, raw strings, char literals, or interpolation, so `TextLiteral`/`ByteStringLiteral`/`InvalidStaticLiteral` is the complete set. `Invalid` tokens can never spell a namespace (`Lexer.ts:353-357`).

Also verified: one call site repo-wide, so the CLI, Pipeline, and LSP all route through `loadProject` consistently; no import cycle; O(tokens); and incremental reuse keeps `source` and `tokens` from the same `syntax`, so `spelling` cannot silently return `None`.

**The design question from my previous comment still stands and is now confirmed by the second reviewer:** `NameResolution.ts:160` seeds every closure module's namespace into every other module's scope, so a program that *legitimately* writes `Result.succeed(...)` and also declares its own top-level `Result` still gets `SEM0016` rather than shadowing. That is pre-existing and out of scope for this fix, but this PR makes it reachable for three likely user-chosen names. Conflict or user-declaration-wins shadowing — your call, and it applies to all 24+ namespaces.

PR #61, which stacks on this branch, has been rebased onto the fix.

---

Issue #48, opened 2026-08-12
lsp: Show diagnostic edits as code actions


julia-script:
## Context

The `Diagnostic` model carries an optional `edits` field
(`packages/compiler/src/Diagnostic.ts:639`). Ticket 41 makes the compiler phases
populate that field. After a phase emits an `Edit`, the language server must show it
to the user. The server capability list (`packages/lsp/src/Server.ts:50-65`) has no
`codeActionProvider`. This ticket is protocol wiring only. It adds no compiler logic.

## Current behavior

The server publishes diagnostics with the Error severity. The editor shows the
message and the span. The editor offers no correction. The user reads the message
and then makes the correction by hand.

## Requirements

1. The server must advertise `codeActionProvider` with the kind `quickfix`.
2. The server must map each `Edit` of a diagnostic to one code action.
3. The code action must use the kind `quickfix` and must reference its source diagnostic.
4. The code action title must name the correction, for example "Remove the redundant alias".
5. The server must return the actions in a deterministic order that follows the diagnostic order.
6. The server must return an empty list for a diagnostic that carries no edit.
7. The `WorkspaceEdit` of the action must use the negotiated position encoding.

## Example

```silk
// The compiler reports SEM0013 on this line.
import collections as collections

// The editor offers one quick fix: "Remove the redundant alias".
// The fix applies the Edit and gives:
import collections
```

## Out of scope

- New `Edit` values from the compiler phases. Ticket 41 covers that work.
- Refactor actions such as "extract function" that no diagnostic carries.
- A fix-all command that applies every action in one file.
- Source actions such as "organize imports".

## Dependencies

- Ticket 41. No code action can appear until a phase emits an `Edit`.

## Acceptance criteria

- [ ] A test asserts the initialize response advertises `codeActionProvider`.
- [ ] A test on a source with `SEM0013` returns one quick fix with the correct edit range.
- [ ] A test asserts a diagnostic with no edit returns no code action.
- [ ] A test asserts two diagnostics in one file return their actions in diagnostic order.
- [ ] A new `language-server-code-actions` spec records the requirements.

julia-script:
🔒 Claimed by orchestrator run-b4c2 at 2026-08-12T19:05:24Z. Lane: lsp. Worker dispatching.

julia-script:
PR opened: https://github.com/julia-script/silk-effect/pull/78

PR #56, opened 2026-08-12
diagnostics: Emit machine-applicable edits for unambiguous errors

### LINKED ISSUE #45: diagnostics: Emit machine-applicable edits for unambiguous errors
## Context

The diagnostic model already carries edits. A diagnostic "MAY additionally carry
labeled related spans, notes, and unambiguous machine-applicable edits"
(`openspec/specs/bootstrap-diagnostics/spec.md:11-12`). The `Edit` type exists at
`packages/compiler/src/Diagnostic.ts:639`, and its comment says "Modeled; no phase
emits one yet". No spec amendment is necessary. The phases must populate the field.

## Current behavior

The compiler has 117 stable diagnostic codes. Every `Diagnostic` value carries an
optional `edits` field (`packages/compiler/src/Diagnostic.ts:669`). No constructor in
`Diagnostic.ts` sets that field. A user reads the message and then makes the
correction by hand.

```silk
import math as math   // SEM0013: Alias math does not change the name
```

## Requirements

1. The `redundantAlias` constructor must accept the span of the `as` clause.
2. The `SEM0013` diagnostic must carry one `Edit` that replaces the `as` clause with an empty string.
3. The `MOD0003` diagnostic must carry one `Edit` that removes the repeated import line.
4. The `SEM0048` diagnostic must carry no edit, because the correct new name is ambiguous.
5. Each edit must apply to one file and must not overlap another edit of the same diagnostic.
6. The compiler must produce byte-identical edits for the same source on every run.
7. A test must apply each edit to the source and must then compile the result without the original diagnostic.

## Example

```silk
// Before. The alias repeats the module name.
import collections as collections

// SEM0013 carries this edit:
//   span:        the ` as collections` bytes
//   replacement: ""

// After the edit applies:
import collections
```

## Out of scope

- Edits for the 93 `SEM` codes that need a type decision or a name choice.
- A command-line flag such as `silk check --fix`.
- The language server protocol wiring. Ticket 44 covers that work.
- A new severity level. Every diagnostic keeps `severity: 'error'`.

## Dependencies

- None.

## Acceptance criteria

- [ ] `Diagnostic.redundantAlias` emits one `Edit` and a test verifies the span and the replacement text.
- [ ] `Diagnostic.duplicateImport` emits one `Edit` and a test verifies the span and the replacement text.
- [ ] A test applies each emitted edit and then compiles the corrected source with no diagnostic.
- [ ] A test runs the same source two times and compares the edits for byte equality.
- [ ] The callers at `packages/compiler/src/NameResolution.ts:215`, `:248`, and `:279` pass the spans the edits need.

julia-script:
Closes #45

`Diagnostic.Edit` was modelled but never populated. Two codes whose correction needs no name choice and no type decision now carry one deletion edit each, and the ambiguous code keeps carrying none.

- `SEM0013` (redundant alias) carries an edit deleting the ` as name` clause, for both the namespace form (`import a.B as B`) and the member form (`import a.B { one as one }`).
- `MOD0003` (duplicate import) carries an edit deleting the repeated import line whole, including its terminating newline.
- `SEM0048` (pattern binding conflict) carries no edit, because the correct new name is ambiguous.

Both spans are computed in `NameResolution.ts` from byte offsets alone, so repeated compilations of the same source produce byte-identical edits. No LSP wiring is included — that is #48's scope.

### Acceptance criteria

- [x] `Diagnostic.redundantAlias` emits one `Edit` and a test verifies the span and the replacement text.
- [x] `Diagnostic.duplicateImport` emits one `Edit` and a test verifies the span and the replacement text.
- [x] A test applies each emitted edit and then compiles the corrected source with no diagnostic.
- [x] A test runs the same source two times and compares the edits for byte equality.
- [x] The callers at `packages/compiler/src/NameResolution.ts:215`, `:248`, and `:279` pass the spans the edits need.

Tests: baseline 3 failures, after 3 failures, +5 new tests

---

### DIFF
--- a/packages/compiler/src/Diagnostic.ts
+++ b/packages/compiler/src/Diagnostic.ts
@@ -636,7 +636,16 @@ export interface RelatedSpan {
   readonly span: SourceSpan.SourceSpan
 }
 
-/** One unambiguous machine-applicable replacement. Modeled; no phase emits one yet. */
+/**
+ * One unambiguous machine-applicable replacement: the corrected bytes for one source range.
+ *
+ * A phase emits an edit only where the correction needs no name choice and no type decision from
+ * the author; every other diagnostic carries none. Applying every edit of one diagnostic to its
+ * source removes that diagnostic. All edits of one diagnostic address the source that owns the
+ * diagnostic's primary span, and no two of them overlap, so a consumer may apply them in any
+ * order. Every edit is derived from byte offsets alone, so repeated compilations of the same
+ * source produce byte-identical edits.
+ */
 export interface Edit {
   readonly span: SourceSpan.SourceSpan
   readonly replacement: string
@@ -942,7 +951,16 @@ export const reservedModuleIdentity = (module: string, span: SourceSpan.SourceSp
     span,
   })
 
-export const duplicateImport = (module: string, span: SourceSpan.SourceSpan): Diagnostic =>
+/**
+ * Creates the diagnostic for one module imported more than once.
+ *
+ * `removal` covers the repeated import line, so the emitted edit deletes that line whole.
+ */
+export const duplicateImport = (
+  module: string,
+  removal: SourceSpan.SourceSpan,
+  span: SourceSpan.SourceSpan,
+): Diagnostic =>
   Object.freeze({
     _tag: 'Diagnostic',
     phase: 'module',
@@ -951,9 +969,20 @@ export const duplicateImport = (module: string, span: SourceSpan.SourceSpan): Di
     message: `Module ${module} is imported more than once`,
     reason: Object.freeze({ _tag: 'DuplicateImport', module }),
     span,
+    edits: Object.freeze([Object.freeze({ span: removal, replacement: '' })]),
   })
 
-export const redundantAlias = (spelling: string, span: SourceSpan.SourceSpan): Diagnostic =>
+/**
+ * Creates the diagnostic for an alias that repeats the name it renames.
+ *
+ * `clause` covers the ` as name` bytes, so the emitted edit deletes the clause and leaves the
+ * imported name it duplicated.
+ */
+export const redundantAlias = (
+  spelling: string,
+  clause: SourceSpan.SourceSpan,
+  span: SourceSpan.SourceSpan,
+): Diagnostic =>
   Object.freeze({
     _tag: 'Diagnostic',
     phase: 'semantic',
@@ -962,6 +991,7 @@ export const redundantAlias = (spelling: string, span: SourceSpan.SourceSpan): D
     message: `Alias ${spelling} does not change the name`,
     reason: Object.freeze({ _tag: 'RedundantAlias', spelling }),
     span,
+    edits: Object.freeze([Object.freeze({ span: clause, replacement: '' })]),
   })
 
 export const unknownImportedMember = (
--- a/packages/compiler/src/NameResolution.ts
+++ b/packages/compiler/src/NameResolution.ts
@@ -4,6 +4,7 @@ import * as Diagnostic from './Diagnostic.js'
 import * as Intrinsic from './Intrinsic.js'
 import type * as ModuleClosure from './ModuleClosure.js'
 import * as SourceFile from './SourceFile.js'
+import * as SourceSpan from './SourceSpan.js'
 import * as Stdlib from './Stdlib.js'
 import * as SyntaxTree from './SyntaxTree.js'
 import type * as Token from './Token.js'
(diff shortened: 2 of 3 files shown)


julia-script:
Orchestrator review (run-s1k9) — P1 reviewer pass. **One blocking correctness bug, confirmed by running it.** Everything else is clean; a fix is being dispatched to this branch.

## Blocking: `MOD0003` edit can corrupt source

`NameResolution.ts:141` consumes the trailing newline unconditionally:

```ts
let end = declaration.span.end
while (end < source.bytes.length && isBlank(source.bytes[end])) end += 1
if (source.bytes[end] === lineFeed) end += 1
```

The leftward scan at `:138` only walks over blanks (space/tab/CR), so it stops at the end of a *preceding declaration on the same line* rather than at a line start — but the newline is then eaten anyway, splicing the following line onto the remainder. Silk's grammar has no statement terminator, so two top-level declarations on one line parse fine and this is reachable.

I verified it against the built branch rather than reasoning from the source. Input:

```
import library.One import library.One as Again
```

Result:

```
deleted:   " import library.One as Again\n"
corrected: "import library.Onepub fn main() -> i32 { return 0 }"
after fix: ["MOD0001"]
```

The corrected source is not merely ugly — `import library.One` and `pub fn main` are glued into one token run, and recompiling reports a *new* diagnostic. That contradicts acceptance criterion 3 ("applies each emitted edit and then compiles the corrected source with no diagnostic") for this input, and requirement 7 of the issue. CR-only line endings join lines the same way, for the same reason.

**Fix:** take the trailing newline only when the deletion actually begins a line.

```ts
if ((start === 0 || source.bytes[start - 1] === lineFeed) && source.bytes[end] === lineFeed) end += 1
```

with a regression test covering a duplicate import sharing its line with a preceding declaration.

## Non-blocking

- **Orphaned trailing comment.** `import library.One as Again // dup\n` stops the rightward scan at `/`, leaving `// dup\n` on its own line. Valid source and the diagnostic clears, but it isn't "the line whole" as the doc comment claims.
- The test helpers `editedText`/`applyEdits` index a JS string with *byte* offsets. Fine for the ASCII fixtures, but not a general edit applier — worth a comment so nobody reuses them on non-ASCII source.

## Verified clean

- **`SEM0013` spans are correct.** Start is the `AsKeyword` walked left over blanks, end is `alias.span.end`; spans are half-open and `ImportAlias`'s last element is the alias identifier token, so no trailing trivia, `}`, or `,` is ever swallowed. `isBlank` excludes `,` and `{`, so the member form `import a.B { one as one, two }` is safe. Multiple spaces, a tab before `as`, and a newline before `as` all behave.
- **Edge cases that are fine:** CRLF (the `\r` is a blank, the leftward scan halts on the preceding `\n`); last line with no trailing newline (`bytes[end]` is `undefined`, nothing extra taken); first line of file (a MOD0003 is never the first occurrence).
- **Determinism holds.** Every offset comes from token spans plus byte scans; `seenTargets` is only membership-tested and `module.imports` is an ordered array, so nothing depends on map/set iteration order.
- **Call sites complete.** `duplicateImport`/`redundantAlias` appear only in `Diagnostic.ts` and the three updated sites; no other package, test, or fixture references `MOD0003`/`SEM0013`.
- **Scope is right.** Three files, no LSP wiring (`packages/lsp/src/Document.ts:88` still reads only `relatedSpans`), and `SEM0048` still emits no edit.

Acceptance criteria 1, 2, 4, and 5 are genuinely established by the new tests. Criterion 3 holds for the three cases tested and fails only for the shared-line input above.

---

julia-script:
Orchestrator follow-up (run-s1k9) — **blocking bug fixed and independently verified.** Pushed as `812b2f7` on this branch; no force-push, so the history above is intact.

I re-ran the exact reproduction from my previous comment against the updated branch:

```
input:     import library.One import library.One as Again

deleted:   " import library.One as Again"          ← newline no longer consumed
corrected: "import library.One\npub fn main() -> i32 { return 0 }"
after fix: []                                       ← recompiles with no diagnostics
```

Acceptance criterion 3 now holds for the shared-line input that previously produced `["MOD0001"]`.

**The fix went one step deeper than the guard I proposed, correctly.** My suggested change only gated the trailing-newline consumption. The worker found that the leftward scan was independently unsafe: `isBlank` includes CR, so in a CR-only source the scan swallowed a preceding lone `\r` and glued lines together *even without* the newline consumption — my one-line guard would not have covered that case. The scans in `importLineSpan` now use a narrower `isHorizontalBlank` (space/tab only) so neither can cross a line terminator, and trailing-break consumption handles CRLF, LF, and lone CR explicitly. `isBlank` is untouched and still used by `aliasClauseSpan`, where CR-skipping is harmless.

**Regression test:** "keeps the following line intact for a repeated import that shares a line", added next to the existing MOD0003 edit test. The worker verified it fails against the pre-fix source (stashing the `src` change and running it alone) and passes after — so it genuinely pins the bug rather than merely passing.

**Verification:** `NameResolution.test.ts` 15/15; full suite run twice end to end, both times exactly the 3 pre-existing `Instances.test.ts` host-triple golden mismatches and nothing else (1008 passed); `pnpm typecheck` and `biome check` clean.

Net for this PR: baseline 3 / after 3 / **+6 new tests** (5 original, 1 regression). Remaining non-blocking nits from the review stand as noted and are fine to leave: the orphaned trailing comment on `import ... // dup`, and the byte-offset-indexed test helpers.

---

julia-script:
✅ Orchestrator round 2 (run-s1k9) — **overlap hardening landed (`9918f42`) and independently verified.**

I re-ran the exact case from the finding — three imports, two sharing a line:

```
codes  = ["MOD0003","MOD0003"]
spans  = [19,37) [37,56)      ← edge-to-edge, overlap=false
result = "import library.One\n\npub fn main() -> i32 { return 0 }"
after  = []                    ← recompiles clean
```

Previously these were `[19,38)` and `[37,56)`, sharing byte 37 and corrupting when applied together.

**Approach:** the rightward blank scan now runs into a probe and commits only when nothing but a line terminator or EOF follows the blanks. Blanks separating a duplicate import from a neighbouring declaration are left to that neighbour, which already claims them leftward — so the spans meet edge-to-edge by construction rather than by arithmetic luck. The line-break consumption is now explicitly guarded by `ownsLine && endsLine`.

**One expected shape worth knowing:** for two duplicate imports on one line, applying both edits leaves an **empty line** behind. Neither span may take the break — the first shares its line with the second, and the second taking it would splice the remainder onto `pub fn main`, which is exactly the regression `812b2f7` fixed. The result is valid and recompiles with zero diagnostics, and the new tests assert that shape rather than glossing it.

Both required tests were verified failing pre-fix (added before the `src` edit and run alone): test 1 reported `[19,38)` and `[37,56)` overlapping — precisely the spans in the finding — and the mixed-file test reported `[52,71)` and `[70,89)`. A third test locks in that trailing blanks are still absorbed when they *do* run to the line break, guarding the scan that changed.

Full suite: 1011 passed, 3 failed — the pre-existing `Instances.test.ts` host-triple goldens only. `NameResolution.test.ts` 18/18.

This PR has now closed two review loops: the shared-line newline bug (`812b2f7`) and this cross-diagnostic overlap (`9918f42`). Both were found by review rather than by tests, and both now have regression tests that fail against the pre-fix code.

---

PR #57, opened 2026-08-12
effects: add Effect.flatten to remove one Effect layer

### LINKED ISSUE #23: effects: add Effect.flatten to remove one Effect layer
## Context

`stdlib/silk/effects.silk` holds the Effect combinator suite. `flatMap` is there and it
unions both rows: `! E | F ? R | S`. There is no `flatten`, so a caller who holds an
`Effect<Effect<A>>` must write `flatMap` with an identity function each time. The spec
requires combinators to "resolve to canonical ordinary Silk declarations"
(`bootstrap-flow-functions:246-248`), so `flatten` needs no compiler work.

## Current behavior

A caller writes the identity function by hand:

```silk
let flattened = Effect.flatMap(move nested, identity)
```

The closed combinator list in `bootstrap-flow-functions:246-248` names `map`, `mapError`,
`mapBoth`, `flatMap`, `tap`, `catch`, `retry`, `provide`, `provideMut`, and `provideWith`.
It omits `flatten`, but nothing rejects it.

## Requirements

1. `Effect.flatten` must be an ordinary Silk declaration in `stdlib/silk/effects.silk`.
2. `Effect.flatten` must remove exactly one Effect layer.
3. `Effect.flatten` must union the failure rows of both layers.
4. `Effect.flatten` must union the requirement rows of both layers.
5. The compiler must not add an intrinsic or a dedicated HIR operation for it.

## Example

```silk
/// Removes one Effect layer.
pub effect fn flatten<A, !E, !F, ?R, ?S>(
  self: once Effect<Effect<A ! F ? S> ! E ? R>
) -> A ! E | F ? R | S {
  let inner = run self
  return run inner
}
```

## Out of scope

- `flatMap` changes. It stays as it is.
- Recursion depth safety. See ticket 20 (`Effect.suspend`).

## Dependencies

- None.

## Acceptance criteria

- [ ] An `EffectRuntime` test runs a flattened Effect and gets the inner success value.
- [ ] A test asserts the result type unions both failure rows and both requirement rows.
- [ ] A test asserts `flatten` compiles through the ordinary declaration path with no
      intrinsic.

julia-script:
Closes #23

A caller holding an `Effect<Effect<...>>` had to write `flatMap` with a hand-written identity function each time. `Effect.flatten` removes exactly one layer:

```silk
/// Removes one Effect layer from a nested Effect.
pub effect fn flatten<A, !E, !F, ?R, ?S>(
  self: once Effect<Effect<A ! F ? S> ! E ? R>
) -> A ! E | F ? R | S {
  let inner = run self
  return run inner
}
```

It is an ordinary Silk declaration in `stdlib/silk/effects.silk`, placed after `flatMap`, and unions both the failure rows and the requirement rows of the two layers. No intrinsic and no dedicated HIR operation were added — `bootstrap-flow-functions:246-248` requires combinators to resolve to canonical ordinary Silk declarations, and the closed list there omits `flatten` without rejecting it. `Stdlib.generated.ts` is regenerated from the manifest.

### Acceptance criteria

- [x] An `EffectRuntime` test runs a flattened Effect and gets the inner success value — `flattens one nested Effect layer across evaluator, LLVM, and Wasm` compiles a nested Effect and asserts the inner value `42` on all three engines, with the Wasm export and the native LLVM IR both checked.
- [x] A test asserts the result type unions both failure rows and both requirement rows — `unions both failure rows and both requirement rows through flatten` pins the encoded row set.
- [x] A test asserts `flatten` compiles through the ordinary declaration path with no intrinsic — `resolves flatten through the ordinary declaration path without an intrinsic` resolves the occurrence to `silk/effects`, asserts the constructed intrinsic set is exactly `['Intrinsic.i32Multiply']`, and asserts the intrinsic catalog does not include `flatten`.

Verified after rebasing onto `49bfeb4`: `Test Files 1 failed | 128 passed (129)` / `Tests 3 failed | 1005 passed (1008)` — the same three pre-existing `test/Instances.test.ts` golden-MIR failures, and 1008 total against the baseline's 1005.

Note on CI: the `validate` and `webcontainer-browser` checks are red on `main` itself and have been for the last 15 pushes, so both will show red here for reasons unrelated to this change. Details in [this note on #53](https://github.com/julia-script/silk-effect/pull/53#issuecomment-5268292276).

---


julia-script:
Orchestrator adversarial review (run-s1k9) — **no blocking defect; no change requested to this PR.**

Verified clean: `Stdlib.generated.ts` is a byte-exact regeneration; row dedup and canonical sort (`Type.ts:364-400`) make a duplicate `E`/`F` collapse safely; `E ⊆ E|F` and `R ⊆ R|S` propagation uses the same mechanism `flatMap` already relies on; `inner` is never bound when the outer Effect fails, so nothing leaks or double-drops; and generic inference works without annotations in both direct and piped call forms.

One finding, **filed separately as #64 rather than fixed here**: `run self` on a `once` *parameter* is not tracked as a consuming use. `Ownership.ts:637` gates the consuming `Run` branch on `subject._tag === 'BindingReference'`, but parameters lower to `ParameterReference` (`Ownership.ts:314`), so the affine value is never marked dead. It is benign today only because `cleanupPlan` returns `NoCleanup` for Effect-typed values (`Ownership.ts:1599`).

This is why the neighbouring combinators (`flatMap`, `tap`, `map`, `mapError`, `catchAll`) consume via `result(move self)` — `move` into a call *is* tracked. `flatten` is the first stdlib declaration on the untracked path.

I deliberately did **not** widen this PR to route around it. The code here is correct as written and matches issue #23's example exactly; the defect is in the ownership checker, and working around it in stdlib source would hide the bug and leave the checker gap in place for everyone else. #64 proposes widening the `Run` case to use `useSite`, which already handles `ParameterReference`.

Two test-quality notes, non-blocking: the intrinsic assertion `[...intrinsics]` scans the user module where a stdlib intrinsic could never appear, so criterion 3 is really carried by the `Intrinsic.all()` catalog check and the `EffectConstruct` target — the weaker assertion just adds noise. And `Effect<once Effect<…>>` will not unify with `flatten` because inner access is pinned to `Shared`; that matches `flatMap`'s callback-result limitation, so it is house-consistent, but it is untested.

---

PR #58, opened 2026-08-12
lexer: Accept _ as a digit separator in number literals

### LINKED ISSUE #10: lexer: Accept _ as a digit separator in number literals
## Context

Silk number literals accept digits only. A long constant is therefore hard to read. A
user who writes a memory limit or a byte count must count the digits to check the
magnitude. Many languages accept `_` inside a literal for this reason.

## Current behavior

The byte `_` is an identifier-start byte (`packages/compiler/src/Lexer.ts:24-25`). The
number branch stops at the first byte that is not a decimal digit
(`packages/compiler/src/Lexer.ts:322-339`). This source:

```silk
const limit: u32 = 1_000
```

lexes as `DecimalInteger(1)` followed by the identifier `_000`. The parser then reports
an error, because two expressions sit next to each other. The user gets no message about
the separator.

## Requirements

1. The lexer must accept `_` between two digits of an integer literal.
2. The lexer must accept `_` between two digits of a float literal.
3. The lexer must reject a literal that ends with `_`.
4. The lexer must reject a literal that contains two `_` bytes in a row.
5. The lexer must reject a `_` directly after a base prefix.
6. The lexer must reject a `_` directly before or after the decimal point.
7. The lexer must reject a `_` directly before or after the exponent letter.
8. Each rejection must produce one lexical diagnostic with the span of the literal.
9. The value of a literal must not change when a separator is present.
10. The change must not add a new token kind.

## Example

```silk
const maximumBytes: usize = 1_048_576
const ratio: f64 = 3.141_592
const flags: u32 = 0b1010_0000_1111_0001
```

## Out of scope

- A separator in an identifier. The `_` byte keeps its identifier meaning.
- A separator inside a text literal or a byte-string literal.
- A separator in any position other than a number literal.
- A group-size rule. The lexer must not require groups of three digits.

## Implementation note

The lexer uses deterministic longest-token recognition
(`openspec/specs/bootstrap-lexer/spec.md:96`). The number branch must therefore consume
the `_` bytes itself. If the branch stops at the `_`, the identifier branch takes the
rest and the two-token result stays.

The trailing-separator rule needs one byte of lookahead. The branch must accept `_` only
when the following byte is a digit of the selected base. This one rule also rejects two
separators in a row, and rejects a separator before the decimal point.

A separator must not end an identifier scan. The spelling `x_1` stays one identifier,
because the identifier branch runs before the number branch for a leading letter.

The value conversion must remove every `_` byte from the token slice before it reads the
number. A conversion that keeps the separator gives a wrong value or an error.

## Dependencies

None. Ticket 05 adds the base prefixes. If ticket 05 lands first, this ticket must cover
the prefixed forms as well.

## Acceptance criteria

- [ ] The lexer produces one integer token for `1_000`.
- [ ] A lexer test shows that the value of `1_000` equals the value of `1000`.
- [ ] A lexer test shows that `1_` produces one lexical diagnostic.
- [ ] A lexer test shows that `1__0` produces one lexical diagnostic.
- [ ] A lexer test shows that `_1` stays one identifier token.
- [ ] A lexer test shows that `1_000.5` and `1.000_5` are float tokens.
- [ ] `openspec/specs/bootstrap-lexer/spec.md:14-15` gets an amendment for the separator.

julia-script:
Closes #10

Stacked on #53 (`agent/9-hex-binary-octal-literals`) — both change the same lexer number scan. Review/merge #53 first.

The number branch now takes its digit run greedily over both base digits and `_` separators. A separator is well placed only with a digit of the same run immediately on each side, which is the single rule that rejects a leading, trailing, doubled, prefix-adjacent, point-adjacent, and exponent-adjacent separator alike. A literal whose separators are not all well placed becomes one `Invalid` token covering the literal with exactly one `LEX0005` diagnostic over that same span — the recovery shape #9 established for `LEX0004`, adding no token kind.

Note on the issue's implementation note: it suggests accepting `_` only when the next byte is a digit, which would leave `1_` lexing as `1` followed by the identifier `_` and produce **no** lexical diagnostic. That contradicts Requirements 3/8 and the `1_` acceptance criterion, so the requirements were followed: the separators are consumed into the literal and then judged.

### Acceptance criteria

- [x] The lexer produces one integer token for `1_000`.
- [x] A lexer test shows that the value of `1_000` equals the value of `1000`.
- [x] A lexer test shows that `1_` produces one lexical diagnostic.
- [x] A lexer test shows that `1__0` produces one lexical diagnostic.
- [x] A lexer test shows that `_1` stays one identifier token.
- [x] A lexer test shows that `1_000.5` and `1.000_5` are float tokens.
- [x] `openspec/specs/bootstrap-lexer/spec.md:14-15` gets an amendment for the separator.

Tests: baseline 3 failures, after 3 failures, +10 new tests

The 3 failures are the pre-existing `packages/compiler/test/Instances.test.ts` host-triple golden mismatches; re-run in isolation and unchanged. `test:native-acceptance` passes separately. Pre-existing failures in `packages/wasm` (1) and `packages/compiler-cli` (3) were confirmed identical on the base branch.

New coverage: five `Lexer.test.ts` cases (separated integers of every base, separated floats, `_1`/`x_1` identifiers, a table of nine misplaced-separator spellings asserting one diagnostic over the literal span, and `0x_` keeping `LEX0004`), two `IntegerLiteral.test.ts` cases, and `DigitSeparatorAcceptance.test.ts` with three-engine agreement (evaluator, Wasm, native) on separated integer, hex, binary, octal, and float values.

---

### DIFF
--- a/openspec/specs/bootstrap-lexer/spec.md
+++ b/openspec/specs/bootstrap-lexer/spec.md
@@ -13,8 +13,9 @@ distinct token kind, the keywords `pub`, `fn`, `return`, `let`, `move`, and the 
 `-`, `->`, and end-of-file. An identifier SHALL begin with an ASCII letter or underscore and
 continue with ASCII letters, digits, or underscores. An integer literal written without a base
 prefix SHALL contain one or more ASCII digits and SHALL be read in base ten; the base prefixes are
-specified by "Integer literals select a base from their prefix". A `-` immediately followed by `>`
-SHALL remain one arrow token; any other `-` SHALL be one minus token.
+specified by "Integer literals select a base from their prefix", and the `_` digit separator is
+specified by "Number literals accept a digit separator between digits". A `-` immediately followed
+by `>` SHALL remain one arrow token; any other `-` SHALL be one minus token.
 
 #### Scenario: Lex the first parser fixture
 
@@ -299,3 +300,43 @@ to a value SHALL read the base from the token's own source slice rather than ass
 
 - **WHEN** a compiled program returns `0xff` where another returns `255`
 - **THEN** both programs produce the same value, and a prefixed literal outside its selected type's range produces the existing out-of-range diagnostic
+
+### Requirement: Number literals accept a digit separator between digits
+
+An integer or floating-point literal SHALL accept the byte `_` as a digit separator, and the
+separator SHALL be well placed only with a digit of the same digit run immediately on each side.
+A separator therefore SHALL NOT open or close a literal, SHALL NOT follow another separator, SHALL
+NOT sit immediately after a base prefix, and SHALL NOT sit immediately before or after the decimal
+point or the exponent letter. The lexer SHALL consume the separators of a literal itself, keeping
+one token rather than a literal followed by an identifier, and SHALL NOT require any group size.
+A literal whose separators are not all well placed SHALL be one invalid token covering the literal
+and SHALL produce exactly one lexical diagnostic over that same span, without introducing a new
+token kind. Every conversion of a literal to a value SHALL read it without its separators, so a
+separated literal SHALL retain the exact value of its separator-free spelling. The `_` byte SHALL
+keep its identifier meaning everywhere else, so a spelling beginning with `_` remains one
+identifier.
+
+#### Scenario: Separate the digits of an integer literal
+
+- **WHEN** the source bytes spell `1_000 1_048_576 0b1010_0000 0xff_ff`
+- **THEN** the stream contains exactly four integer literal tokens whose slices retain their separators, with no lexical diagnostic
+
+#### Scenario: Separate the digits of a float literal
+
+- **WHEN** the source bytes spell `1_000.5 1.000_5 1e1_0`
+- **THEN** the stream contains exactly three floating-point literal tokens, with no lexical diagnostic
+
+#### Scenario: Preserve the value of a separated literal
+
+- **WHEN** a compiled program compares `1_000` with `1000` and `1.000_5` with `1.0005`
+- **THEN** each pair is equal on every execution engine
+
+#### Scenario: Reject a misplaced separator
+
+- **WHEN** the source bytes spell any of `1_`, `1__0`, `0x_ff`, `1_.5`, `1._5`, `1_e5`, or `1e_5`
+- **THEN** the lexer emits one invalid token covering the literal and exactly one lexical diagnostic
+
+#### Scenario: Keep a leading underscore an identifier
+
+- **WHEN** the source bytes spell `_1` followed by `x_1`
+- **THEN** both spellings remain single identifier tokens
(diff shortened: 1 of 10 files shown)


julia-script:
Orchestrator note (run-s1k9) — verification, one decision that needs your confirmation, and a finding that explains the repo's "pre-existing" test failures.

**Verified:** `Closes #10`, 7/7 acceptance criteria ticked, correctly stacked (`base: agent/9-hex-binary-octal-literals`, base sha `59ebbba` = the rebased #9), test delta 3 → 3.

## Needs your confirmation: the issue contradicts itself

Issue #10's **Implementation note** says to accept `_` only when the next byte is a digit. Its **Requirements 3/8** and its acceptance criterion *"`1_` produces one lexical diagnostic"* require the opposite. The note's approach lexes `1_` as `DecimalInteger(1)` followed by the identifier `_`, producing **zero** lexical diagnostics — which fails that acceptance criterion outright.

The worker followed the normative Requirements over the non-normative note: the digit run is scanned greedily over digits and separators, then judged, so a misplaced separator stays inside the literal and yields one `Invalid` token with one new `LEX0005` diagnostic over the literal's span — the same recovery shape #9 established for `LEX0004`. I agree with that reading, since the acceptance criteria are decisive and the note cannot satisfy them. But it is a normative call made by an agent, so please confirm it is what you want before merging.

## Behavior change worth a look

`1._5` previously lexed as `1`, `.`, `_5` — a field projection of a member named `_5`. It is now a rejected float literal, which Requirement 6 ("reject a `_` directly after the decimal point") demands. `x._5` and `1.foo` are unaffected. No valid program in the repo relies on the old reading, but it is a real change to how an existing spelling parses.

## What the three "pre-existing" failures actually are

This worker ran them in isolation and identified the cause: the three `packages/compiler/test/Instances.test.ts` failures are **host-triple golden mismatches** — the goldens were recorded for `aarch64-apple-darwin` and this machine is `x86_64-unknown-linux-gnu`. They are not flaky, not timeouts, and not toolchain breakage; they will fail deterministically on any non-macOS host and pass on an Apple-silicon one. That resolves a discrepancy I hit at the start of this run: the brief predicted four macOS failures in `StringBackend`/`StringAcceptance`/`OsFileSystem` from the external `clang -c -x ir` step, and I measured three entirely different ones here. Both observations are consistent — the failing set is host-dependent, so "pre-existing failures" is only meaningful with a host attached.

The full pre-existing set on `x86_64-unknown-linux-gnu` is now: **3 in `@silk-effect/compiler`** (`Instances.test.ts`), **3 in `@silk-effect/compiler-cli`** (`FormatCommand`/`FormatWorkflow`), and **1 in `packages/wasm`** — the last confirmed identical on the base branch with this PR's work stashed.

As on the other PRs in this run, the red `validate` and `webcontainer-browser` checks are inherited from `main`, which has been red for its last 15 pushes ([details](https://github.com/julia-script/silk-effect/pull/53#issuecomment-5268292276)).

---

PR #59, opened 2026-08-12
syntax: Accept else if without a brace-delimited arm

### LINKED ISSUE #5: syntax: Accept else if without a brace-delimited arm
## Context

Silk has an `if` statement with an optional `else` arm. The parser accepts only a
brace-delimited block after `else`. A user who writes `else if` gets a parse error.
This makes a chain of conditions difficult to read, because each condition adds one
more level of indentation.

## Current behavior

The parser expects a `LeftBrace` token after the `else` keyword
(`packages/compiler/src/Parser.ts:1411`). This code fails:

```silk
if x < 0 {
  return 0
} else if x < 10 {
  return 1
} else {
  return 2
}
```

The user must write nested blocks instead:

```silk
if x < 0 {
  return 0
} else {
  if x < 10 {
    return 1
  } else {
    return 2
  }
}
```

## Requirements

1. The parser must accept an `if` statement directly after the `else` keyword.
2. The parser must keep the current behavior for `else` followed by a block.
3. The syntax tree must record the chained `if` as a nested conditional statement.
4. The syntax tree must stay lossless, so the formatter can print the source again.
5. The formatter must print `} else if condition {` on one line.
6. The change must not add a new node type to the syntax tree.

## Example

```silk
pub fn classify(value: i32) -> i32 {
  if value < 0 {
    return 0
  } else if value < 10 {
    return 1
  } else if value < 100 {
    return 2
  } else {
    return 3
  }
}
```

## Out of scope

- An `if` expression that returns a value. The `if` statement stays a statement.
- A `match` replacement for long condition chains.
- An `else if` arm in any position other than after `else`.

## Dependencies

None.

## Acceptance criteria

- [ ] The parser accepts `else if` and makes a nested conditional statement.
- [ ] A parser test shows a chain of three conditions.
- [ ] The formatter prints `} else if condition {` on one line.
- [ ] A formatter test shows that the output is stable after two runs.
- [ ] `openspec/specs/bootstrap-syntax/spec.md:370-385` gets an amendment for the new arm form.

julia-script:
Closes #5

Stacked on #58 (`agent/10-digit-separator`), which is itself stacked on #9. Review/merge that one first; this PR targets `agent/10-digit-separator`, not `main`.

Lets an `if` follow the `else` keyword directly, so a chain of conditions reads at one indentation level. A chained arm is recorded as a `ConditionalStatement` nested directly inside the conditional it continues — the same tree shape the equivalent nested-block spelling already produced — so no node kind is added and the formatter needed no change at all to print `} else if condition {` on one line.

### Acceptance criteria

- [x] The parser accepts `else if` and makes a nested conditional statement. `parseConditionalStatement` recurses when an `IfKeyword` follows `else`; `IfKeyword` joins `LeftBrace` as a recovery anchor, so an `else` arm that is neither stays the existing bounded recovery.
- [x] A parser test shows a chain of three conditions. `parses a chain of three conditions as nested conditional statements` asserts three nesting levels, one `Block` per chained arm and two on the innermost, no fourth level, and lossless token traversal.
- [x] The formatter prints `} else if condition {` on one line. Satisfied by the existing `ConditionalStatement` printer, since the chained arm is just that node kind printed with a space prefix — no formatter change was required.
- [x] A formatter test shows that the output is stable after two runs. `formats a chained else if arm on one line and idempotently` asserts the exact bytes, that every `else if` line matches `} else if … {`, then reparses and reformats asserting identical bytes and `changed === false`.
- [x] `openspec/specs/bootstrap-syntax/spec.md` gets an amendment for the new arm form. The "Conditional statements parse losslessly" requirement now specifies the chained arm form, that it introduces no additional node kind and nests one level per arm, that `if` introduces a chained arm only immediately after `else`, and the one-line printing rule — plus three scenarios (three-condition chain, one-line printing, and recovery of an `else` arm that is neither a block nor a chained `if`).

### Note on scope

The Requirements cover parser, tree, and formatter only, but `Elaboration` selected conditional arms with `directNodes(element, 'Block')`, which finds no second block for a chained arm. Left alone, `else if` would have parsed cleanly and then elaborated with an empty else, silently dropping every arm after the first — accepted syntax that miscompiles. The conditional branch of `analyzeStatements` is therefore extracted into a recursive helper so a chained arm elaborates to a nested `IfStatement` fact in the `otherwise` list, identical to what the nested-block spelling already produced. Region allocation order is unchanged, so lowering and determinism goldens are untouched.

Also worth recording: the issue's Example has no trailing `return` after the chain, which trips the pre-existing "function body ends in exactly one trailing return" rule (`PAR0004`) — that rule is unrelated to this issue and unchanged here, so the tests spell the trailing return.

Tests: baseline 3 failures, after 3 failures, +6 new tests

New `packages/compiler/test/ElseIfAcceptance.test.ts` gives the three-engine agreement (interpreter, WebAssembly, native via clang): a chain selecting each of its four arms, that chain proved equal to the same logic written as nested blocks, and a `SEM0011` non-boolean condition on a chained arm. The 3 remaining failures are the documented `Instances.test.ts` host-triple goldens (recorded for `aarch64-apple-darwin`, run here on `x86_64-unknown-linux-gnu`). `compiler-cli` still shows only its 3 pre-existing `FormatCommand`/`FormatWorkflow` failures. `pnpm typecheck` and `biome check .` are clean.

---

### DIFF
--- a/openspec/specs/bootstrap-syntax/spec.md
+++ b/openspec/specs/bootstrap-syntax/spec.md
@@ -371,12 +371,17 @@ and the expression recovers at the existing statement and argument boundaries.
 
 The parser SHALL accept `if <expression> { <statements> } else { <statements> }` wherever a
 statement may appear, with the `else` arm optional and no parentheses around the condition, per
-the accepted surface. Each arm SHALL be a brace-delimited statement sequence (bindings,
-conditionals, and return statements), and the function body SHALL still end in exactly one
-trailing return statement after any conditionals. `true` and `false` SHALL parse as boolean
-literal expressions wherever an expression is allowed. Missing conditions, braces, and damaged
-arms SHALL remain explicit recovery data bounded by the existing statement and declaration
-anchors, and every token SHALL be retained losslessly.
+the accepted surface. The taken arm SHALL be a brace-delimited statement sequence (bindings,
+conditionals, and return statements). The `else` arm SHALL be either such a brace-delimited
+statement sequence or a chained conditional statement introduced by `if` immediately after the
+`else` keyword; the chained form SHALL be recorded as a conditional statement nested directly
+inside the conditional it continues, so a chain of conditions SHALL introduce no additional node
+kind and SHALL nest one level per arm. `if` SHALL introduce a chained arm only immediately after
+`else`. The function body SHALL still end in exactly one trailing return statement after any
+conditionals. `true` and `false` SHALL parse as boolean literal expressions wherever an expression
+is allowed. Missing conditions, braces, and damaged arms SHALL remain explicit recovery data
+bounded by the existing statement and declaration anchors, and every token SHALL be retained
+losslessly, so the formatter SHALL print a chained arm as `} else if <condition> {` on one line.
 
 #### Scenario: Parse a conditional with both arms
 
@@ -387,6 +392,21 @@ anchors, and every token SHALL be retained losslessly.
 
 - **WHEN** a body spells `if flag { return 1 } return 0`
 - **THEN** the conditional retains one arm and no else branch, and the trailing return remains a sibling statement
+
+#### Scenario: Parse a chain of three conditions
+
+- **WHEN** a body spells `if first { return 1 } else if second { return 2 } else if third { return 3 } else { return 4 } return 0`
+- **THEN** the block contains one conditional statement whose else keyword is followed by a nested conditional statement, nested once more for the third condition, with the two chained conditionals retaining one brace-delimited arm each and the innermost retaining both arms, and the trailing return remains a sibling statement
+
+#### Scenario: Print a chained arm on one line
+
+- **WHEN** the formatter prints a conditional whose else arm is a chained conditional
+- **THEN** the arm is printed as `} else if <condition> {` on one line, and printing the result again produces identical bytes
+
+#### Scenario: Recover an else arm that is neither a block nor a chained if
+
+- **WHEN** a body spells `if first { return 1 } else while second { }`
+- **THEN** the else keyword retains an explicit missing arm with one parser diagnostic, and every token stays in the tree
 
 #### Scenario: Parse boolean literals
 
--- a/packages/compiler/src/Elaboration.ts
+++ b/packages/compiler/src/Elaboration.ts
@@ -6118,6 +6118,57 @@ const analyzeStatements = (
     return region
   }
 
+  const analyzeConditional = (
+    element: SyntaxTree.Node,
+    armScope: Scope,
+    armLoopStack: ReadonlyArray<Hir.LoopId>,
+  ): StatementFact => {
+    const region = nextRegion()
+    const conditionNode = statementExpressionNode(element)
+    const condition = analyzeExpression(
+      context.source,
+      conditionNode,
+      context.declarations,
+      context.declaration,
+      armScope,
+      context.resolution,
+    )
+    if (condition === undefined) {
+      throw new RangeError(`Semantic analysis cannot analyze ${conditionNode.kind}`)
+    }
+    context.diagnostics.push(...condition.diagnostics)
+    if (condition.fact.type._tag === 'Available' && condition.fact.type.type !== 'bool') {
+      context.diagnostics.push(
+        Diagnostic.conditionNotBool(
+          Type.encode(condition.fact.type.type),
+          condition.fact.syntax.span,
+        ),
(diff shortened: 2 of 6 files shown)


julia-script:
Orchestrator review (run-s1k9) — P1 reviewer pass. **Approve, nothing blocking.** Three non-blocking notes at the end, one of which is a correction to this PR's own description.

## The scope extension was necessary — and the risk was larger than stated

The "Note on scope" understates it. Pre-change, `analyzeStatements` selected both arms with `directNodes(element, 'Block')` + `arms.at(1)`. For a chained arm the outer `ConditionalStatement`'s direct children are `[IfKeyword, condition, Block, ElseKeyword, ConditionalStatement]` — exactly **one** `Block`, so `arms.at(1)` is `undefined` and `otherwise` becomes `[]`. A parser-only change would therefore have done two bad things, not one:

1. silently dropped every arm after the first from the fact tree, and
2. **never elaborated the chained conditions or blocks at all** — so a type error inside a chained arm (e.g. `SEM0011` on a non-boolean condition) would go entirely unreported.

Accepted syntax that both miscompiles *and* suppresses diagnostics. Going outside the stated Requirements was correct.

## Equivalence to the nested-block spelling — verified

- **Condition evaluation** happens at the same point (region allocated first, then `analyzeExpression`, then the bool check); the extracted helper is logically byte-identical to the old inline block, differing only in `scope`→`armScope` and `loopStack`→`armLoopStack`.
- **`otherwise` shape matches.** Chained yields `[IfStatement]`; nested-block yields `analyzeStatements(elseBlock)`, which inlines a block's statements without emitting a wrapper fact — so the two lists are identical.
- **A chained arm's own trailing `else` binds innermost.** In the recursive call `chained` is `undefined` for the last link, so its `else` Block is used; and `Parser.ts:1412-1415` consumes the inner `else` inside the inner call.
- **`directNode(element, 'ConditionalStatement')` cannot false-positive** — a conditional inside a *recovered* `else` arm ends up within the synthesized `Block`, not as a direct child.

## Region order — the silent-miscompile risk — is genuinely unchanged

`nextRegion` pushes onto the single shared `context.regions` array with `ordinal = regionBase + regions.length`, and `regionOrder` is that array verbatim: a **global** acquisition sequence, not per-block. Both spellings visit in the same order (outer region → outer condition → taken-block regions → inner region → inner condition → inner arms), and the nested-block form's extra `Block` allocates nothing at entry. Every downstream `IfStatement` consumer (`Ownership.ts:1019`, `:1513`, `:1851`, `Analysis.ts:587`, `SemanticOccurrence.ts:793`, `ModuleTooling.ts:86`) is generic over `taken`/`otherwise` and keys off `statement.region`. No drop-order divergence.

## Recovery and losslessness

Recovery stays bounded in every traced case — `else return 1`, `else while …`, `else }` (all already `parseBlock` anchors, missing `{` inserted with zero consumption), `else )` (bounded by `EndOfFile`, identical to pre-change), and `else` at EOF. The new recursion always consumes the `if` token before recursing, so it cannot loop without progress.

Losslessness holds: trivia between `else` and `if` is consumed by the recursive `expect` and lands inside the nested node. `Parser.test.ts:1786` and `:1801` run `assertOriginalTokenTraversal`, which asserts the flattened tree's tokens are *reference-identical* and same-count/same-order against the lexer output — a strong check, applied to both the chain and the `else while` recovery case.

All five acceptance criteria are genuinely met, including criterion 3: the formatter really does work unchanged, via the unfiltered local `directNodes` at `Formatter.ts:882` reaching `printToken(IfKeyword, prefix=' ')` at `:885`.

## Non-blocking

1. **This PR's description is inaccurate on one point.** It credits adding `'IfKeyword'` to `expect(state, 'ElseKeyword', ['LeftBrace', 'IfKeyword'])` (`Parser.ts:1411`) with keeping recovery bounded. That anchor is **unreachable**: the call site is guarded by an `ElseKeyword` lookahead, and `expect`'s fast path matches and returns before ever consulting `following`. The list at that site was dead before this PR too. Boundedness comes from `parseBlock`/`endsBlock`. Either drop the argument or fix the wording.
2. `Elaboration.ts:6155` — `otherwiseArm` is computed on the branch where it is unconditionally ignored.
3. Issue #5's Example still will not compile as written (no trailing `return`, trips pre-existing `PAR0004`). The code and tests handle this correctly; the **issue text** is what should be amended, so a future reader does not mistake it for a regression.

---

Issue #34, opened 2026-08-12
stdlib: Add HashMap, HashSet, HashKey, and HashSeed


julia-script:
## Context

`Vector<T>` is the only collection in the standard library. Silk has no hashed lookup
of any kind. The TypeScript compiler uses 348 `new Map` and `new Set` sites across 35
of 64 modules. A self-hosted compiler needs hashed collections for its
symbol table, its declaration index, and its type interner. This makes the work a
self-hosting prerequisite.

## Current behavior

Silk source cannot express a symbol table. A program that needs a keyed lookup must
scan a `Vector<T>` in linear time. There is also no sort, so a program cannot use a
binary search instead.

No spec text mentions hash collections. This work therefore needs a new spec, not an
amendment to an existing one.

## Requirements

1. A new spec must charter the hashed collection surface before the work starts.
2. The `HashKey` interface must declare an equivalence operation and a hash operation.
3. The generic bound form must accept an interface that has two operations.
4. `HashMap<K, V>` must require a `HashKey` witness for the key type.
5. `HashSet<K>` must require a `HashKey` witness for the element type.
6. Two values that are equivalent must give the same hash value.
7. `HashSeed` must make the iteration order deterministic for one seed value.
8. The implementation must use ordinary Silk source over `Allocator` and typed storage.
9. The compiler, the MIR, the evaluator, and the backends must contain no hash operation.

## Example

```silk
pub interface HashKey<T> {
  fn equivalent(left: &T, right: &T) -> bool
  fn hash(value: &T, seed: HashSeed) -> u64
}

pub effect fn insert<K, V>(
  self: &mut HashMap<K, V>,
  key: K,
  value: V
) -> Option<V> ! OutOfMemory ? &mut Allocator
```

## Out of scope

- An ordered map or an ordered set.
- A concurrent collection.
- An iteration protocol. The spec states that no iterable abstraction is required.
- The sort work. A separate ticket covers the sort.

## Dependencies

- The generic bounds decision. `packages/compiler/src/DeclarationIndex.ts:51` declares
  `readonly bound?: string`, which is one unparameterized bound. A `HashKey` witness
  needs two operations, so the bound form must change first.
- The namespace ticket, because a new module needs a manifest namespace.

## Acceptance criteria

- [ ] A new spec records the hashed collection requirements.
- [ ] A test shows insert, lookup, and removal on `HashMap<K, V>`.
- [ ] A test shows that two runs with the same `HashSeed` give the same order.
[...]

Issue #38, opened 2026-08-12
stdlib: Add MAX, MIN, EPSILON, BITS, PI, and INFINITY constants


julia-script:
## Context

No module in the standard library names a numeric limit. None of the 10 integer modules and
none of the 2 float modules declares `MAX`, `MIN`, `EPSILON`, `BITS`, `PI`, or `INFINITY`.
The compiler knows every one of these bounds, because the `checked` intrinsics test against
them. Silk source cannot name them.

## Current behavior

A user who wants the largest `i32` value must write the literal `2147483647` in the source.
A user who wants the smallest value must write `-2147483648`. Each literal is a place where
the user can make a mistake, and no reader can confirm the value without a calculation.

The float case is worse. A user cannot write an infinity literal at all, so a user must
reach for `f64.fromBits` and a hexadecimal bit pattern (`f64.silk:102`). A user cannot write
an epsilon value without the same trick.

Typed `const` already ships, so these declarations are expressible today. This ticket adds
declarations only.

## Requirements

1. Each integer module must declare `MAX` and `MIN` with the type of that module.
2. Each integer module must declare `BITS` as a `u32` value.
3. Each float module must declare `MAX`, `MIN`, `EPSILON`, `INFINITY`, and `NAN`.
4. Each float module must declare `PI` and `E`.
5. Each constant must have an explicit type, so no constant may depend on literal defaults.
6. The value of `MAX` must equal the largest value that the checked intrinsics accept.
7. The `usize` and `isize` constants must match the pointer width of the target.
8. The change must add no intrinsic and no compiler phase.

## Example

```silk
// In i32.silk
pub const MAX: i32 = 2147483647
pub const MIN: i32 = -2147483648
pub const BITS: u32 = 32

// In f64.silk
pub const EPSILON: f64 = 0.0000000000000002220446049250313
pub const PI: f64 = 3.141592653589793

pub fn wouldOverflow(value: i32) -> bool {
  return value > i32.MAX - 1
}
```

## Out of scope

- **A computed constant expression.** The compiler rejects `const computed: i32 = 40 + 2`
  with `SEM0086`, and a test pins that behavior
  (`packages/compiler/test/TypedConstants.test.ts:108`). Each constant must be one literal.
- A generic bound that gives `MAX` for any integer type. That needs the bound form change
  from #34.
- A change to the checked arithmetic intrinsics.
- A conversion between a number and text. See #40.
- The float math functions. See #41.

## Dependencies

[...]

PR #61, opened 2026-08-12
stdlib: Remove the counted identity function from 5 modules

### LINKED ISSUE #39: stdlib: Remove the counted identity function from 5 modules
## Context

Five standard-library modules declare the same private helper function:

```silk
fn counted(value: usize) -> usize {
  return value
}
```

The function returns its argument without a change. It exists only to give a literal the
`usize` type. A bare `0` defaults to `i32`, and a parameter type drives the literal type, so
`counted(0)` makes a `usize` zero. The comment in `vector.silk:22` states this: "A
usize-typed literal anchor: parameters drive literal typing, so bare counts stay explicit."

## Current behavior

The helper is copy-pasted into 5 modules:

- `packages/compiler/stdlib/silk/vector.silk:23`
- `packages/compiler/stdlib/silk/bytes.silk:15`
- `packages/compiler/stdlib/silk/string.silk:34`
- `packages/compiler/stdlib/silk/filesystem.silk:175`
- `packages/compiler/stdlib/silk/os_filesystem.silk:60`

The first three modules use a multi-line body. The last two use a one-line body. The
declarations are otherwise identical.

Each call site reads worse than the value it makes. A struct literal shows the cost:

```silk
return Vector<T> { storage: Empty<T> { anchor: [] }, length: counted(0), capacity: counted(0) }
```

A loop shows it again:

```silk
index = index + counted(1)
```

Typed `const` now ships, so a constant gives the same type with no call. This is cleanup,
not a feature. The generated code must not change, because a constant and an identity call
lower to the same value.

## Requirements

1. One module must declare the shared typed constants for zero and one.
2. Each of the 5 modules must use the shared constants instead of a private `counted`.
3. Each of the 5 `counted` declarations must be removed.
4. The change must not change the behavior of any function.
5. The change must not change the generated MIR for any existing program.
6. Any remaining `usize` literal must get an explicit type, so no literal may default to `i32`.

## Example

```silk
// In usize.silk
pub const ZERO: usize = 0
pub const ONE: usize = 1

// In vector.silk, after the change.
pub fn make<T>() -> Vector<T> {
  return Vector<T> { storage: Empty<T> { anchor: [] }, length: usize.ZERO, capacity: usize.ZERO }
}
```

## Out of scope

- The remaining numeric constants. See ticket 34.
- A change to literal type inference in the compiler. The default stays `i32`.
- A new syntax for a typed literal, such as a `0usize` suffix.
- Any change to a public function signature.

## Dependencies

- Ticket 34 is a good place to declare `ZERO` and `ONE`, because that ticket adds the other
  numeric constants. This ticket can also declare them alone.

## Acceptance criteria

- [ ] A grep for `fn counted` under `packages/compiler/stdlib/silk/` gives no result.
- [ ] A test shows the MIR for a vector program is the same before and after the change.
- [ ] The existing standard-library tests pass with no change to their expectations.
- [ ] A check shows that every `usize` value in the 5 modules has an explicit type.

julia-script:
Closes #39

Stacked on #55 (`agent/33-stdlib-namespace-field`) — both touch the stdlib manifest and the generated table. Review/merge #55 first; the base retargets to `main` automatically once it lands.

`silk/usize` now declares the shared `pub const ZERO: usize = 0` and `pub const ONE: usize = 1`, and silk/vector, silk/bytes, silk/string, silk/filesystem, and silk/os_filesystem name them instead of each copy-pasting `fn counted(value: usize) -> usize { return value }`.

The counts the shared constants do not name — 2, 3, 4, 64, 256 — all sit in operand, argument, or assignment positions whose `usize` type the compiler already fixes, so they became bare literals. The compiler proves they never fall back to `i32`: a `usize`/`i32` mismatch is a hard diagnostic, and the layout plan records each of them as an `AvailableUsizeLiteral`. Naming them would have meant inventing constants like `SIXTY_FOUR`, and the remaining numeric constants are ticket #34's scope.

### Acceptance criteria

- [x] A grep for `fn counted` under `packages/compiler/stdlib/silk/` gives no result. Asserted over every manifest module in `StdlibTypedCountAcceptance.test.ts`.
- [x] A test shows the MIR for a vector program is the same before and after the change — **with one correction, see below.**
- [x] The existing standard-library tests pass with no change to their expectations — **with one narrowing, see below.**
- [x] A check shows that every `usize` value in the 5 modules has an explicit type. The layout plan's literal verdicts are all `AvailableUsizeLiteral`, `silk/usize` contributes `0` and `1`, and `silk/string` contributes `2`, `3`, `4` — every remaining count carries `usize`.

### Two deviations, stated plainly

**Requirement 5 / criterion 2 — MIR does change, by exactly the removed calls.** The issue assumes "the generated code must not change, because a constant and an identity call lower to the same value." The second half is true; the first half is not. `counted` is a real MIR function and each use is a real `call`. A vector program's pre-change MIR carried `fn silk/vector.counted` plus 11 `call silk/vector.counted(...)` operations; afterwards both are gone. What is preserved — and what the test asserts — is every *value*: `silk/vector.make<i32>` still ends in `Literal 0 : usize`, `Literal 0 : usize`, `Construct`, exactly the two typed immediates the identity calls used to return, and the program returns the same result on evaluation, direct WebAssembly, and native LLVM. A literally byte-identical MIR is unreachable for any change that deletes a called function, so the test proves value and observable-behavior equivalence instead.

**Criterion 3 — one existing expectation is narrowed, not relaxed.** `test/Usize.test.ts` asserted `plan.value.literalVerdicts.length === 3`, counting the three `usize` literals in its own program. Its program calls `usize.add`, so `silk/usize` joins the closure, and a `usize` constant is target-checked whether used or not — the two new constants push the count to 5. The assertion now filters verdicts to `sourceId === 'usize/program'` and still expects 3, which is the number the test always meant. No other test expectation moved.

The 3 are the pre-existing host-triple golden mismatches in `test/Instances.test.ts` (goldens recorded for `aarch64-apple-darwin`, run here on `x86_64-unknown-linux-gnu`). `pnpm typecheck` and `pnpm exec biome check .` are clean.

---


julia-script:
Orchestrator note (run-s1k9) — verified: `Closes #39`, 4/4 criteria addressed, base `agent/33-stdlib-namespace-field` (correct stack), delta 3 → 3.

**Both deviations are sound, and one of them means issue #39 has a false premise you may want to correct in the issue text.**

Requirement 5 and acceptance criterion 2 assert the MIR is unchanged, reasoning that "a constant and an identity call lower to the same value." The second half is true; the first is not, and it cannot be. `counted` is a real MIR function and each use is a real `call` — a vector program carried `fn silk/vector.counted` plus 11 `call silk/vector.counted(...)` operations before this change, and none afterward. **Byte-identical MIR is unreachable for any change that deletes a called function**, so the criterion as written could never be satisfied by a correct implementation of Requirement 1. Substituting value-and-behavior equivalence (`silk/vector.make` still ending in `Literal 0 : usize, Literal 0 : usize, Construct`, plus identical results on all three engines) is the right reading of the intent.

The criterion-3 narrowing is legitimate too, and worth understanding because it is a property of the language, not a test smell: `Usize.test.ts` asserted a *global* `literalVerdicts.length === 3`, but a `usize` constant is target-checked whether or not it is used, so adding `ZERO`/`ONE` to `silk/usize` pushed the global count to 5. Filtering to `sourceId === 'usize/program'` and keeping 3 restores the number the assertion always meant. No expectation was weakened.

The one judgment call left implicit: counts 2, 3, 4, 64, and 256 became bare literals rather than named constants, since Requirement 1 names only `ZERO` and `ONE` and naming the rest would mean inventing `SIXTY_FOUR`. They sit in positions where the compiler fixes `usize`, and the layout plan records each as `AvailableUsizeLiteral` — which is exactly what criterion 4 checks. Remaining numeric constants belong to #38/#34's scope, both currently `needs-definition`.

---

PR #65, opened 2026-08-12
lsp: Add find-all-references and rename

### LINKED ISSUE #46: lsp: Add find-all-references and rename
## Context

The compiler part of this work exists today. `Analysis.semanticOccurrencesInRange`
(`packages/compiler/src/Analysis.ts:199`) returns the token-level occurrences that
back go-to-definition. Find-all-references reads the same index in the opposite
direction. The server capability list
(`packages/lsp/src/Server.ts:50-65`) advertises `definitionProvider` but no
`referencesProvider` and no `renameProvider`.

## Current behavior

A user gets the declaration of a name from go-to-definition. The editor gives no
list of the uses of that name. The editor gives no rename command. A user renames a
declaration by hand in every module and then compiles to find the modules that were
missed.

## Requirements

1. The server must advertise `referencesProvider` in its initialize response.
2. The server must answer a references request with every occurrence that has the same semantic identity.
3. The server must include the declaration in the response when the client sets `includeDeclaration`.
4. The server must advertise `renameProvider` with `prepareProvider` set to true.
5. The prepare-rename response must give the range of the name token, or must fail on a token that has no declaration.
6. The rename must reject a new name that collides with an existing binding in the same flat namespace.
7. The rename must use the existing collision machinery of `SEM0016` for that check, because Silk has one flat non-shadowing namespace (`openspec/specs/bootstrap-name-resolution/spec.md:60-63`).
8. The server must return one `WorkspaceEdit` that covers every module in the project.

## Example

```silk
// module: geometry
pub fn area(width: i32, height: i32) -> i32 { width * height }

// module: main
import geometry

// A rename of `area` to `surface` changes the declaration and every use.
// A rename of `area` to an existing top-level name fails with the SEM0016 reason.
```

## Out of scope

- A rename across a project that does not compile.
- A rename of a module name or of a file name.
- A references request that reaches outside the project source graph.
- A call hierarchy. Ticket 45 covers that work.

## Dependencies

- None. The occurrence index and the collision diagnostic exist today.

## Acceptance criteria

- [ ] A test asserts that the initialize response advertises `referencesProvider` and `renameProvider`.
- [ ] A test finds every use of one declaration across two modules.
- [ ] A test renames a declaration and asserts the `WorkspaceEdit` covers both modules.
- [ ] A test rejects a rename to a name that already exists at the top level.
- [ ] A test asserts a prepare-rename request on a keyword token fails.
- [ ] The `language-server-navigation` spec gets a requirement for references and one for rename.

julia-script:
Closes #46

Stacked on #56 (`agent/45-diagnostic-edits`) — the LSP lane shares `packages/lsp/src/`, so this branches from that PR rather than `main`. Review the two-commit delta on top of #56.

## What this does

`Document.references`, `Document.prepareRename`, and `Document.rename` read the same token-level occurrence index that backs go-to-definition, in the opposite direction: every occurrence whose `SemanticOccurrence.identityKey` equals the identity selected at the request position. `Server.ts` advertises `referencesProvider` and `renameProvider: { prepareProvider: true }` and wires the three handlers, failing prepare-rename and refused renames with `LSPErrorCodes.RequestFailed`.

Two correctness devices are worth calling out, since a rename that edits some but not all references is worse than no rename at all:

- **Project-wide, not root-scoped.** A root view's `closure.rootModule` names one root, but `ModuleClosure.view` keeps the whole project's `modules` and `sources`, and the occurrence index spans every module of the accepted project revision. Enumerating `Analysis.sources` therefore reaches every module, not just the request document's import closure.
- **Spelling-filtered rename.** An imported member and its local alias share one semantic identity, so `area` and `region` in `import geometry { area as region }` are the same identity with different spellings. References returns both (they *are* the same identity). Rename edits only the occurrences whose analyzed spelling equals the selected name, so renaming the source name yields `import geometry { surface as region }` with `region`'s uses untouched, and renaming the alias leaves the declaration alone. A rename that cannot place some occurrence in a document is refused outright rather than emitted partially.

The `SEM0016` check runs against each module whose *flat namespace* the rename would extend — the declaration site and import bindings — via `Analysis.lookupName` plus `Diagnostic.bindingConflict`. A module that reaches the declaration only through a qualifier (`geometry.area`) keeps its own unrelated top-level names, which is why the check is scoped by occurrence role rather than applied to every edited module.

## Note on scope

Requirement 8 says the `WorkspaceEdit` must cover "every module in the project". The server's project is the accepted project revision: open documents as roots plus every module they transitively import (closed project files included, resolved from disk). A *closed* module that imports the renamed declaration is not part of that closure and is not edited, because rooting analysis at open documents is what `language-server-synchronization`'s "Open documents form project-scoped overlays" requirement specifies; widening the root set is a synchronization change, not a navigation one. The spec text added here says "every module of the accepted project revision" so the boundary is stated rather than implied. This is a visible failure mode (the closed module stops compiling), not a silent one, and the issue puts renames across a project that does not compile out of scope.

## Acceptance criteria

- [x] A test asserts that the initialize response advertises `referencesProvider` and `renameProvider` — `serves project-wide references and renames over real stdio` in `packages/lsp/test/Server.test.ts`.
- [x] A test finds every use of one declaration across two modules — `finds every use of one declaration across two modules`, which also asserts `includeDeclaration: false` removes exactly the declaration-site occurrence.
- [x] A test renames a declaration and asserts the `WorkspaceEdit` covers both modules — `renames a declaration with one workspace edit covering both modules`, plus the stdio test.
- [x] A test rejects a rename to a name that already exists at the top level — `refuses a rename that collides with a top-level name in any edited module`, covering both a collision in the declaring module and one in an importing module's flat namespace.
- [x] A test asserts a prepare-rename request on a keyword token fails — `prepares a rename on a name token but fails on a keyword`, over the wire too.
- [x] The `language-server-navigation` spec gets a requirement for references and one for rename — two requirements with nine scenarios in `openspec/specs/language-server-navigation/spec.md`.

One extra test beyond the criteria, `renames the imported source name without disturbing a local alias`, guards the aliased-import case described above.

Tests: baseline 3 failures, after 3 failures, +6 new tests
`packages/lsp`: before 46 passed / 0 failed, after 52 passed / 0 failed

The 3 baseline failures are the pre-existing `packages/compiler/test/Instances.test.ts` host-triple golden mismatches (goldens recorded for `aarch64-apple-darwin`, host here is `x86_64-unknown-linux-gnu`); this change touches no compiler source. `test/OsFileSystem.test.ts > lowers the OS directory-list provider runner` failed once under parallel load and passes in isolation (`--maxWorkers=1`, 10/10). `pnpm typecheck` and `pnpm exec biome check .` are clean.

---

### DIFF
--- a/openspec/specs/language-server-navigation/spec.md
+++ b/openspec/specs/language-server-navigation/spec.md
@@ -155,3 +155,72 @@ because its body calls an intrinsic.
 
 - **WHEN** definition is requested on `Intrinsic.i32Add`
 - **THEN** no source location is invented and the intrinsic semantic identity remains available
+
+### Requirement: References enumerate one semantic identity across the project
+
+The language server SHALL advertise references support and SHALL answer a references request with
+every occurrence in the accepted project revision whose semantic identity equals the identity
+selected at the request position, rather than every token that shares its spelling. Occurrences
+SHALL be returned in canonical module order and, within a module, in source order. The
+declaration-site occurrence SHALL be included when the client sets `includeDeclaration` and
+excluded otherwise. A position with no available semantic occurrence SHALL produce no references.
+
+#### Scenario: Client initializes the server
+
+- **WHEN** a compatible client initializes a Silk language-server session
+- **THEN** the returned capabilities advertise references support
+
+#### Scenario: Uses in another module
+
+- **WHEN** references are requested on a declaration used from a second module
+- **THEN** the response contains that module's uses at their exact ranges together with the uses in the declaring module
+
+#### Scenario: Selected member and its local alias
+
+- **WHEN** references are requested through an aliased selected-member import
+- **THEN** the response contains the declaration, the imported source name, the local alias, and every use of that alias, because all of them carry one semantic identity
+
+#### Scenario: Declaration excluded on request
+
+- **WHEN** a references request clears `includeDeclaration`
+- **THEN** the declaration-site occurrence is the only occurrence removed from the response
+
+### Requirement: Rename rewrites one semantic identity in one workspace edit
+
+The language server SHALL advertise rename support with `prepareProvider` enabled. A prepare-rename
+request SHALL return the range of the selected name token, and SHALL fail for a token with no
+source-backed declaration, including keywords, trivia, and intrinsics that have no Silk
+declaration. A rename SHALL return one `WorkspaceEdit` covering every module of the accepted
+project revision, and SHALL edit only the occurrences whose analyzed spelling equals the selected
+name so that an aliased import keeps the local name its own module chose. A rename SHALL be refused
+rather than partially applied when an occurrence cannot be placed in a document.
+
+Because Silk has one flat non-shadowing module namespace, a rename of a name that occupies that
+namespace SHALL be refused when the new spelling is already bound in any module whose flat namespace
+the rename would extend. The refusal SHALL carry the existing `SEM0016` binding-collision code and
+message rather than a rename-specific diagnostic.
+
+#### Scenario: Client initializes the server
+
+- **WHEN** a compatible client initializes a Silk language-server session
+- **THEN** the returned capabilities advertise rename support with prepare support enabled
+
+#### Scenario: Prepare a rename on a keyword
+
+- **WHEN** a prepare-rename request selects a keyword token
+- **THEN** the request fails instead of returning a range
+
+#### Scenario: Rename a declaration used in another module
+
+- **WHEN** a top-level declaration used from a second module is renamed
+- **THEN** one workspace edit changes the declaration and every use in both modules
+
+#### Scenario: Rename an imported source name
+
+- **WHEN** a declaration reached through an aliased selected-member import is renamed
+- **THEN** the imported source name changes and the local alias and its uses keep their own spelling
+
+#### Scenario: Rename onto an existing top-level name
+
+- **WHEN** a rename would give a declaration a spelling already bound in a module's flat namespace
+- **THEN** the rename is refused with the `SEM0016` collision reason and no edit is returned
--- a/openspec/specs/language-server-navigation/spec.md
+++ b/openspec/specs/language-server-navigation/spec.md
@@ -195,6 +195,21 @@ project revision, and SHALL edit only the occurrences whose analyzed spelling eq
 name so that an aliased import keeps the local name its own module chose. A rename SHALL be refused
 rather than partially applied when an occurrence cannot be placed in a document.
 
+The accepted project revision is the rename's outer bound: the open documents that serve as roots
+plus every module they transitively import, closed project files included, resolved from disk. A
+closed module that imports the renamed declaration but is reachable from no open root is therefore
+outside that revision and SHALL NOT be edited. Rooting analysis at open documents is what the
+`language-server-synchronization` capability specifies, so widening the root set is a change to
+synchronization rather than to navigation. The resulting failure is visible — the unedited module
+stops compiling — rather than silent.
+
+Within that revision, an `as` clause binds a name in one module only, so a rename selected through
+such a binding SHALL be confined to the module that wrote the clause. Two modules that alias one
+declaration to the same spelling agree on both identity and spelling, and neither owns the other's
+choice of name. The declaration site and the source half of an import clause name the declaration
+itself and SHALL stay project-wide. This confinement applies to rename alone; references are
+read-only and SHALL continue to report every occurrence of the identity.
+
 Because Silk has one flat non-shadowing module namespace, a rename of a name that occupies that
 namespace SHALL be refused when the new spelling is already bound in any module whose flat namespace
 the rename would extend. The refusal SHALL carry the existing `SEM0016` binding-collision code and
@@ -219,6 +234,16 @@ message rather than a rename-specific diagnostic.
 
 - **WHEN** a declaration reached through an aliased selected-member import is renamed
 - **THEN** the imported source name changes and the local alias and its uses keep their own spelling
+
+#### Scenario: Two modules alias one declaration to the same spelling
+
+- **WHEN** an alias is renamed in a module while a second module aliases the same declaration to the same spelling
+- **THEN** only the selecting module's clause and uses change, and the second module keeps the name it chose
+
+#### Scenario: Rename a project alias of a standard-library member
+
+- **WHEN** a project module's alias of a standard-library member is renamed and the installation aliases that member to the same spelling
+- **THEN** the edit covers the project module alone and reaches no file inside the installed toolchain
 
 #### Scenario: Rename onto an existing top-level name
 
--- a/packages/lsp/src/Document.ts
+++ b/packages/lsp/src/Document.ts
@@ -306,11 +306,51 @@ interface RenameSubject {
 }
 
 /**
- * Resolves the occurrences one rename request would rewrite. An imported member and its local
- * alias share a single identity, so the selected spelling is what separates them: renaming the
- * alias `eq` of `equals as eq` edits the `eq` occurrences alone, all of them owned by the module
- * that chose the alias. Narrowing here, before any refusal is decided, is what lets prepare and
- * rename answer from the same facts.
+ * Tests whether one spelling names, in one module, a declaration the module reaches through its own
+ * `as` clause. Such an alias is the importing module's own name for a declaration it does not own:
+ * the clause introduces it, Silk's one flat module namespace makes the spelling resolve to it
+ * throughout that module, and no other module can see it. A member imported under the declaration's
+ * own spelling introduces no separate name, so the source half of `make as vectorMake` — and a
+ * plain `{ equals }` — still reach the declaration itself.
+ */
+const bindsLocalAlias = (
+  snapshot: Analysis.FrontendSnapshot,
+  module: string,
+  spelling: string,
+  identity: SemanticOccurrence.Identity,
+): boolean => {
+  const key = SemanticOccurrence.identityKey(identity)
+  for (const imported of Analysis.moduleScope(snapshot, module)?.imports ?? []) {
+    if (imported._tag !== 'Available') continue
+    for (const binding of imported.bindings) {
+      if (binding._tag !== 'ImportedMember') continue
+      if (binding.spelling !== spelling || binding.spelling === binding.sourceSpelling) continue
+      const bound = SemanticOccurrence.identityKey(
+        Object.freeze({ _tag: 'DeclarationIdentity', id: binding.declaration }),
+      )
+      if (bound === key) return true
+    }
+  }
+  return false
+}
+
+/**
+ * Resolves the occurrences one rename request would rewrite. A declaration's identity is canonical
+ * and module-independent, so it alone cannot say which occurrences a rename owns; two narrowings
+ * finish the job.
+ *
+ * The selected spelling separates an imported member from its local alias, which share that one
+ * identity: renaming the alias `eq` of `equals as eq` edits the `eq` occurrences alone.
+ *
+ * Spelling is not enough on its own, because unrelated modules pick the same alias for the same
+ * declaration all the time — `silk/bytes` and a project module both writing `make as vectorMake`
+ * agree on identity *and* spelling. An `as` clause binds a name in one module only, so when the
+ * selection resolves through such a binding the rename is confined to the module that wrote it.
+ * The declaration site and the source half of an import clause name the declaration itself and
+ * stay project-wide.
+ *
+ * Narrowing here, before any refusal is decided, is what lets prepare and rename answer from the
+ * same facts.
  */
 const renameSubjectAt = (
   self: Document,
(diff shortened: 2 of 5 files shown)


julia-script:
⚠️ Orchestrator adversarial review (run-s1k9) — **one blocking defect, confirmed in code. A fix is being dispatched.** Plus one scope question that needs your ruling, and three smaller notes.

## BLOCKING — a rename can rewrite the installed toolchain stdlib

```silk
// main.silk
import silk.bool { equals }
pub fn main() -> bool { equals(true, true) }
```

Rename `equals` at 0:19 → `same`.

`matchesOfIdentity` iterates every module in `Analysis.sources(snapshot)` with **no origin filter** (`Document.ts:205-232`), and `uriOf` returns a real URI for toolchain sources:

```ts
case 'Memory':
case 'ToolchainFile':
  return source.origin.uri
```

`Workspace.ts:196-199`. Stdlib modules are in `closure.sources` with `file://` toolchain hrefs, so the emitted `WorkspaceEdit` targets the user's **installed** `stdlib/silk/bool.silk`. Editors apply workspace edits without confirmation, so this silently mutates the toolchain installation — and only the stdlib modules that happen to be in the closure, leaving the rest of the stdlib on the old name.

**Fix being applied:** exclude toolchain-origin sources from rename edits, and refuse outright (`RequestFailed`) when the selected identity is declared in a toolchain module — consistent with this PR's own principle that a rename it cannot complete is refused rather than emitted partially. `references` should keep listing stdlib occurrences; it is read-only.

## Needs your ruling — requirement 8's scope

The reviewer argues the "Note on scope" section understates a real gap. With `src/geometry.silk` declaring `area` and `src/main.silk` (closed, on disk) calling `geometry.area`, opening **only** `geometry.silk` and renaming `area` → `surface` edits just the declaration. `roots = documents` (`Workspace.ts:162`), geometry imports nothing, so the project closure is `{geometry}` — nothing ever scans `sourceRoot` for reverse dependencies. `main.silk` stops compiling. `references` is silently truncated the same way, and the `SEM0016` check never sees `main`.

Your PR argues this is the accepted-project-revision boundary mandated by `language-server-synchronization`, and that the failure is visible rather than silent. That is a coherent reading, and I am not overruling it with an agent. But requirement 8 says "every module in the project", and a user renaming from a single open file is the *common* case, not an edge case — so I want your decision: **is open-roots-closure the intended scope (and requirement 8 should be reworded), or should rename scan `sourceRoot` for reverse dependencies (a synchronization-spec change)?**

One thing that is not a judgment call: **the new tests cannot catch this**, because they pass all modules to `ProjectAnalysis.make` directly rather than going through the open-documents path the server actually uses. Whichever scope you choose, a test that exercises the real root-selection path would pin it.

## Smaller notes

- **False refusal.** `geometry` declares `area`; `main` has `import geometry { area as region }` plus its own `pub fn surface()`. Renaming `area` → `surface` is refused for `SEM0016`, but `main` binds only the local spelling `region` (`NameResolution.ts:32-33`) and never gains `surface`. The collision check is too broad for the alias case.
- **Local and parameter renames get no collision check at all.** `occupiesFlatNamespace` (`Document.ts:315`) excludes `BindingIdentity`/`ParameterIdentity`, so renaming a local to an existing top-level name silently creates a shadowing conflict.
- **Non-ASCII identifiers.** `SourceFile.spelling` decodes latin-1 (`SourceFile.ts:51`), so `prepareRename` returns a mojibake placeholder. The spelling-equality filter still holds, so edits stay correct — cosmetic only.

## Verified clean

LSP ranges are genuinely UTF-16 code units (`LineIndex.positionOf`/`offsetOf` decode bytes), so the range correctness I was most worried about holds. The alias device works as described — `sourceToken` and `localToken` are both emitted (`SemanticOccurrence.ts:946-956`) and the spelling filter keeps only the source half. Qualified `geometry.area` uses are renamed and correctly excluded from the flat-namespace check. Comments and strings are not occurrences. Edits are deduped by `module:start:end` and non-overlapping. Prepare-rename refuses keywords, trivia, and unresolved positions.

---

julia-script:
✅ Orchestrator follow-up (run-s1k9) — **blocking toolchain-rename defect fixed** (`3dc1f61`, now `6ffad20` after rebase onto #56's updated base). `packages/lsp` 52 → 55 passing; compiler unchanged at the 3 baseline failures.

**The most valuable part of this fix is not the fix — it is that the test suite could not have caught the bug.** The existing `openProject` helper used `SourceResolver.memory`, whose `resolveStandardLibrary` returns `SourceOrigin.memory()`, so stdlib modules never had toolchain origins and the defect was **unreproducible in tests by construction**. The worker replaced it with a `toolchainResolver` mirroring the real `FileSourceResolver` — project modules memory-origin, reserved modules `SourceOrigin.toolchainFile(entry.sourceUrl.href)` — and made `uriOfModule` answer stdlib modules with their real installed href, matching `Workspace.uriOf`. All 30 pre-existing Document tests still pass under that stricter resolver, which is a good sign it is faithful rather than merely convenient.

With that in place the defect reproduced **end to end against the real installed stdlib**, not a mock. Reverting only `src` and rebuilding, the e2e timeout dump captured the raw `textDocument/rename` response containing:

```
"file:///…/packages/compiler/stdlib/silk/bool.silk"  newText: "sameAs"
```

**The fix:** `isToolchainModule` reads `Analysis.sources(snapshot).get(module)?.origin._tag === 'ToolchainFile'` — origin, not a `silk/` name prefix, as asked. `rename` refuses with a new `LSP0002` when the selected identity's declaring module is toolchain-origin, checked *before* the `SEM0016` flat-namespace check so the toolchain reason wins. A second `LSP0002` guard inside the match loop refuses if any collected occurrence lands in a toolchain module — unreachable today since stdlib never imports user modules, but it makes "never emit a toolchain edit" structural rather than incidental, which is the right instinct. `prepareRename` also refuses, so editors never offer the UI at all.

`references` and `matchesOfIdentity` are untouched, so listing stdlib occurrences still works — correct, since that path is read-only.

Two of the three new tests pass both before and after by design, as regression guards: a user rename beside stdlib imports still edits only user modules, and `references` still lists the stdlib declaration.

**Still open for you, unchanged:** the requirement-8 scope question from my previous comment — whether open-roots-closure is the intended project scope (and requirement 8 should be reworded), or whether rename should scan `sourceRoot` for reverse dependencies. That one is a spec decision, and the new realistic resolver makes it straightforward to test whichever way you rule.

---

julia-script:
⚠️ Orchestrator round 2 (run-s1k9) — **the toolchain fix over-refuses. Round 3 dispatched.** This one is on me: the guard I specified was broader than the bug required.

## The regression

```silk
// main.silk
import silk.bool { equals as eq }
pub fn main() -> bool { return eq(true, true) }
```

Renaming the **local alias** `eq` — at the import clause or the call site — is now refused with `LSP0002 "eq is declared in silk/bool, which the installed toolchain owns"`, and `prepareRename` returns `undefined`, so the editor greys out F2.

`eq` is a binding the **user's own module** owns. `SemanticOccurrence` pushes `sourceToken` (`equals`) and `localToken` (`eq`) with the *same* identity and the *same* `declaration`, so the declaring-module pre-check at `Document.ts:378-384` fires on a purely local name. The message is also factually false — `eq` is not declared in `silk/bool`.

**This worked correctly before the toolchain fix**, producing a `RenameEdit` touching only `main`.

## Why the pre-check was never needed

The in-loop guard at `Document.ts:396` is sufficient on its own, and I verified the reasoning against the code:

- Renaming the stdlib declaration, or a non-aliased `import silk.bool { equals }` occurrence — spelling is `equals`, so the filtered matches include the declaration occurrence *inside* `silk/bool`, the in-loop guard fires, correctly refused.
- Renaming an alias `eq` — the spelling filter at `Document.ts:386-388` keeps only `eq`-spelled matches, all in user modules, so no toolchain edit is possible and it should be allowed.

The spelling filter this PR introduced for the alias case already does the work; the pre-check only removed capability. Round 3 removes it, keeps the structural in-loop guard, aligns `prepareRename` with `rename` so there is no editor divergence, fixes the false message, and pins the ordering between the toolchain refusal and `flatNamespaceRefusal` so a future change cannot silently swap which error a user sees.

## Verified clean in this round

The interesting bypasses do not exist: a memory overlay cannot shadow `silk/*` because `ModuleClosure.ts:392` routes every reserved module to `resolveStandardLibrary`, which `Workspace.resolver` passes to `FileSourceResolver` (`ToolchainFile`); opening a stdlib file as a root is rejected at `ModuleClosure.ts:370-372`, and `moduleOf` would name it `bool` anyway so `uriOf`'s open-document branch never matches `silk/bool`. `references` has exactly one caller (`Server.ts:207`) returning Locations only, so leaving it untouched is safe.

One note for later: the test helper gives project modules `SourceOrigin.memory()`, so the real `ProjectFile` origin path (`FileSourceResolver.ts:66`) is still untested by the Document tests.

---

julia-script:
✅ Orchestrator round 3 (run-s1k9) — **over-refusal regression fixed.** `packages/lsp` 55 → 57; compiler unchanged at the 3 baseline failures.

I verified the structure directly: the declaring-module pre-check is gone, and `prepareRename` (`Document.ts:377`) and `rename` (`:433`) now resolve from one shared `renameSubjectAt` and share one `toolchainRefusal` predicate — so a prepare/rename divergence is structurally impossible rather than merely absent today. The in-loop guard at edit-building time is retained as a backstop with its comment updated to say why it is now redundant-but-kept.

The reasoning was confirmed before acting rather than assumed: renaming `equals` keeps the `silk/bool` declaration occurrence in the spelling-filtered set and is refused; renaming the alias `eq` filters it out, leaving only user-module matches. So the pre-check was removable outright and no narrower replacement was needed.

**Ordering is now pinned in both directions** — toolchain + collision → `LSP0002`, collision alone → `SEM0016`. The justification is sound: no replacement spelling can make a toolchain rename legal, so surfacing a binding-conflict message there would send the user chasing a fix that does not exist.

**The false message is gone.** The only remaining `LSP0002` text is `Renaming <name> would edit <module>, which the installed toolchain owns` — a claim about where the edit lands, which is true whenever it fires, rather than a claim about where a name is declared, which was not.

The regression test was verified failing pre-fix (`expected 'RenameRefusal' to equal 'RenameEdit'`) and passing after.

One detail worth recording because it nearly invalidated the test: the existing `positionOf` helper searches by **spelling**, and `'eq'` is a prefix of `'equals'` — so the alias test would have addressed the wrong token. It was refactored onto a `positionAt(source, offset)` helper and the two alias tokens addressed by computed offset. A test that silently points at the wrong token is worse than no test, so catching that mattered.

## Where this PR stands

Three defects found across three rounds, each narrower than the last: the toolchain-stdlib rewrite (blocking, mine to catch), the over-refusal my own fix introduced (blocking, caught by round 2), and now convergence. Per this run's policy of at most two fix rounds per PR, I am not dispatching further fixes here — anything the closing review turns up will be reported for you to judge rather than patched.

**Two things remain open for your decision**, neither a defect:
1. **Requirement 8's scope** — is open-roots-closure the intended project scope (reword requirement 8), or should rename scan `sourceRoot` for reverse dependencies (a synchronization-spec change)?
2. The Document tests give project modules `SourceOrigin.memory()`, so the real `ProjectFile` origin path (`FileSourceResolver.ts:66`) is still unexercised — worth a follow-up now that the `toolchainResolver` infrastructure exists to make it easy.

---

julia-script:
🔎 Orchestrator closing review (run-s1k9) — **one defect found, reported rather than fixed.** Per this run's policy of at most two fix rounds per PR, this is yours to judge: it is narrow, and the correct fix touches the identity model where the two previous fixes both went wrong.

## The over-refusal survives via spelling collision

```silk
import silk.bytes { make as bytesMake }
import silk.vector { make as vectorMake }
```

Rename `vectorMake` — the **project's own** alias binding — to anything. Result: `LSP0002 "Renaming vectorMake would edit silk/bytes, which the installed toolchain owns"`, and `prepareRename` returns `undefined`, so F2 is greyed out. Expected: edits confined to `main`.

**I verified the premise in the tree.** The stdlib independently chose the same alias spellings:

```
packages/compiler/stdlib/silk/bytes.silk:3           make as vectorMake
packages/compiler/stdlib/silk/os_filesystem.silk:52  make as vectorMake
packages/compiler/stdlib/silk/core.silk:1            Layout as LayoutValue
packages/compiler/stdlib/silk/logging.silk:7         utf8Bytes as stringUtf8Bytes
```

`DeclarationIdentity` is canonical and module-independent (`SemanticOccurrence.ts:20-24, 72-76`), so the project's `vectorMake` and `silk/bytes`'s `vectorMake` share **both** identity and spelling. Importing `silk.bytes` puts that module in the closure, and the spelling filter — the device that fixed the previous alias regression — cannot separate them.

So this is fix #2's defect class surviving through spelling instead of declaring module. Refusing is arguably *self-consistent* (an edit really would land in `silk/bytes`), which is what makes it interesting: the guard is behaving correctly given the match set, and the match set is what is wrong.

## Why I am not dispatching a fix

The root cause is that a **local alias binding carries the declaration's identity**, so two independent aliases of the same declaration are indistinguishable. Fixing it properly means giving alias bindings their own binding identity, or narrowing matches by owning module — a change to the identity model, not a guard. Both previous fixes on this PR went wrong in exactly that area, and a third attempt without a settled model would be guesswork.

**Practically:** the trigger requires a user to choose the same alias spelling the stdlib uses internally for the same declaration. That is plausible (`vectorMake`, `LayoutValue`) but not common, and the failure is a refusal with a message that is *true about where the edit would land* — not corruption, and not silent.

**Options, your call:** merge as-is and open a follow-up for the identity-model fix; or hold this PR until alias bindings get distinct identities. If you merge, this is worth an issue so it does not get lost — I have not filed one, since the choice affects whether it is a PR change or a follow-up.

## Everything else in the closing round came back clean

The `positionAt` refactor is byte-identical to the old offset math and `positionOf` semantics are unchanged, so no existing test silently moved token. `toolchainRefusal` covers every emitted edit — edits are built only from `matches`, each origin-checked at `Document.ts:442`. Refusal ordering is correct. `references` is untouched by both fixes, and the wire path (`Server.ts:204-249`) maps refusals to `RequestFailed` consistently.

---

Issue #26, opened 2026-08-12
effects: add Effect.ensuring to run a finalizer on every exit


julia-script:
## Context

Silk has no combinator that runs a finalizer after an Effect, whatever the outcome.
`Effect.catch` recovers a typed failure, but it does not run on success. The closed
combinator list omits `ensuring` (`bootstrap-flow-functions:246-248`), and nothing rejects
it. The matrix names `ensuring` as the one item in its group that is more than a library
addition (`FEATURE-MATRIX.md:368`).

## Current behavior

A caller who needs cleanup after both outcomes writes the cleanup call twice: once in the
success path and once in each `catch` arm. There is no single place to put it.

## Why this is not a pure library addition

A finalizer that must run on both success and typed failure interacts with two existing
mechanisms:

1. **Cleanup plans.** Local cleanup runs in reverse acquisition order. The finalizer must
   have a defined order against the cleanup of the Effect's own locals.
2. **`provideWith` acquisition.** `provideWith` acquires a provider once
   (`effects.silk:185-192`). A finalizer that releases that provider must agree with the
   acquisition point. #27 changes that acquisition point, so this ticket waits for
   it.

## Requirements

1. `Effect.ensuring` must run the finalizer after the Effect succeeds.
2. `Effect.ensuring` must run the finalizer after the Effect fails with a typed failure.
3. `Effect.ensuring` must return the original success value or the original typed
   failure, and must not replace it.
4. The finalizer must not add a new typed failure to the failure row, or the ticket must
   define what happens when the finalizer fails.
5. The specification must define the order between the finalizer and the local cleanup of
   the Effect.
6. A trap (overflow, divide by zero, or an out-of-bounds access) must bypass the
   finalizer, because traps bypass everything today.
7. This work must amend `bootstrap-flow-functions` through the openspec process.

## Example

```silk
pub effect fn ensuring(
  self: once Effect,
  finalizer: once Effect<() ! never ? S>
) -> A ! E ? R | S
```

```silk
let outcome = run Effect.ensuring(move work, effect { return run closeHandle(&mut handle) })
```

## Out of scope

- A fallible finalizer, unless requirement 4 decides otherwise.
- Interruption. Silk has no interruption today.
- Trap recovery. Traps stay uncatchable.

## Dependencies

- #27. `provideWith` must acquire per execution before the finalizer order is
  settled. #27 has landed, so this dependency is met. The blocker is now #68, which is
[...]

PR #67, opened 2026-08-12
effects: make provideWith acquire a provider for each execution

### LINKED ISSUE #27: effects: make provideWith acquire a provider for each execution
## Context

`Effect.provideWith` acquires a service provider and gives it to an Effect. The current
declaration runs the acquisition once, outside the bound Effect
(`effects.silk:185-192`). A second execution of the same Effect reuses the same provider.
This blocks every retry-with-fresh-resource pattern and every Layer-style composition
(`FEATURE-MATRIX.md:352`).

## Current behavior

The declaration at `effects.silk:185-192` is:

```silk
pub effect fn provideWith<A, C, !E, !F, ?R, ?S>(
  self: once Effect<A ! E ? &mut C | R>,
  acquire: Effect<C ! F ? S>
) -> A ! E | F ? R | S {
  let mut provider = run acquire
  let bound = Intrinsic.bindRequirement(move self, &mut provider)
  return run bound
}
```

`run acquire` happens once, before `bindRequirement`. The `let mut provider` binding then
lives for the whole call. A caller who wraps this Effect in `Effect.retry` gets the same
provider on every attempt. If the first attempt failed because the provider itself is
bad, every later attempt fails the same way.

This is bug-shaped. The `provide` and `provideMut` combinators take a provider that the
caller already holds, so the question does not arise for them. Only `provideWith` owns the
acquisition, and it does it in the wrong place.

## Requirements

1. `provideWith` must run the acquisition Effect one time for each execution of the
   resulting Effect.
2. A retry of a `provideWith` Effect must get a new provider on each attempt.
3. `provideWith` must release each acquired provider before the execution that acquired it
   ends.
4. The failure row and the requirement row of the result must not change.
5. Existing `UserServices` tests must keep passing.

## Example

```silk
// The retry must get a new connection on each attempt.
let outcome = run Effect.retry(
  Effect.provideWith(move query, openConnection()),
  counted(3)
)
```

## Out of scope

- `Layer` composition. This ticket removes the blocker; the Layer design is separate.
- `provide` and `provideMut`. The caller supplies the provider for both.
- A finalizer combinator. See ticket 22.

## Dependencies

- None.

## Acceptance criteria

- [ ] A test runs a `provideWith` Effect twice and asserts the acquisition count is 2.
- [ ] A test wraps `provideWith` in `Effect.retry` and asserts each attempt gets a new
      provider.
- [ ] An allocation acceptance test asserts the acquire count equals the release count
      after 3 attempts.
- [ ] The existing `UserServices` tests pass with no change.

julia-script:
Closes #27

Stacked on #57 (`agent/23-effect-flatten`) — both branches edit `stdlib/silk/effects.silk`. Review/merge #57 first; this PR's diff against `main` will be the union of the two.

## What the investigation found

The issue's "Current behavior" is half right. The cited declaration at `effects.silk:185-192` is accurate, but the behavior claim is not: `run acquire` is inside the effect body, so it already re-ran per execution. Measured on the base branch, before any change:

- running one `provideWith` Effect twice → **2** acquisitions,
- `Effect.retry(_, 2)` over a `provideWith` Effect → **3** acquisitions, one per attempt.

So Requirements 1 and 2 already held. Requirement 3 did not. With a provider that owns a heap block, the allocation trace on the base branch reads:

| scenario | trace |
| --- | --- |
| two successful executions | `Acquire, Release, Acquire, Release` |
| one failing execution | `Acquire` |
| retry, three failing attempts | `Acquire, Acquire, Acquire` |

Every provider acquired by an execution that ended in typed failure leaked.

## Root cause

`fallibleRunSites` (`packages/compiler/src/Ownership.ts:839-852`) classifies a run as able to propagate only when its effect type has `failures.length > 0`. Inside a generic combinator the caller's failures arrive as a row *parameter*: `failures` is empty and `failureParameters` is not. The run is therefore treated as infallible, no `Propagation` exit is published for it, and the owners live at that run get no releases attached — so after specialization substitutes a non-empty row and the run really does fail, they are gone. This is a general hazard for any generic effect body holding an affine owner across a run; `provideWith` is the one combinator in the stdlib that does.

I confirmed the mechanism by bisecting the generic parameters of a user-space replica: `<A, C>` with a concrete failure row releases correctly, `<A, C, !E>` leaks. Lowered MIR for the two is identical apart from the instance's type arguments — the difference is entirely the missing propagation release list on the `RunEffectValue`.

## Note on scope

The direct fix — count an open failure row as fallible — is sound and does fix `provideWith`, but it is not this ticket. It publishes propagation cleanup at generic run sites across the whole codebase and immediately hits an unimplemented backend path, `WasmBackend.ts:1217` "Wasm cleanup does not yet lower hook-bearing union cases". It took the suite from the 3 pre-existing failures to 14, across `LexerPressure`, `StackVmPressure`, `MultiAffineReturn`, `SynchronousEffectCost` and two determinism suites. Making that green means implementing hook-bearing union cleanup in the Wasm backend, which deserves its own ticket and its own design.

**The latent compiler bug is therefore still open and worth filing**: any generic effect body that holds an owner across a fallible run leaks it when the specialized failure row is non-empty. This PR removes `provideWith` from the blast radius; it does not fix the underlying classification.

What this PR does instead is arrange `provideWith` so it never depends on propagation cleanup. The provided Effect is reified into `Result` data *before* the provider is bound, inside a helper frame whose only exit is a normal return:

```silk
effect fn acquireProvider<A, C, !E, !F, ?R, ?S>(...) -> Result<A, Row<!E>> ! F ? R | S {
  let mut provider = run acquire
  let bound = Intrinsic.bindRequirement(result(move self), &mut provider)
  return run bound
}
```

`run bound` is now infallible, so the frame returns normally and the ordinary scope-end release drops the provider. `provideWith` re-raises the failure only after that frame has returned — the provider is already gone, which is exactly the `bootstrap-flow-functions` scenario "Catch outside per-run acquisition: the per-run provider drops before recovery begins". Three other stdlib-only shapes were tried and rejected: an explicit `drop provider` trips OWN0011 against the still-open slice loan, a helper returning `Result` from an already-bound Effect trips `InvalidNormalization`, and moving the provider into the bound Effect (`move provider`) breaks lowering outright.

The acquisition and release points are now stated in the doc comment, which is what #26 (`Effect.ensuring`) needs settled before its finalizer ordering can be defined.

No OpenSpec amendment: the issue carries no `spec-change` label, no acceptance criterion names one, and `openspec/specs/bootstrap-flow-functions/spec.md:84` already requires this behavior — the implementation was out of conformance with a spec that was already correct.

## Acceptance criteria

- [x] A test runs a `provideWith` Effect twice and asserts the acquisition count is 2 — `acquires one provider for each execution of a provideWith Effect`.
- [x] A test wraps `provideWith` in `Effect.retry` and asserts each attempt gets a new provider — `gives every retry attempt its own provider and releases each one in turn`, which asserts the acquire/release events strictly *interleave*, so no attempt can have shared a provider with another.
- [x] An allocation acceptance test asserts the acquire count equals the release count after 3 attempts — `balances acquired and released providers after three attempts`.
- [x] The existing `UserServices` tests pass with no change — untouched, all 8 green.

All three new tests are three-engine acceptance tests following `OwnedAllocationAcceptance`/`DropHookExecution`: the acquire/release trace is asserted on the evaluator, the only engine that publishes one, and Wasm and native clang are held to the result that trace predicts. Verified that 2 of the 3 fail against the unfixed stdlib (3 acquires, 0 releases) and pass with it.

---


julia-script:
Orchestrator note (run-s1k9) — verified: `Closes #27`, 4/4 criteria, base `agent/23-effect-flatten` (correct stack), delta 3 → 3.

**This is the strongest investigation of the run, and it corrects the issue's own premise.** Issue #27's "Current behavior" claims `provideWith` acquires once; measured on the base branch *before any change*, it already acquired per execution — 2 acquisitions for two runs, 3 for a 3-attempt retry. Requirements 1 and 2 already held. The real defect was Requirement 3: every provider acquired by an execution ending in typed failure **leaked** (`Acquire, Acquire, Acquire`, zero releases).

Two things I want to highlight because they are the difference between a fix and a patch:

**The root cause was isolated, not guessed.** `fallibleRunSites` (`Ownership.ts:839-852`) treats a run as fallible only when `failures.length > 0`; in a generic combinator the caller's failures arrive as a row *parameter*, so no `Propagation` exit is published and live owners get no releases. Confirmed by bisecting a user-space replica: concrete failure row releases, row parameter leaks, with lowered MIR identical apart from type arguments.

**The tempting fix was tried, measured, and correctly rejected.** Counting an open failure row as fallible is sound and does fix `provideWith` — but it publishes propagation cleanup repo-wide and hits `WasmBackend.ts:1217` ("Wasm cleanup does not yet lower hook-bearing union cases"), taking the suite from 3 failures to 14. Reverting that and restructuring `provideWith` so it never needs propagation cleanup — reifying to `Result` before binding the provider, inside a helper frame whose only exit is a normal return — is the right call for a ticket scoped to stdlib. Three alternative stdlib-only shapes were tried and rejected with specific diagnostics (`OWN0011`, `InvalidNormalization`, broken lowering), which is the sort of negative evidence that makes a design choice reviewable.

**I have filed the latent compiler bug as #68**, as the PR body recommends. It matters beyond this combinator: any generic effect body holding an affine owner across a fallible run leaks it. In particular **#26 (`Effect.ensuring`) would reintroduce exactly this leak** if landed on the current classification — worth sequencing #68 before #26, which is already `needs-definition` for unrelated reasons.

Also noted for #26: this PR documents `provideWith`'s acquisition and release points — the provider is released when the helper frame returns, strictly before any recovery runs — which is one of the two things #26's finalizer ordering was blocked on.

Two unrelated pre-existing limitations were found and correctly left alone: `Effect.provideMut(&mut x)` trips an `InvalidLoan` MIR violation when run twice, and an acquire Effect capturing a mutable borrow fails inference with `SEM0052` (which is why the new tests count acquisitions via the allocation trace rather than a counter).

---

PR #69, opened 2026-08-12
stdlib: expose allocation metrics as an ordinary Silk value

### LINKED ISSUE #32: stdlib: expose allocation metrics as an ordinary Silk value
## Context

The allocation acceptance tests already assert acquire counts, release counts, and
peak-live counts. That data exists in the test harness only. A Silk program cannot read
it, so no Silk program can report its own memory use or assert its own allocation
behavior.

## Current behavior

A test harness observes allocation counts from outside the program. Silk source has no
way to ask how many allocations are live, how many the program has made in total, or what
the peak was. The `Allocator` service has one operation, `allocate`, and it returns an
`Allocation` and nothing else.

## Requirements

1. The stdlib must declare an `AllocationMetrics` struct.
2. `AllocationMetrics` must hold the acquire count, the release count, and the peak live
   count.
3. `AllocationMetrics` must be Copyable, so a caller can read it without a move.
4. A provider must be able to give its own metrics without compiler privilege.
5. A program that does not read the metrics must pay no cost for them.
6. The counts must use `usize`.

## Example

```silk
pub struct AllocationMetrics {
  acquired: usize
  released: usize
  peakLive: usize
}

pub fn live(self: &AllocationMetrics) -> usize {
  return self.acquired - self.released
}
```

## Out of scope

- Byte totals, as against allocation counts, unless the provider tracks them already.
- Per-call-site attribution and allocation stack traces.
- A compiler-known metrics capability. Metrics stay ordinary Silk.
- Metrics for the Wasm bump allocator release path. See ticket 16.

## Dependencies

- None.

## Acceptance criteria

- [ ] A test makes 3 allocations, drops 1, and asserts the live count is 2.
- [ ] A test asserts the peak live count keeps its value after a drop.
- [ ] A test asserts a program that never reads the metrics compiles with no extra
      allocation.

julia-script:
Closes #32

Stacked on #61 (`agent/39-counted-identity`), which is stacked on #55. Review or merge those first; this branch's diff is the last commit only.

## What ships

`packages/compiler/stdlib/silk/metrics.silk`, a new standard-library module with manifest namespace `AllocationMetrics`:

- `pub struct AllocationMetrics { pub acquired: usize, pub released: usize, pub peakLive: usize }`
- `make()` — the zeroed counts a provider starts from, built from the shared `usize.ZERO` of #39.
- `copy(&self)` — one independent snapshot read through a shared borrow.
- `live(&self)` — `acquired - released`.
- `recordAcquire(&mut self)` — one acquire, raising the peak only when the new live count exceeds it.
- `recordRelease(&mut self)` — one release; the peak keeps the highest live count ever observed.

Nothing in the compiler knows the module. A provider written in ordinary Silk owns an `AllocationMetrics` field, folds its own events into it, and publishes it — the acceptance tests do exactly that with a `CountingAllocator` that wraps `SystemAllocator`.

## Acceptance criteria

- [x] A test makes 3 allocations, drops 1, and asserts the live count is 2 — "counts three acquires and one release as an ordinary Silk value on all three engines".
- [x] A test asserts the peak live count keeps its value after a drop — "keeps the peak live count across drops on all three engines" (peak stays 2 after one drop, after both drops, and after a later acquire below the peak).
- [x] A test asserts a program that never reads the metrics compiles with no extra allocation — "costs a program that never reads the metrics nothing at all": `silk/metrics` is outside the module closure, no `silk/metrics` function reaches MIR, and the allocation trace is exactly one acquire and one release.

A fourth test, "declares metrics as a cleanup-free Copy value", pins Requirement 3 (see below).

## Note on Requirement 3 ("must be Copyable")

Measured on the base branch before implementing: **no** nominal struct in Silk is copyable in the whole-value sense today. `Ownership.categoryOf` (`packages/compiler/src/Ownership.ts:258`) classifies every nominal type as `MoveOnly`, so using any struct value twice by value is `OWN0003`, and moving a struct field out of an enclosing struct is `OWN0002 Struct fields cannot be moved independently`. That is true of `Layout` as well, which `openspec/specs/bootstrap-owned-allocation/spec.md` already specifies as "a Copy value". Making whole structs implicitly copy would be a language-wide ownership change, far outside a stdlib ticket and unmentioned by the issue.

So Requirement 3 is implemented in the sense the language itself defines Copy — `DeclarationIndex.isCopyType`, the predicate that governs `Slot.copy`/`RawBuffer.copy` elements and forbids `impl Drop`:

- all three fields are `usize`, so `AllocationMetrics` satisfies `isCopyType`;
- the test asserts `impl Drop for` a struct holding one is rejected with `SEM0084 Copy type ... cannot implement Drop`, which is the compiler confirming the Copy classification;
- holding and reading metrics produces an empty allocation trace, so it is never an owner with a cleanup obligation;
- "a caller can read it without a move" holds through the API: counts are `pub` and read from a shared borrow, and `copy(&self)` hands out an independent snapshot without consuming the provider's value. The counted test reads the same provider three times and reuses an earlier snapshot afterwards.

## Note on scope

Release counting stays with the caller. A provider only sees `allocate`; a release goes through the allocation's own captured reclaim authority, which no Silk-level provider observes — consistent with the issue putting release-path metrics for the Wasm bump allocator in ticket 16. The tests record the release next to the `drop` that causes it.

Unrelated pre-existing gap found while writing the tests, not touched here: a namespace-qualified stdlib call with a borrow argument is rejected with `SEM0055 A slice borrow is only valid as an immediate ordinary-call argument`. This is not specific to the new module — `Vector.length<i32>(&values)` fails identically on the base branch, while the imported form `length<i32>(&values)` compiles. The tests therefore use `import silk.metrics { ... }`, the same style the existing stdlib acceptance tests use.

No OpenSpec amendment: the issue carries no `spec-change` label and no acceptance criterion names one.

Tests: baseline 3 failures, after 3 failures, +4 new tests

The 3 failures are the pre-existing `packages/compiler/test/Instances.test.ts` golden mismatches (goldens pinned to `aarch64-apple-darwin`, host is `x86_64-unknown-linux-gnu`). Full compiler suite run as `npx vitest run --exclude test/DriverNativeAcceptance.test.ts --maxWorkers=2`: 1015 passed, same 3 failed. `pnpm typecheck` and `pnpm exec biome check .` are clean.

---


julia-script:
Orchestrator note (run-s1k9) — verified: `Closes #32`, 3/3 criteria plus a fourth pinning test, base `agent/39-counted-identity` (correct stack), delta 3 → 3.

**Requirement 3's handling is right, and it exposes a spec/implementation mismatch worth your attention.** The worker measured before implementing: `Ownership.categoryOf` (`Ownership.ts:258`) classifies **every** nominal type as `MoveOnly`, so no struct in Silk is copyable in the whole-value sense — using one twice by value is `OWN0003`, and moving a field out is `OWN0002`. Requirement 3 as literally written ("must be Copyable") is therefore unsatisfiable without a language-wide ownership change no stdlib ticket should make.

The notable part: **that is also true of `Layout`, which `openspec/specs/bootstrap-owned-allocation/spec.md` already describes as "a Copy value."** So the spec asserts a property of an existing type that the implementation does not provide. Not this PR's problem, and I have not filed it since it may be intentional shorthand for the `isCopyType` notion — but if the spec means whole-value copy, it is wrong about `Layout` today.

Implementing Requirement 3 against `DeclarationIndex.isCopyType` (the predicate that governs `Slot.copy`/`RawBuffer.copy` elements and forbids `impl Drop`) is the right reading, and the fourth test pins it properly by asserting `impl Drop` on a holder is rejected with `SEM0084` — that is the compiler itself confirming the classification, rather than the test asserting its own assumption. Holding metrics produces an empty allocation trace, so there is genuinely no cleanup obligation.

**I have filed the `SEM0055` gap you found as #70**, with an independent measurement confirming it (`Vector.length<i32>(&values)` → `SEM0055`, imported `length<i32>(&values)` → clean). It was reported independently by the #33 worker too, so two agents tripped over it during unrelated work. It matters more than it looks: it means the namespace feature #33/PR #55 adds cannot reach any `&self`-style stdlib operation through the qualified spelling, which is why your tests use the import form.

The third acceptance criterion is the one I would have most expected to be fudged, and it is not — it asserts `silk/metrics` is outside the module closure, that no `silk/metrics` function reaches MIR, *and* that the allocation trace is exactly one acquire and one release. That is a real zero-cost proof rather than a compile check.

---

PR #71, opened 2026-08-12
operators: Add bitwise operator spellings for &amp; | ^ ~

### LINKED ISSUE #6: operators: Add bitwise operator spellings for & | ^ ~
## Context

Silk has bitwise operations, but only as named functions. The integer modules expose
`bitAnd`, `bitOr`, `bitXor`, and `bitNot` (`packages/compiler/src/Scalar.ts:268-273`).
The specification requires these operations: "Every integer SHALL expose bitwise
operations, shifts, and rotates" (`openspec/specs/bootstrap-integer-scalars/spec.md:73`).
Only the operator spelling is absent. Code that does bit work is therefore hard to read.

## Current behavior

A user must call the named function:

```silk
crc = crc |> u32.bitXor(3988292384)
```

The operator table has 11 entries and none of them is bitwise
(`packages/compiler/src/Operator.ts:41-103`).

Two of the four bytes already lex. `&` and `|` are single punctuation tokens
(`packages/compiler/src/Lexer.ts:57-58`). **`^` (0x5e) and `~` (0x7e) are not in the
lexer byte vocabulary at all.** Today they reach the invalid-byte path and produce
`LEX0001`.

## Requirements

1. The lexer must accept `^` and `~` as punctuation tokens.
2. The operator table must map `&`, `|`, `^` to the existing named operations.
3. The operator table must map prefix `~` to the existing `bitNot` operation.
4. Each operator must resolve to the same actor operation that the named function uses.
5. The operators must keep the homogeneous rule: both operands must have the same type.
6. The precedence must sit below the equality operators and above the pipeline operator.
7. The change must not add a new intrinsic.

## Example

```silk
pub fn checksum(value: u32, mask: u32) -> u32 {
  let masked = value & mask
  let flipped = ~masked
  return flipped ^ mask
}
```

## Out of scope

- **The shift operators `<<` and `>>`.** These need a parser decision, because `>>`
  collides with the end of a nested generic argument list, as in `Vector<Vector<i32>>`.
  A separate ticket must cover the shifts.
- The rotate operations. These stay named functions.
- Any change to the operations themselves. This ticket adds spellings only.

## Dependencies

None.

## Acceptance criteria

- [ ] The lexer produces tokens for `^` and `~` instead of `LEX0001`.
- [ ] A lexer test shows all four bytes as punctuation tokens.
- [ ] The four operators compile to the same result as the named functions.
- [ ] A test shows that `a & b` and `u32.bitAnd(a, b)` give the same value.
- [ ] Mixed operand types produce a diagnostic, as the named functions do.
- [ ] `openspec/specs/bootstrap-operator-semantics/spec.md:22-35` gets an amendment,
      because the operator surface is closed and ordered.

julia-script:
Closes #6

**Stacked on #59** (`agent/5-else-if`), which is itself stacked on #58 → #53. Review/merge in that order; this PR's base is `agent/5-else-if`, not `main`.

## What changed

`&`, `|`, `^`, and prefix `~` now spell the integer operations that were previously reachable only through `u32.bitAnd`, `bitOr`, `bitXor`, and `bitNot`. No new intrinsic: the operator table points at the operations that already exist in `Scalar.ts`.

- `Lexer.ts` / `Token.ts` — `^` (0x5e) and `~` (0x7e) join the punctuation vocabulary as `Caret` and `Tilde`. Both previously reached the invalid-byte path and produced `LEX0001`. `&` and `|` already lexed.
- `Operator.ts` — `Ampersand`/`Caret`/`Pipe` map to `BitAnd`/`BitXor`/`BitOr` at precedences 37/35/33, left-associative; `Tilde` maps to prefix `BitNot`.
- `Parser.ts` — `~` joins `-` and `!` as a prefix-expression start.
- Spec amendments in `openspec/specs/bootstrap-operator-semantics/spec.md` and `openspec/specs/bootstrap-lexer/spec.md`.

## Precedence

**This deviates from issue #6's Requirement 6 by maintainer decision.** Requirement 6 asks for one bitwise level "below equality, above the pipeline" and does not order the three operators against each other. The maintainer decided to adopt the Rust model instead: three distinct levels, all **above** comparison.

The three infix spellings get three left-associative levels between additive (40) and relational (30) — `&` = 37, `^` = 35, `|` = 33 — so `&` binds tighter than `^`, which binds tighter than `|`, and all three bind tighter than every comparison. Rust orders them exactly this way, and Go, Python, and Swift also place bitwise operators above comparison. The gap above `&` leaves room for `<<` and `>>` when they gain spellings; they remain out of scope here. Prefix `~` is unaffected. Consequences, all pinned by tests and written into the spec:

- `a & b == c` groups as `(a & b) == c` — both comparison operands are `u32`, and it compiles. This avoids the C precedence gotcha rather than reproducing it.
- `a | b & c` groups as `a | (b & c)`, matching C, Rust, Go, Python, and Swift.
- `a | b ^ c & d` groups as `a | (b ^ (c & d))`.
- `a & b |> f` groups as `(a & b) |> f`, unchanged — the pipeline still sits below everything.

`Parser.ts` needed no change for the tiering: it is a precedence-climbing parser driven entirely by the `Operator.ts` table.

## Note on scope

Two changes go past a literal reading of the Requirements, both to avoid accepted source that miscompiles or crashes:

1. **Non-integer actor fallback.** `Operator.target` previously assumed every scalar actor carries every operator's operation. Floats carry no `bitAnd`, so `a & b` over `f64` would have looked up `f64.bitAnd`, found nothing, and thrown `RangeError: Compiler operator table is inconsistent` out of analysis. Bitwise operators now fall back to the default integer when the actor is not an integer, which produces the same `SEM0012` the named function produces. Covered by a test.
2. **Silk-language lexer fixture.** `examples/language-pressure/lexer/main.silk` reimplements the lexer in Silk and is diffed against the canonical one by `LexerPressure.test.ts`. It needed the two new bytes or the corpus fingerprints diverged.

Three parser fixtures used `^` as their canonical unsupported byte (`'pub fn ^ main()...'`, `'^^^'`, `'^^'` between declarations). Since `^` is now valid, they were switched to `$`, which keeps each test exercising invalid-byte recovery as written. Editor surfaces (`CodeMirror.ts`, `TextMate.ts`, and the generated `packages/vscode` grammar) classify the new spellings as operators.

## Acceptance criteria

- [x] The lexer produces tokens for `^` and `~` instead of `LEX0001`.
- [x] A lexer test shows all four bytes as punctuation tokens.
- [x] The four operators compile to the same result as the named functions — three-engine parity (interpreter, Wasm, native clang) in `BitwiseOperatorAcceptance.test.ts`.
- [x] A test shows that `a & b` and `u32.bitAnd(a, b)` give the same value.
- [x] Mixed operand types produce a diagnostic, as the named functions do — the test asserts the operator's diagnostic codes equal the named function's on the same operand pair.
- [x] `openspec/specs/bootstrap-operator-semantics/spec.md` gets an amendment. The `bootstrap-lexer` spec's operator-vocabulary requirement was amended too, since it enumerates exactly which bytes the lexer recognizes and would otherwise be false.
- [x] Requirement 6's ordering is superseded by the maintainer decision above; the spec states the new ordering rather than the one the ticket asked for.

Out of scope per the ticket and untouched here: `<<` and `>>`, the rotates.

Tests: baseline 3 failures, after 3 failures, +12 new tests

The 3 are the pre-existing host-triple golden mismatches in `Instances.test.ts` (goldens target `aarch64-apple-darwin`, host is `x86_64-unknown-linux-gnu`). No golden or fixture moved as a result of the precedence change. `pnpm typecheck` and `pnpm exec biome check .` are clean.

---

### DIFF
--- a/examples/language-pressure/lexer/main.silk
+++ b/examples/language-pressure/lexer/main.silk
@@ -11,6 +11,8 @@ const tokenConst: u8 = 67
 const tokenInvalidStaticLiteral: u8 = 68
 const tokenService: u8 = 69
 const tokenInterface: u8 = 70
+const tokenCaret: u8 = 71
+const tokenTilde: u8 = 72
 
 struct Token {
   kind: u8
@@ -114,6 +116,8 @@ fn isPunctuation(byte: u8) -> bool {
   if byte == 62 { return true }
   if byte == 124 { return true }
   if byte == 38 { return true }
+  if byte == 94 { return true }
+  if byte == 126 { return true }
   return byte == 46
 }
 
@@ -204,6 +208,8 @@ fn punctuationKind(byte: u8) -> u8 {
   if byte == 62 { return 57 }
   if byte == 124 { return 59 }
   if byte == 38 { return 61 }
+  if byte == 94 { return tokenCaret }
+  if byte == 126 { return tokenTilde }
   if byte == 46 { return 62 }
   return 65
 }
--- a/openspec/specs/bootstrap-lexer/spec.md
+++ b/openspec/specs/bootstrap-lexer/spec.md
@@ -154,10 +154,10 @@ spelling SHALL remain one identifier token.
 
 ### Requirement: Expression operators use deterministic longest tokens
 
-The lexer SHALL recognize `+`, `-`, `*`, `/`, `%`, `!`, `<`, `<=`, `>`, `>=`, `==`, `!=`, and
-`|>` as distinct operator tokens with exact source spans. Longest recognition SHALL prefer `->`
-over `-`, `<=` over `<`, `>=` over `>`, `==` over `=`, `!=` over `!`, and `|>` over unsupported
-prefix fragments. `//` SHALL continue to begin a line comment while a single `/` SHALL be the
+The lexer SHALL recognize `+`, `-`, `*`, `/`, `%`, `!`, `<`, `<=`, `>`, `>=`, `==`, `!=`, `&`,
+`|`, `^`, `~`, and `|>` as distinct operator tokens with exact source spans. Longest recognition
+SHALL prefer `->` over `-`, `<=` over `<`, `>=` over `>`, `==` over `=`, `!=` over `!`, and `|>`
+over a bare `|`. `//` SHALL continue to begin a line comment while a single `/` SHALL be the
 division token. Operator recognition SHALL preserve the existing lossless coverage and invalid-byte
 recovery guarantees.
 
@@ -165,6 +165,11 @@ recovery guarantees.
 
 - **WHEN** source contains the complete operator vocabulary separated by trivia
 - **THEN** each spelling produces one supported token with its exact span and source slice
+
+#### Scenario: Lex the bitwise operator bytes
+
+- **WHEN** source contains `& | ^ ~ |>`
+- **THEN** the four bitwise spellings are punctuation tokens, `|>` still wins over a bare `|`, and no unsupported-byte diagnostic is reported
 
 #### Scenario: Prefer comments over division pairs
 
--- a/openspec/specs/bootstrap-operator-semantics/spec.md
+++ b/openspec/specs/bootstrap-operator-semantics/spec.md
@@ -7,7 +7,7 @@ data-first behavior while giving source programs conventional precedence and pip
 ## Requirements
 ### Requirement: Operators resolve homogeneously across integers
 
-Arithmetic and comparison operators SHALL resolve only for compatible operands of the same integer type. They SHALL select that type's signed or unsigned checked semantics without implicit conversion, overload lookup, truthiness, or operand reordering. Prefix negation SHALL support signed integers only; logical negation SHALL support `bool` only.
+Arithmetic, bitwise, and comparison operators SHALL resolve only for compatible operands of the same integer type. They SHALL select that type's signed or unsigned checked semantics without implicit conversion, overload lookup, truthiness, or operand reordering. Prefix negation SHALL support signed integers only; logical negation SHALL support `bool` only; bitwise complement SHALL support integers only. Each bitwise operator SHALL select exactly the operation its named counterpart selects — `&` selects `bitAnd`, `|` selects `bitOr`, `^` selects `bitXor`, and prefix `~` selects `bitNot` — without introducing an intrinsic of its own.
 
 #### Scenario: Resolve unsigned multiplication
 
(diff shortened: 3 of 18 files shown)


julia-script:
Orchestrator note (run-s1k9) — verified: `Closes #6`, 6/6 criteria, base `agent/5-else-if` (correct stack), delta 3 → 3, with `compiler-cli` (3) and `wasm` (1) also re-run and matching their pre-existing counts.

## A crash was averted, and it is the most valuable part of this PR

`Operator.target` assumed every scalar actor carries every operator's operation. Floats have no `bitAnd`, so `a & b` over `f64` would have looked up `f64.bitAnd`, found nothing, and thrown **`RangeError: Compiler operator table is inconsistent` out of analysis** — a compiler crash on ordinary bad user code, not a diagnostic. The non-integer fallback makes it report the same `SEM0012` the named function does, and it is covered by a test.

This is the pattern I briefed for and it keeps paying: a literal reading of the Requirements ("add operator spellings for functions that already exist") would have shipped that crash, because the spellings work fine on the type the ticket was thinking about.

## Precedence needs your ruling

Requirement 6 constrains only one thing — below equality, above pipeline — so all three infix spellings share **one** precedence level (15), left-associative. The consequence:

```silk
a | b & c    // groups as (a | b) & c
```

**C would give `a | (b & c)`**, since C tiers them `& > ^ > |`. The ticket does not order the three against each other, and I would rather an agent leave that unordered and say so than invent a tier silently — but a developer arriving from C, Rust, Go, or Java will read `a | b & c` and expect the C grouping. Getting it wrong is silent: the program compiles and computes something else.

`a & b == c` grouping as `a & (b == c)` is the *intended* C gotcha and is what the Requirements ask for — that one is fine, and it produces `SEM0012` rather than a wrong answer.

If you want C ordering it is a one-line change to three precedence numbers plus the spec sentence, and the tests that pin the current grouping would need updating. **Worth deciding before this merges**, since changing operator precedence after code exists is a silent-breakage migration.

## Collateral worth knowing about

Three parser fixtures used `^` as their canonical *unsupported byte* (`'pub fn ^ main()'`, `'^^^'`), and two CodeMirror tests used `let ^`. Since `^` is now valid, they were switched to `$` so they keep testing invalid-byte recovery rather than silently testing nothing. That is exactly the failure mode I have seen twice elsewhere in this run — a test that still passes while proving nothing — and it was handled correctly here.

`examples/language-pressure/lexer/main.silk` reimplements the lexer in Silk and is fingerprint-diffed by `LexerPressure.test.ts`, so it had to learn the two new bytes. The generated `packages/vscode` grammar was regenerated via the sync script rather than hand-edited.

No interaction problems with the rest of the syntax stack: `0xff ^ 0x0f`, `~0b1010`, and `1_0 ^ 2` all lex correctly against #53's base prefixes and #58's separators, with a test covering it.

---

PR #77, opened 2026-08-12
docs: Write the Silk language reference, tutorial, and stdlib docs

### LINKED ISSUE #52: docs: Write the Silk language reference, tutorial, and stdlib docs
## Context

This is the largest gap in the project. `apps/docs/content/` holds 28 files, and
every one of them documents the LLVM builder API. The `content/language` path is a
symlink to `packages/language/docs/`, and that directory holds one README about the
editor package. Silk has no language reference, no tutorial, and no rendered stdlib
API documentation. The raw material is strong: 92 specs, 638 catalogued intrinsics,
117 diagnostic codes in `packages/compiler/src/Diagnostic.ts`, 10 working examples,
and 759 `///` comments across 27 stdlib modules. The `silk doc` command already
emits JSON from those comments. The gap is a writing gap, not a knowledge gap.

## Current behavior

A new user reads `README.md`, `VISION.md`, and `AGENTS.md`. All three address a
contributor, not a user. The closest text to user documentation is
`examples/algorithms/README.md`. A user who wants the meaning of `SEM0013` reads
`Diagnostic.ts`. A user who wants the signature of a stdlib function opens the
`.silk` file in `packages/compiler/stdlib/silk/`.

## Requirements

**These three documents come first, in this order.** Each one answers a different
question and each one has a different source.

1. The getting-started tutorial must come first, because it makes the language usable.
2. The tutorial must take a user from `silk init` to a program that runs.
3. The tutorial must build on `examples/algorithms`, because those examples compile today.
4. The language reference must come second, because it makes the language complete.
5. The language reference must cover the lexical form, the type system, the memory and ownership rules, and the effect system.
6. The language reference must take its rules from the 92 specs in `openspec/specs/`.
7. The stdlib API documentation must come third, because it is generated and needs the least writing.
8. The stdlib API documentation must come from the `silk doc` JSON and the 759 existing `///` comments.
9. A diagnostic index must give the meaning of each of the 117 codes with one example.
10. Every Silk code block in the tutorial and in the reference must compile with the current compiler.
11. The documents must live under `packages/language/docs/`, because `apps/docs/content/language` points there.

## Example

```silk
// From the tutorial. The first program a user writes.
pub fn main() -> i32 {
  let total = 2 + 3
  total
}
```

## Out of scope

- An HTML site, doctests, and a search index. Ticket 47 covers that work.
- Documentation for the LLVM builder package. That documentation exists.
- A change to the language. This ticket writes about the language as it is today.
- Documentation of an absent feature. The feature matrix records those gaps.

## Dependencies

- None. Every source the three documents need exists today.

## Acceptance criteria

- [ ] A getting-started tutorial exists and takes a user from `silk init` to a program that runs.
- [ ] A language reference exists and covers the lexical form, the types, the memory rules, and the effects.
- [ ] The stdlib API documentation covers 27 modules and comes from the `silk doc` JSON.
- [ ] A diagnostic index lists all 117 codes with a short meaning for each one.
- [ ] A test compiles every Silk code block in the tutorial and in the reference.
- [ ] `apps/docs/content/language` shows the new documents through its symlink.
- [ ] `result.silk` gets `///` comments, because it has none today.

julia-script:
Closes #52

Silk had no user-facing documentation. Every one of the 28 files under `apps/docs/content/` documented the LLVM builder API, and `packages/language/docs/` — the directory `apps/docs/content/language` symlinks to — held a single README about the editor package. This adds the four documents a Silk user needs, all of them under that symlink so the docs site picks them up automatically.

## What's here

- **`tutorial.md`** — from `silk init` to a program that runs. Introduces explicit `move`, borrows and slices, unions with `match`, and the effect system, in that order, building on the forms `examples/algorithms/` already uses.
- **`reference.md`** — the rules: lexical form, the seven declaration forms, the type system, memory and ownership, effects, and control flow, plus an explicit list of what the language deliberately does not have.
- **`stdlib.md`** — all 28 modules, generated from the same documentation model `silk doc` emits, so every signature and description is the `///` comment on the declaration.
- **`diagnostics.md`** — all 119 diagnostic codes grouped by phase, each with a meaning and the message the compiler actually prints.

`scripts/generate-documentation.mjs` regenerates the two generated pages, with a `--check` mode (verified to fail on drift) for CI.

## Every example is compiled, and that mattered

`test/DocumentationExamples.test.ts` extracts every fenced Silk block from the prose documents and compiles it against the current compiler. It caught four examples that did not compile — **including the one in the issue body**, which omits the trailing `return` that Silk requires (`PAR0004`). Also caught: a loop index must be seeded at `usize` (`SEM0033`, the `usize.add(0, 0)` idiom), and passing a struct by value needs an explicit `move` (`OWN0003`). The pre-existing README example was missing its trailing `return` too, and is fixed here.

The test also asserts that every stdlib module and every declared diagnostic code appears in the generated pages, so a new module or code cannot land undocumented.

## Acceptance criteria

- [x] A getting-started tutorial exists and takes a user from `silk init` to a program that runs — verified end to end: `silk init` → edit → `silk check` → `silk run` all succeed against the documented scaffold.
- [x] A language reference exists and covers the lexical form, the types, the memory rules, and the effects.
- [x] The stdlib API documentation covers 27 modules and comes from the `silk doc` JSON — covers all **28** modules currently in `stdlib/silk/` (the issue's count was slightly stale), built from the `@silk-effect/documentation` project model that backs `silk doc`.
- [x] A diagnostic index lists all 117 codes with a short meaning for each one — lists all **119** codes now declared in `Diagnostic.ts`; zero rows are blank.
- [x] A test compiles every Silk code block in the tutorial and in the reference.
- [x] `apps/docs/content/language` shows the new documents through its symlink — the docs build goes from 57 to 61 static pages, emitting `/docs/language/{tutorial,reference,stdlib,diagnostics}`.
- [x] `result.silk` gets `///` comments, because it has none today.

## Note for reviewers

Writing §2.5 surfaced a real limitation worth a decision: an `interface` conformance only resolves targets spelled `Intrinsic.*` (`DeclarationIndex.ts` ~3170-3215), so an ordinary user function mapped into an `impl` is always rejected with `SEM0083`. Interfaces therefore abstract over compiler primitives but cannot let a user type satisfy a user-declared contract. The reference documents the behavior as it stands; if that's a gap rather than an intentional bootstrap limit, it deserves its own issue.

Tests: baseline 0 failures (measured in step 0), after 0 failures, +21 new tests


julia-script:

Verified independently: diff matches the body with no scope creep; 28 stdlib modules and 119 diagnostic codes both confirmed against source (both of the issue's counts were stale, as the PR says); the `SEM0083` interface limitation in the "Note for reviewers" is **correct** — `DeclarationIndex.ts:3172-3178` resolves a mapping target only when it is a two-segment `TypePath` whose first segment is literally `Intrinsic`, and `:3198-3206` turns anything else into `Diagnostic.invalidConformance`. Worth keeping as written.

The compile-test for fenced examples is the right call and already earned its keep. But it cannot catch prose that is confidently wrong, and there are **two WRONG claims — the same root error, both contradicted by this PR's own examples.**

### 1. `reference.md:341` — implicit moves

> "Handing a value away is written `move`, and there is no implicit move — even for copyable types, a `move` consumes the binding."

This is backwards. `Ownership.ts:431` raises `OWN0003` **only** when `binding.category._tag === 'MoveOnly'`. `categoryOf` (`Ownership.ts:258-281`) classifies as `Copyable` — and therefore exempt from any `move` requirement — all builtins (every scalar and `bool`), strings, `never`, `OutOfMemory`, shared slices, shared references, shared effects, shared callables, and fixed arrays of copyable elements. A copyable binding is reused freely and never triggers `OWN0001`.

The reference's own example at `reference.md:328` calls `identity<i32>(held.value)` with no `move`. Under the stated rule that would be an error.

### 2. `tutorial.md:10` — same claim, higher cost

> "A value is used at most once, and handing it away is written with the `move` keyword. Nothing moves implicitly."

Same contradiction, more damaging: it is a new reader's **first** statement about ownership, and it is false for exactly the copyable set a beginner's early programs are made of — `i32` arithmetic, `bool` conditions, loop counters. `tutorial.md:86-98` reuses `index` and `total` across loop iterations, which the stated rule forbids. A reader who believes this will expect `move` on every `let total = 2 + 3` reuse.

**Suggested fix for both:** scope the rule to move-only types — "a value of a move-only type is used at most once; scalars, `bool`, strings, and shared borrows are copyable and are reused freely."

### Minor (IMPRECISE, not blocking)

`reference.md:200-201` calls the `Drop` hook "the one exception" that has a body. Accurate for `interface`, but the surrounding prose covers `impl` generally, where the service-conformance path (`DeclarationIndex.ts:3225+`) rejects hook bodies with a distinct message.

Blocking on the two WRONG items only. Everything else in the semantic review checked out — keyword list (28, exact), the precedence table including the three bitwise levels and non-associative comparisons, escape set, digit-separator rules, callable substitution direction, union normalization, guarded-arm exhaustiveness, and the "exactly two compiler primitives" seam.

julia-script:
## Review addendum — run-b4c2 (corrects my finding 1, adds a third defect)

Deeper verification against the specs refines the above. **My finding 1 was overstated** — correcting it before anyone acts on it.

### Correction to finding 1

The sentence conflates two sub-claims, and only one is wrong:

- **"a `move` consumes the binding even for copyable types" is CORRECT** — `Ownership.ts:13` states exactly this, and `checkUse` (`Ownership.ts:333-351`) deletes the key from `live` in the consuming branch with **no** category test. Spec agrees at `bootstrap-ownership/spec.md:82-83`.
- **"there is no implicit move" is WRONG for copyable types** — `move` is not *required* for them. `OWN0003` is gated on `MoveOnly` (`Ownership.ts:430-435`), so passing an `i32` into a consuming position silently copies. `bootstrap-ownership/spec.md:100-103` confirms this is intentional: reads before a consuming move "copy rather than consume".

So the fix is narrower than I first wrote: keep the "`move` consumes even copyable bindings" half, and scope only the *requirement* to move-only types. `tutorial.md:10`'s "Nothing moves implicitly" is fine as literal prose — the misleading part is the adjacent implication that `move` is needed for `i32`.

### Third defect — `reference.md` borrow prose over-generalizes (WRONG)

If the reference says borrowing "a subplace" is rejected, that is too broad. `&thing.field` **is** a supported value borrow: `borrowRoot` (`Elaboration.ts`) recurses through `FieldProjection` accumulating a field path, `exclusiveBorrowRoot` has a dedicated `root.path.length > 0` clause for exclusive projection, and `bootstrap-runtime-slices/spec.md:156-161` is a first-class requirement for it ("value borrows preserve stable field projections… MAY project through resolved nominal fields"). The real prohibition is narrower — **array/slice** subplaces only (`bootstrap-runtime-slices/spec.md:34`).

### Fourth defect — `Effect.catch` is documented as spec, not as shipped (WRONG)

Any prose describing `Effect.catch<E>` as selecting one member, removing it once coverage is complete, and passing the rest through describes **`bootstrap-flow-functions/spec.md:47-49`, not the code**. What ships in `effects.silk:119-127` is an unconditional alias for `catchAll`:

```silk
pub effect fn catch<A, !E, !F, ?R, ?S>(
  self: once Effect<A ! E ? R>,
  onFailure: once fn(Row<!E>) -> Effect<A ! F ? S>
) -> A ! F ? R | S {
  return run catchAll(move self, move onFailure)
}
```

The `E` there is the protected row, not a selected member — `Effect.catch<E>(handler)` cannot be written. The whole row `!E` is discarded and replaced by `!F`; there is no residual row. Its own doc comment at `effects.silk:118` reads "Recovers **every** typed failure" — **identical** to `catchAll`'s at `:112`. Every call site in the repo passes two positional args with no type argument (`Elaboration.test.ts:137`, `BootstrapEvaluation.test.ts:318,386,477`, `Ownership.test.ts:608,627`, `FileSystemAcceptance.test.ts:339-341`), and `SEM0067` (`invalidEffectHandler`, whose message is literally `Invalid Effect.catch handler:`) is used only for `Intrinsic.effectResult` arity checking, never for member coverage.

Document the current whole-row behavior. **The implementation gap is worth its own issue** — the spec requires selective catch and the stdlib does not provide it.

### Everything else verified clean

Path-insensitive moves, partial-move rejection, the four borrow-conflict/reborrow rules, whole-array borrow, `&mut` mutable-root requirement, no implicit decay, `drop` consuming early, `OWN0011` on dropping a borrowed owner, Drop-hook restrictions (`SEM0084`), cleanup ordering (last-acquired-first-released via the double reverse in `frameSitesInnerFirst`), arm-local release boundaries, per-return exits, the `effect { }` block form (real and plumbed to `Lower.ts:1520`), `run` extent, `fail` semantics, the exactly-two-primitives seam (`Intrinsic.ts:697-718`), all ten named effect functions, `provide`/`provideMut`/`provideWith` single-entry removal, and every entry-point rule in `Instances.ts:333-393` — all confirmed against source and spec.

julia-script:
## Review round 2 — orchestrator run-b4c2: all three fixes verified, approving

Confirmed on pushed HEAD `9204e58`: **`validate` and `webcontainer-browser` both pass, mergeable state clean.** CI on a clean machine is the authoritative signal here — my local host has been running ~40 vitest workers across agent worktrees, which made local timing results unreliable for every PR in this run.

All three review items are correctly resolved:

1. **Ownership** — the two conflated rules are now separated exactly right: `move` is *required* only for move-only types (`OWN0003` gated on `MoveOnly`), while writing `move` *consumes* in both categories (`checkUse` deletes from `live` with no category test). Adding the `categoryOf` table and a compiled example that reuses a copyable binding twice is better than the minimal fix I asked for.
2. **Borrows** — subplace claim correctly scoped to array subplaces, with a compiled `&mut outer.inner` example demonstrating the supported case.
3. **`Effect.catch`** — documented as shipped, with the spec gap noted but not implemented. Correct call; the gap is tracked at #79.

### The new finding is the valuable one

`SEM0041` (`matchScrutineeNotNominalCode`, `Diagnostic.ts:119`) — verified present — means matching a scalar or a borrow of one is rejected. That made the tutorial's "use `match` to select a value" advice **actively wrong for precisely the case a reader reaches for**, since Silk has no `if` expression. A new user following it would hit `SEM0041` on their first attempt at conditional assignment. Catching that during a self-directed audit, rather than waiting to be told, is the difference between docs that compile and docs that are true.

Also useful: the confirmation that `SEM0083` restricts only the **interface** path, while the **service** path does accept user-defined functions. That makes §2.5's original wording correct as written — "interface" is right where the more natural "impl" would have been wrong.

### Accepted, with one standing follow-up

The generator's `--check` mode is still not wired into CI. Everything in `stdlib.md` and `diagnostics.md` is generated, so those two pages can silently drift the moment a module or diagnostic code is added. Wiring `documentation:check` into the compiler `test` script is a small change that converts a documentation-rot risk into a build failure — worth doing, but not a reason to hold this PR.

Nothing blocking. This is ready to merge from my side; the merge itself is the maintainer's call.

PR #80, opened 2026-08-12
lexer: A malformed float exponent is silently accepted and compiles

### LINKED ISSUE #63: lexer: A malformed float exponent is silently accepted and compiles
Found by an adversarial review pass during orchestrator run run-s1k9 and verified on clean `main` (`49bfeb4`). **Pre-existing — not introduced by any open PR.**

## Current behavior

A float literal with an `e` and no exponent digits lexes cleanly and compiles with no diagnostic at all.

`Lexer.lex` on clean `main`:

```
1e    => [DecimalFloat, EndOfFile]  diagnostics=0
1e+   => [DecimalFloat, EndOfFile]  diagnostics=0
1e-   => [DecimalFloat, EndOfFile]  diagnostics=0
1E    => [DecimalFloat, EndOfFile]  diagnostics=0
1.5e  => [DecimalFloat, EndOfFile]  diagnostics=0
```

And end to end, via `Analysis.ofSourceRealized`:

```silk
pub fn main() -> f64 {
  return 1e
}
```

reports **zero diagnostics**.

## Why it compiles

The three stages disagree, and nothing reports the disagreement:

1. `Lexer.ts:389-396` accepts the exponent marker without requiring a digit after it, so `1e` becomes one `DecimalFloat` token.
2. `FloatingPoint.fromDecimal` (`FloatingPoint.ts:49-51`) matches against `/^([+-]?)(\d+)(?:\.(\d+))?(?:[eE]([+-]?\d+))?$/`, which is anchored and requires at least one digit after `[eE]`. `"1e"` does not match, so it returns `undefined`.
3. `Elaboration.ts:1231-1234` maps that `undefined` to an `Unavailable` fact — **and emits no diagnostic**. The malformed literal simply disappears.

So the literal is neither rejected by the lexer nor reported by elaboration, and the program is accepted.

## Why this matters

Silence is the wrong outcome for input no stage can represent. The contrast is visible inside the current backlog: PR #58 (digit separators) correctly makes `1e_` produce one `Invalid` token plus a `LEX0005` diagnostic, while `1e` — no less malformed — stays silently valid. A user who typos an exponent gets no feedback from the compiler.

## Suggested direction

Either reject at the lexer (an exponent marker must be followed by at least one digit, with a `LEX00xx` diagnostic and the same `Invalid`-token recovery shape `LEX0004`/`LEX0005` use), or have `Elaboration.ts:1231` report a diagnostic when `fromDecimal` returns `undefined` rather than silently producing `Unavailable`. The lexer is the better place — it keeps the error at the token that caused it.

The second change is worth making regardless: an `Unavailable` float fact with no diagnostic is a silent-acceptance hazard for any other input that fails conversion.

## Acceptance criteria

- [ ] `1e`, `1e+`, `1e-`, `1E`, and `1.5e` each produce exactly one lexical diagnostic.
- [ ] `pub fn main() -> f64 { return 1e }` fails to compile.
- [ ] Well-formed exponents (`1e5`, `1e+5`, `1e-5`, `1.5e10`) are unaffected.
- [ ] A test asserts that a float spelling `fromDecimal` cannot convert never reaches an `Unavailable` fact without a diagnostic.

julia-script:
Closes #63

A float literal with an exponent marker and no exponent digits lexed as one `DecimalFloat` token with no diagnostic, and elaboration then mapped the unconvertible spelling to an `Unavailable` fact silently — so `1e` compiled with zero diagnostics.

Fixed at the lexer, as the issue recommends, keeping the error on the token that caused it:

- `Lexer.ts` requires at least one digit after `e`/`E` and its optional sign. Otherwise the literal becomes one `Invalid` token plus a new `LEX0006`, matching the `LEX0004`/`LEX0005` recovery shape. The digit-separator and exponent diagnostics are mutually exclusive, so `1e_` still reports exactly one `LEX0005`.
- `Elaboration.ts` — the second change the issue asks for regardless: `analyzeFloating` now returns `{ fact, diagnostics }` and reports a new `SEM0095` when `fromDecimal` cannot convert, so no float fact can reach `Unavailable` in silence.
- `examples/language-pressure/lexer/main.silk` — the Silk-implemented lexer kept in parity with the canonical one, caught by the differential `LexerPressure` corpus test.

### Acceptance criteria

- [x] `1e`, `1e+`, `1e-`, `1E`, and `1.5e` each produce exactly one lexical diagnostic.
- [x] `pub fn main() -> f64 { return 1e }` fails to compile.
- [x] Well-formed exponents (`1e5`, `1e+5`, `1e-5`, `1.5e10`) are unaffected.
- [x] A test asserts that a float spelling `fromDecimal` cannot convert never reaches an `Unavailable` fact without a diagnostic. Asserted as the invariant that makes it unreachable: `Lexer.test.ts` sweeps 92 whole/fraction/exponent/separator combinations and asserts every spelling accepted as a `DecimalFloat` with no diagnostic is convertible by `fromDecimal`. The `SEM0095` guard remains as the backstop for any future divergence.

### Note for review

`Parser.test.ts` had an assertion that `1.25e-` produces a `DecimalFloat` token inside a `FloatingLiteralExpression` — that test encoded exactly the silent acceptance this issue reports as a bug. I kept its real guarantee (lossless byte round-tripping through parser recovery) and updated it to assert the new invalid-token shape.

Tests: baseline 4 failures (measured in step 0), after 7 failures, +5 new tests

Baseline and post-change failures are all host-environment timeouts, and the set is flaky between runs. Every failing test name observed after the change (`Logging`, `OsFileSystem` x2, the `LexerPressure`/`StackVmPressure` allocation-rollback tests, `IntegerScalars`, `OccurrencePerformance`) was reproduced failing on unmodified `origin/main` with the same timeout errors, or passes in isolation. The `LexerPressure` corpus test that this change does touch passes. `pnpm typecheck` is green repo-wide.

### DIFF
--- a/examples/language-pressure/lexer/main.silk
+++ b/examples/language-pressure/lexer/main.silk
@@ -327,6 +327,7 @@ fn scanNumber(source: &[u8], start: usize) -> Decision {
       }
     }
   }
+  let mut exponentDigits = true
   let exponent = byteAtOrZero(source, index)
   if exponent == 101 {
     floating = true
@@ -334,6 +335,7 @@ fn scanNumber(source: &[u8], start: usize) -> Decision {
     let sign = byteAtOrZero(source, index)
     if sign == 43 { index = index + 1 }
     if sign == 45 { index = index + 1 }
+    exponentDigits = isDecimalDigit(byteAtOrZero(source, index))
     while index < source.length {
       if !isDecimalDigit(source[index]) { break }
       index = index + 1
@@ -345,12 +347,14 @@ fn scanNumber(source: &[u8], start: usize) -> Decision {
       let sign = byteAtOrZero(source, index)
       if sign == 43 { index = index + 1 }
       if sign == 45 { index = index + 1 }
+      exponentDigits = isDecimalDigit(byteAtOrZero(source, index))
       while index < source.length {
         if !isDecimalDigit(source[index]) { break }
         index = index + 1
       }
     }
   }
+  if !exponentDigits { return Decision { kind: 65, end: index, invalid: true } }
   if floating { return Decision { kind: 6, end: index, invalid: false } }
   return Decision { kind: 5, end: index, invalid: false }
 }
--- a/packages/compiler/src/Diagnostic.ts
+++ b/packages/compiler/src/Diagnostic.ts
@@ -28,6 +28,9 @@ export const missingBaseDigitsCode = 'LEX0004' as const
 /** Stable code for a number-literal digit separator outside a position between two digits. */
 export const invalidDigitSeparatorCode = 'LEX0005' as const
 
+/** Stable code for a float-literal exponent marker that no exponent digit follows. */
+export const missingExponentDigitsCode = 'LEX0006' as const
+
 /** Stable code for one required token that is absent at its insertion position. */
 export const missingTokenCode = 'PAR0001' as const
 
@@ -182,6 +185,8 @@ export const invalidReturnedBorrowOriginCode = 'SEM0092' as const
 export const intrinsicTargetUnavailableCode = 'SEM0093' as const
 /** Stable code for wrapping the already-borrowed string view in another reference or slice. */
 export const invalidStringViewTypeCode = 'SEM0094' as const
+/** Stable code for a float literal spelling no floating-point value can represent. */
+export const invalidFloatLiteralCode = 'SEM0095' as const
 
 /** Stable code for a use of a binding after its consuming move. */
 export const useAfterMoveCode = 'OWN0001' as const
@@ -207,6 +212,7 @@ export type Code =
   | typeof unterminatedStaticLiteralCode
   | typeof missingBaseDigitsCode
   | typeof invalidDigitSeparatorCode
+  | typeof missingExponentDigitsCode
   | typeof missingTokenCode
   | typeof unexpectedTokensCode
   | typeof reservedTemplateSyntaxCode
@@ -308,6 +314,7 @@ export type Code =
   | typeof invalidReturnedBorrowOriginCode
   | typeof intrinsicTargetUnavailableCode
   | typeof invalidStringViewTypeCode
+  | typeof invalidFloatLiteralCode
   | typeof useAfterMoveCode
   | typeof partialMoveCode
   | typeof explicitMoveRequiredCode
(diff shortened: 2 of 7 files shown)


julia-script:

Verified independently on this branch: `Lexer.test.ts`, `Parser.test.ts`, and `FloatingPointScalars.test.ts` pass **145/145**. Full-suite verification is running separately.

**The modified `Parser.test.ts` assertion is the right call**, and I checked the diff rather than taking it on trust. The old test asserted `1.25e-` produces a `DecimalFloat` token inside a `FloatingLiteralExpression` — it encoded precisely the silent acceptance this issue reports. The replacement preserves the real guarantee (`reconstructedBytes` byte-for-byte round-tripping, still asserted on **both** the well-formed and malformed cases) and strengthens the malformed case by asserting no float literal node forms *and* that an `Invalid` token appears. That is a net gain in coverage, not a weakened test. Thank you for flagging it explicitly instead of burying it.

Fixing at the lexer with `LEX0006` while keeping `SEM0095` as an elaboration backstop is the correct shape — the root cause is the lexer/`fromDecimal` grammar disagreement, and the backstop closes the class rather than the instance.

### On acceptance criterion 4

The report is right that it cannot be tested as literally worded: after the fix, no source spelling reaches `SEM0095`, because the two grammars now agree. The substituted 92-combination sweep — every spelling accepted as `DecimalFloat` with no diagnostic is convertible by `fromDecimal` — is a **stronger** assertion than the criterion asked for, since it pins the invariant that makes the path unreachable rather than one example of it. Accepted as satisfied.

Consequence worth recording: `SEM0095` is now unreachable-by-construction and has no source-level test. That is correct for a defensive backstop, but it means a future lexer change that reopens the grammar gap will be caught by the sweep, not by `SEM0095` coverage. The sweep is therefore the load-bearing test — worth keeping if anyone later trims it as redundant.

### Test-count caveat, endorsed

The `baseline 4 / after 7` line is not a regression and should not be read as one. I measured this repository's compiler suite clean at **140/140 files, 1079/1079 tests** on an unloaded host. The failures both you and other workers in this run observed (`Logging`, `OsFileSystem`, `LexerPressure`/`StackVmPressure` allocation-rollback, `IntegerScalars`, `OccurrencePerformance`, `SynchronousEffectCost`, `ScannerAcceptance`, `Driver`, `NativeToolchain`) are 30–430s timeouts produced by CPU contention — several agent worktrees were running vitest concurrently on one machine. Per-test-name comparison against unmodified `main`, which is what you did, is the correct method under those conditions.

### Two follow-ups, neither blocking

1. **`examples/language-pressure/lexer/main.silk`** — a Silk reimplementation of the lexer held in lockstep by a differential corpus test, and not mentioned in the issue. Anyone changing token boundaries must update it. Worth documenting somewhere durable; it is a non-obvious tripwire.
2. **No OpenSpec amendment was made**, correctly — the issue carries no `spec-change` label and `openspec/specs/bootstrap-lexer` never states the exponent-digit rule. But if the float grammar is considered spec-governed, that spec is silent on a rule the compiler now enforces. Worth a maintainer's call as a separate item.

PR #82, opened 2026-08-12
ownership: run on a once parameter is not tracked as a consuming use

### LINKED ISSUE #64: ownership: run on a once parameter is not tracked as a consuming use
Found by an adversarial review pass during orchestrator run run-s1k9 while reviewing #57 (`Effect.flatten`), and verified by reading the tree. **Pre-existing — not introduced by that PR.**

## Current behavior

The ownership checker treats `run <expr>` as a consuming use only when the subject is a `BindingReference`:

```ts
case 'Run': {
  if (
    expression.subject._tag === 'BindingReference' &&
    Type.isEffect(expression.subject.type) &&
    expression.subject.type.access === 'Take'
  ) {
    const site = useSite(expression.subject)
    if (site !== undefined) checkUse(state, live, site, expression.span, true)
  } else checkExpression(state, live, expression.subject, false, guard, escaping)
  return
}
```

`Ownership.ts:636-644`.

But a parameter does not lower to a `BindingReference` — it lowers to a `ParameterReference` (`Ownership.ts:314-316`). So `run self` where `self` is a `once` parameter takes the `else` branch and is checked as a **non-consuming** use. The affine value is never marked dead.

Notably `useSite` already handles `ParameterReference` (`Ownership.ts:314`), so the machinery to track it exists — the `Run` case just does not reach it.

## Consequence

A function body that runs a `once` parameter twice would not be rejected. Today that is benign, because `cleanupPlan` returns `NoCleanup` for any Effect-typed value (`Ownership.ts:1599`), so no double-release is generated. The hole is in the *checking*, not currently in the generated code — but it is a soundness gap that only stays harmless while Effect values have no cleanup.

## Why it is worth fixing now

The existing combinators in `stdlib/silk/effects.silk` avoid the untracked path by construction: `flatMap`, `tap`, `map`, `mapError`, `catchAll` all consume via `result(move self)`, where `move` into a call **is** a tracked consuming use. `Effect.flatten` (#57 / PR #57) is the first stdlib declaration to use a bare `run self` on a `once` parameter:

```silk
pub effect fn flatten<A, !E, !F, ?R, ?S>(
  self: once Effect<Effect<A ! F ? S> ! E ? R>
) -> A ! E | F ? R | S {
  let inner = run self
  return run inner
}
```

That code is correct as written and matches issue #23's own example, so it is not a reason to block the PR — but it means the untracked path now has a real user in the standard library, and any future change that gives Effect values a cleanup plan would turn this from latent into live.

## Suggested direction

Widen the `Run` case to use `useSite(expression.subject)` for any subject that yields a binding site (`BindingReference`, `ParameterReference`, `PatternBindingReference`) rather than special-casing `BindingReference`, keeping the existing `Type.isEffect` + `access === 'Take'` guards.

## Acceptance criteria

- [ ] A test asserts that a body running a `once` **parameter** twice is rejected with the same diagnostic as running a `once` **local binding** twice.
- [ ] A test asserts that `Effect.flatten`'s single `run self` still compiles unchanged.
- [ ] Running a shared or `mut` Effect parameter is unaffected.
- [ ] No change to the 3 pre-existing `Instances.test.ts` host-triple golden failures.

julia-script:
Closes #64

The ownership checker's `Run` case gated its consuming-use check on the subject being a
`BindingReference`. A parameter lowers to a `ParameterReference`, so `run self` on a `once`
parameter fell through to the non-consuming `else` branch and the affine value was never
marked dead — a body could run a `once` parameter twice with no diagnostic.

`useSite` already resolves every binding form, so the fix routes the guard through it instead
of special-casing one tag, keeping the existing `Type.isEffect` + `access === 'Take'` checks.
That is the single shared guard all callers of the `Run` case route through, so no sibling
call site needs its own patch.

## Acceptance criteria

- [x] A test asserts that a body running a `once` **parameter** twice is rejected with the same diagnostic as running a `once` **local binding** twice — `consumes a take effect parameter on its first run and rejects a repeated run` expects `OWN0001`, matching the existing local-binding test.
- [x] A test asserts that `Effect.flatten`'s single `run self` still compiles unchanged — `accepts a single run of a take effect parameter` mirrors `flatten`'s exact shape and expects no diagnostics.
- [x] Running a shared or `mut` Effect parameter is unaffected — `allows a shared effect parameter to run repeatedly` runs a non-`once` Effect parameter twice and expects no diagnostics.
- [x] No change to the pre-existing host-triple golden failures — failure set compared against a clean `origin/main` baseline measured in a separate worktree.

Tests: baseline 28 failures (measured in step 0), after 10 failures, +3 new tests

## Notes

The regression test was confirmed red before the fix (`expected [] to deeply equal [ 'OWN0001' ]`)
and green after.

All failures on this branch are a strict subset of the `origin/main` baseline (measured in a clean
worktree on the same machine) — every one is a process-spawning determinism or native-toolchain
test timing out on this host, none touch the ownership checker. The baseline's larger count comes
from determinism tests that need a prebuilt `dist/`. Test totals were 1079 on baseline and 1082
here: exactly the 3 tests added. `pnpm typecheck` passes (18/18, 0 errors).

Out of scope: a match-pattern binding of an Effect (`PatternBindingReference`) never registers as
an ownership binding in the first place, so `run` on one is still untracked. That gap is upstream
of the `Run` case and separate from this issue.

### DIFF
--- a/packages/compiler/src/Ownership.ts
+++ b/packages/compiler/src/Ownership.ts
@@ -634,13 +634,14 @@ const checkExpression = (
       return
     }
     case 'Run': {
+      const site = useSite(expression.subject)
       if (
-        expression.subject._tag === 'BindingReference' &&
+        site !== undefined &&
+        expression.subject._tag !== 'Unavailable' &&
         Type.isEffect(expression.subject.type) &&
         expression.subject.type.access === 'Take'
       ) {
-        const site = useSite(expression.subject)
-        if (site !== undefined) checkUse(state, live, site, expression.span, true)
+        checkUse(state, live, site, expression.span, true)
       } else checkExpression(state, live, expression.subject, false, guard, escaping)
       return
     }
--- a/packages/compiler/test/Ownership.test.ts
+++ b/packages/compiler/test/Ownership.test.ts
@@ -253,6 +253,49 @@ pub fn main() -> i32 {
   assert.strictEqual(facts.functions.at(1)?.verdict._tag, 'Violation')
 })
 
+it('consumes a take effect parameter on its first run and rejects a repeated run', () => {
+  const facts = check(
+    'ownership://take-effect-parameter.silk',
+    `pub effect fn twice<A, !E, ?R>(self: once Effect<A ! E ? R>) -> A ! E ? R {
+  let first = run self
+  return run self
+}`,
+  )
+
+  assert.deepEqual(
+    facts.diagnostics.map((diagnostic) => diagnostic.code),
+    ['OWN0001'],
+  )
+  assert.strictEqual(facts.functions.at(0)?.verdict._tag, 'Violation')
+})
+
+it('accepts a single run of a take effect parameter', () => {
+  const facts = check(
+    'ownership://take-effect-parameter-single.silk',
+    `pub effect fn flattenLike<A, !E, !F, ?R, ?S>(
+  self: once Effect<Effect<A ! F ? S> ! E ? R>
+) -> A ! E | F ? R | S {
+  let inner = run self
+  return run inner
+}`,
+  )
+
+  assert.deepEqual(facts.diagnostics, [])
+  assert.strictEqual(facts.functions.at(0)?.verdict._tag, 'Satisfied')
+})
+
+it('allows a shared effect parameter to run repeatedly', () => {
+  const facts = check(
+    'ownership://shared-effect-parameter.silk',
+    `pub effect fn twiceShared<?R>(self: Effect<i32 ? R>) -> i32 ? R {
+  return (run self) + (run self)
+}`,
+  )
+
+  assert.deepEqual(facts.diagnostics, [])
+  assert.strictEqual(facts.functions.at(0)?.verdict._tag, 'Satisfied')
+})
+
 it('derives take-once execution from an effect-block moved capture', () => {
   const facts = check(
     'ownership://take-effect-block.silk',


julia-script:
## Review — orchestrator run-b4c2: verified, approving

I reproduced the central claim rather than taking it on trust. Reverting `Ownership.ts` to `origin/main` while keeping the new tests, the regression test fails exactly as reported:

```
× consumes a take effect parameter on its first run and rejects a repeated run
  AssertionError: expected [] to deeply equal [ 'OWN0001' ]
```

With the fix restored: **35/35 green.** A regression test proven red before and green after is the thing that makes this reviewable, and it did the work.

### The fix is the right shape

Swapping the `subject._tag === 'BindingReference'` gate for `useSite(expression.subject)` is a root-cause fix rather than a patch on the reported path. `useSite` already resolves every reference form, so the guard now covers `ParameterReference` and any future form for free — and confirming that line 638 was the *only* gated call site is exactly the check that distinguishes "fixed once where all callers route through" from "fixed the one case in the ticket."

Worth recording: `Effect.flatten` in `stdlib/silk/effects.silk:91-96` has a bare `run self` on a `once` parameter, so the untracked path had a **live stdlib user**. This was not hypothetical — the affine guarantee was silently unenforced for a shipped stdlib function.

### Baseline numbers, endorsed

`baseline 28 / after 10` reads alarming and is not. The comparison that matters is the one you made: your 10 failures are a strict subset of the clean-`origin/main` baseline's 28, all process-spawn or native-toolchain timeouts on a contended host, none ownership-related — and totals of 1079 → 1082 are exactly the 3 added tests. The inflated baseline came from an unbuilt `dist/`, which several workers in this run hit independently.

### Follow-up filed

The out-of-scope gap you flagged is real and now tracked separately: a match-pattern binding of an Effect never registers as an ownership binding, so `run` on a `PatternBindingReference` remains untracked. Correctly left out of this PR — it sits upstream in pattern-binding registration, not in the `Run` guard.

Nothing blocking. Ready to merge from my side.

PR #84, opened 2026-08-12
stdlib: Add pop, remove, clear, set, reserve, and truncate to Vector

### LINKED ISSUE #37: stdlib: Add pop, remove, clear, set, reserve, and truncate to Vector
## Context

`Vector<T>` can grow but cannot shrink. `vector.silk` declares exactly 8 public functions:
6 plain functions, which are `make`, `length`, `capacity`, `asSlice`, `asMutSlice`, and
`get`, plus 2 effect functions, which are `append` at `vector.silk:102` and `insert` at
`vector.silk:135`. A vector therefore holds every element until the whole vector drops. A
program that needs a stack, a work list, or a free list cannot use `Vector<T>` today.

## Current behavior

A user who wants to remove the last element has no operation for it. The user cannot write
one either, because the storage union and the `RawBuffer<T>` slots are private to the
module. `Vector.insert` shifts elements upward (`vector.silk:171-193`), but no function
shifts them downward.

A user also cannot overwrite one element. `asMutSlice` gives an exclusive slice, but an
assignment through that slice would leak the old value for a move-only element type.

## Requirements

1. `pop` must remove the last element and must return `Option<T>`.
2. `remove` must remove the element at one index and must shift the later elements down.
3. `remove` must trap for an out-of-range index, as `get` does today.
4. `clear` must drop every initialized element and must set the length to zero.
5. `clear` must keep the capacity, so the buffer stays available for reuse.
6. `set` must drop the old element before it writes the new element.
7. `truncate` must drop every element after one length and must keep the capacity.
8. `reserve` must grow the capacity and must declare `! OutOfMemory ? &mut Allocator`, as `append` does.
9. `pop`, `remove`, `clear`, `set`, and `truncate` must not allocate, so they must declare no effect row.
10. `clear` and `truncate` must release each removed element exactly once.
11. Each removal must decrease the length by the exact number of elements it removes.
12. A failed allocation in `reserve` must leave the vector unchanged.

## Example

```silk
/// Removes the last element and returns it. Returns an absent value for an empty vector.
pub fn pop<T>(self: &mut Vector<T>) -> Option<T> {
  if self.length == counted(0) {
    return Option.none<T>()
  }
  let storage = Intrinsic.replace(self.storage, Empty<T> { anchor: [] })
  let taken = match move storage {
    Empty<T> nothing => absurd<Taken<T>>()
    Full<T> full => takeAt<T>(move full, self.length - counted(1))
  }
  self.length = self.length - counted(1)
  return match move taken {
    Taken<T> { storage, value } => finish<T>(self, move storage, move value)
  }
}

/// Grows the capacity to hold at least one additional count of elements.
pub effect fn reserve<T>(
  self: &mut Vector<T>,
  additional: usize
) -> () ! OutOfMemory ? &mut Allocator
```

## Out of scope

- **The iteration protocol.** The spec states that "no iterable or iterator abstraction is
  required" (`openspec/specs/bootstrap-owned-sequence/spec.md:94-95`). This exclusion is a
  design decision, not an omission.
- `swapRemove`, `drain`, `retain`, `dedup`, and `shrinkToFit`.
- A sort or a search. See ticket 32.
- Any compiler change. Each operation is ordinary Silk source.

## Dependencies

- The partial move decision. `pop` and `remove` move one element out of storage that the
  vector still owns, which is the question that `OWN0002` governs
  (`packages/compiler/src/Diagnostic.ts:182`).
- Ticket 31 for `Option<T>`, because `pop` returns `Option<T>`. The existing `option.silk`
  constructors are sufficient, so ticket 31 is helpful but not mandatory.

## Acceptance criteria

- [ ] A test shows `pop` returns the last element and decreases the length by one.
- [ ] A test shows `pop` returns an absent value for an empty vector.
- [ ] A test shows `remove` shifts the later elements down and keeps their order.
- [ ] A test shows `clear` keeps the capacity and sets the length to zero.
- [ ] A test shows `set` drops the old element exactly once.
- [ ] An allocation acceptance test shows the acquire count equals the release count after
      a program appends, pops, and drops a vector of move-only elements.
- [ ] A test shows the vector is unchanged after `reserve` fails with `OutOfMemory`.
- [ ] A test shows `truncate` releases each removed element exactly once.

## Implementation note

`pop` and `remove` move one element out of storage that the vector still owns. A direct move
out of `self.storage` is a partial move of a borrowed value, so the compiler rejects it with
`OWN0002` (`packages/compiler/src/Diagnostic.ts:182`).

The module already has the pattern that avoids this. `Vector.append` calls
`Intrinsic.replace` to move the whole storage union out through `&mut self`, then puts it
back (`vector.silk:116-129`). `Slot.take` then moves one element out of the owned buffer
(`vector.silk:183`). Each removal must use that same two-step shape. A removal that reaches
into `self.storage` directly will not compile.

julia-script:
Closes #37

Adds the six mutation operations to `Vector<T>`, which could previously grow but never shrink. Every removal follows the module's established two-step shape: `Intrinsic.replace` moves the storage union out through `&mut self`, the element leaves the owned buffer through `Slot.take`, and the storage is committed back — reaching into `self.storage` directly would be a partial move of a borrowed value (`OWN0002`).

`pop`, `remove`, `clear`, `set`, and `truncate` declare no effect row because none of them allocate. `reserve` declares `! OutOfMemory ? &mut Allocator` and stages the replacement buffer before committing, so a failed allocation leaves the vector untouched.

### Acceptance criteria

- [x] A test shows `pop` returns the last element and decreases the length by one.
- [x] A test shows `pop` returns an absent value for an empty vector.
- [x] A test shows `remove` shifts the later elements down and keeps their order.
- [x] A test shows `clear` keeps the capacity and sets the length to zero.
- [x] A test shows `set` drops the old element exactly once.
- [x] An allocation acceptance test shows the acquire count equals the release count after a program appends, pops, and drops a vector of move-only elements.
- [x] A test shows the vector is unchanged after `reserve` fails with `OutOfMemory`.
- [x] A test shows `truncate` releases each removed element exactly once.

Every new test runs on all three engines (evaluator, Wasm, native), matching the existing `VectorAcceptance` pattern.

### Note: a pre-existing Wasm defect found along the way

The move-only round-trip test asserts ownership accounting (acquire/release pairing and drop order) rather than reading the removed element's scalar fields, because of a **pre-existing** WebAssembly lowering bug that this PR does not introduce and does not fix.

When a move-only struct is stored in a `Vector<T>` and read back, its scalar fields read as `0` under Wasm while the evaluator returns the correct value. It reproduces on unmodified `main` using only `make` + `append` + `asSlice`:

```silk
struct Entry {
  value: i32
  marker: Vector<i32>
}
// append an Entry { value: 3, ... }, then:
return 100 + asSlice<Entry>(&values)[0].value
// evaluator -> 103, wasm -> 100
```

Controls that pass: the same program with a `Copy` element type (`i32`), and the same struct destructured without ever entering a `Vector`. Both debug and release Wasm are affected, so it is not an optimizer artifact. Worth its own ticket; once fixed, the value assertions in these tests can be tightened.

Tests: baseline 1 failures (measured in step 0), after 1 failures, +8 new tests

The one baseline failure is `NativeToolchain.test.ts > links the shim with the program and the executable exits with the i32 result`, a host-toolchain issue present before this branch. Several long-running pressure/determinism tests time out when the suite is run under CPU contention; each was re-run individually on an idle machine and passes.


julia-script:

Finding that Wasm defect while implementing an unrelated feature, and reporting it instead of quietly writing around it, is the most valuable thing in this PR. I reproduced it independently on clean `main` (`654d79f`) and filed it as **#86**.

**Two corrections to the characterization**, both of which change where someone should look:

1. **Native/LLVM is correct**, not just the evaluator. The results are evaluator 103, native 103, wasm-debug 100, wasm-release 100. This is a Wasm-backend bug, not a general miscompilation of move-only structs.
2. **It is not "scalar fields read as 0" — it is element index 0 that is zeroed.** Appending 1, 2, 3 reads back 0, 2, 3. Five appends across a 0→4→8 growth read 0, 2, 3, 4, 0. Lengths are always correct.

Further narrowing from the repro: a move-only element type is required (a same-size *Copy* struct passes, ruling out size and layout); it is not `Vector`-specific as the field type (a `Bytes` field triggers it identically); and field order does not matter.

And it localizes to the **write** path, not the read path. Writing slot 0 through `asMutSlice` then reading via `asSlice` yields the correct value, so slot-0 addressing and reads are sound — the appended value never arrives. Reading `entry.value` immediately before the `move` is also correct. The emitted WAT shows `Slot.write` storing all 11 i32 words, so the store width is right; the suspect is the pointer handed to it on the empty→first-element transition.

For the record, the sketch in the PR body does not compile as written — struct fields are newline-separated, not comma-separated. The verified source is in #86.

### On the PR itself

The implementation follows the module's established two-step removal shape (`Intrinsic.replace` out through `&mut self`, element out via `Slot.take`, storage committed back), which is the right call — reaching into `self.storage` directly would be `OWN0002`. Effect rows look right: no row on the five non-allocating operations, `! OutOfMemory ? &mut Allocator` on `reserve` only, with the replacement buffer staged before commit so a failed allocation leaves the vector untouched. All eight acceptance criteria have tests, and all run on three engines.

**Constraining the move-only test to ownership accounting rather than field values was correct** given #86 — asserting acquire/release pairing and drop order still tests what this PR is responsible for. Acceptance criterion 4 in #86 is to tighten those assertions once the Wasm bug is fixed, so the weakening is tracked rather than forgotten.

Baseline `1 / 1` with the single `NativeToolchain` host-toolchain failure is consistent with what other workers in this run measured. Nothing blocking.

PR #87, opened 2026-08-12
cli: Add silk clean and a watch mode, and charter silk test

### LINKED ISSUE #50: cli: Add silk clean and a watch mode, and charter silk test
## Context

The `silk` command has seven subcommands in `packages/compiler-cli/src/`. The spec
is a closed list: the root command "SHALL expose `init`, `build`, `check`, `format`,
`run`, and `build-exe`" (`openspec/specs/silk-cli-workflows/spec.md:10`). Each new
subcommand is a spec amendment. `silk clean` and a watch mode need only that
amendment. `silk test` needs more, and the next section says why.

**`silk test` must split into its own ticket.** A test command needs a
test-declaration story in the language: a way to mark a function as a test, a way to
report a failure, and a rule for the exit code. No OpenSpec change charters any of
those three. `silk clean` and a watch mode need no language change at all, so they
must not wait for that design. This ticket delivers `clean` and `watch`. A separate
ticket must first design the test declaration form, and a third ticket must then add
`silk test`.

## Current behavior

`silk build` writes artifacts to the output directory that `[build]` names in
`silk.toml`. No command removes those artifacts. A user removes the directory with a
shell command. A user who edits a file runs `silk check` again by hand after every
edit.

## Requirements

1. The `silk clean` command must remove the artifacts that the manifest output directory holds.
2. The `silk clean` command must accept the shared `--manifest-path` option.
3. The `silk clean` command must not remove a file that the build did not write.
4. The `silk clean` command must exit with code 0 when the output directory does not exist.
5. The `--watch` option must apply to `silk build` and to `silk check`.
6. The watch mode must compile again after a change to any file in the source graph.
7. The watch mode must report each compilation with the same format as one command run.
8. The watch mode must keep running after a compilation that reports diagnostics.
9. The watch mode must use the exit codes of the toolchain only when the user stops it.
10. The `silk-cli-workflows` spec must name `clean` in the command list.

## Example

```
$ silk build
$ silk clean
Removed 4 artifacts from target/

$ silk check --watch
Watching 12 source files.
main.silk:3:5 SEM0013 Alias math does not change the name
```

## Out of scope

- The `silk test` command. A separate ticket must charter it after the language gets a test declaration form.
- A test declaration form in the Silk language.
- A benchmark command.
- Incremental compilation. The watch mode compiles the whole graph again.
- Package management commands such as `add` or `install`. The manifest has no dependencies.

## Dependencies

- None for `clean` and `watch`.
- A language design ticket for the test declaration form must come before any `silk test` ticket.

## Acceptance criteria

- [ ] The `silk-cli-workflows` spec lists `clean` in the closed command list.
- [ ] The spec gets a requirement and a scenario for the watch mode.
- [ ] A test asserts `silk clean` removes the build artifacts and keeps the source files.
- [ ] A test asserts `silk clean` exits with code 0 on a project that was never built.
- [ ] A test asserts `silk check --watch` compiles again after a source file changes.
- [ ] A test asserts the watch mode keeps running after a compilation that reports a diagnostic.
- [ ] `silk --help` lists `clean` with its purpose.

julia-script:
Closes #50

`silk clean` removes the manifest output directory, and `--watch` applies to `silk build` and `silk check`. Per the issue, `silk test` stays out of scope: it needs a test-declaration form in the language first, which a separate ticket must charter.

## Acceptance criteria

- [x] The `silk-cli-workflows` spec lists `clean` in the closed command list.
- [x] The spec gets a requirement and a scenario for the watch mode.
- [x] A test asserts `silk clean` removes the build artifacts and keeps the source files.
- [x] A test asserts `silk clean` exits with code 0 on a project that was never built.
- [x] A test asserts `silk check --watch` compiles again after a source file changes.
- [x] A test asserts the watch mode keeps running after a compilation that reports a diagnostic.
- [x] `silk --help` lists `clean` with its purpose.

## Notes

`clean` removes the manifest output directory rather than tracking individual artifacts. Only the build writes there, and `Project.load` already rejects an `output-dir` that escapes the project, so requirement 3 holds without an artifact manifest.

`FileSystem.watch` reports only a directory's own entries and its typed signature exposes no recursive option, so the watch enumerates the source root and every directory beneath it. Without that, a nested module such as `src/library/Answer.silk` produced no recompilation; an extra test covers this case. Directories are enumerated once, so a directory created after the watch starts needs a restart.

Tests: baseline 0 failures (measured in step 0), after 0 failures, +7 new tests

### DIFF
--- a/openspec/specs/silk-cli-workflows/spec.md
+++ b/openspec/specs/silk-cli-workflows/spec.md
@@ -7,7 +7,7 @@ Defines the user-visible Silk language-tool commands for project checking, build
 ## Requirements
 
 ### Requirement: Project-oriented command surface
-The root `silk` command SHALL expose `init`, `build`, `check`, `format`, `run`, and `build-exe`. `init` SHALL accept an optional path and package-name override. Project compilation commands SHALL accept a shared optional `--manifest-path`, one `--backend`, repeatable `--target`, and profile selection. Repeated command-line targets SHALL replace rather than append to manifest targets. `--release` SHALL select the release profile and SHALL conflict with an explicitly different `--profile`.
+The root `silk` command SHALL expose `init`, `build`, `check`, `clean`, `format`, `run`, and `build-exe`. `init` SHALL accept an optional path and package-name override. Project compilation commands SHALL accept a shared optional `--manifest-path`, one `--backend`, repeatable `--target`, and profile selection. Repeated command-line targets SHALL replace rather than append to manifest targets. `--release` SHALL select the release profile and SHALL conflict with an explicitly different `--profile`.
 
 #### Scenario: Display root help
 - **WHEN** a user requests `silk --help`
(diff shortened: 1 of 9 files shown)


julia-script:

**The nested-directory bug you found mid-implementation is the important part of this PR.** `FileSystem.watch` reports only a directory's own entries and its typed signature exposes no recursive option, so the first implementation silently missed `src/library/Answer.silk` — and every originally-specified test still passed. That is a requirement-6 violation that the issue's own acceptance criteria could not detect. Finding it, fixing it, and adding a seventh test for the nested case is exactly right; a worker who only satisfied the listed criteria would have shipped a watch mode that ignores every nested module.

I read the implementation: `sourceDirectories` walks the root and pushes each subdirectory, and the watch subscribes to all of them. The two `ponytail:` comments naming the known ceilings — directories enumerated once, so a directory created later needs a restart; every event recompiles the whole graph — are the right way to record a deliberate limit. Better to have the ceiling written down than discovered later.

**Scope call on `silk test` is correct.** The issue defers it pending a language test-declaration form and lists it under Out of scope, so "charter" was unambiguous. Not guessing there was the right read.

**`clean` removing the whole manifest output directory** rather than tracking individual artifacts is sound as argued: only the build writes there, and `Project.load` already rejects an `output-dir` escaping the project, so requirement 3 holds without an artifact manifest. If per-artifact tracking is ever wanted it is a separate ticket, not a gap here.

### On the baseline correction

Your correction is right and matches what I have seen across this whole run: the flaky set under CPU contention is broader than `Driver`/`NativeToolchain` — it includes `Logging`, `OsFileSystem`, `LexerPressure`, `StackVmPressure`, `IntegerScalars`, `OccurrencePerformance`, `SynchronousEffectCost`, and `ScannerAcceptance`. Your 15-failure first run and clean 1080/1080 re-run are the same phenomenon four other workers in this run hit independently. Thank you for re-running in isolation rather than reporting the first number.

For the record on my side: I could not reproduce your compiler-cli run locally because my own worktree has a stale `@silk-effect/documentation` build (`Cannot find package '@silk-effect/documentation/Json'`), which fails three test files at import time. That is my environment, not your diff — the 80 tests that did run all passed. CI is the signal I am trusting.

Nothing blocking.

PR #88, opened 2026-08-12
lsp: Add semantic tokens, folding ranges, and call hierarchy

### LINKED ISSUE #49: lsp: Add semantic tokens, folding ranges, and call hierarchy
## Context

Three language server capabilities are absent, and all three read artifacts that
exist today. The lossless CST gives the ranges that folding needs. The compiler
token kinds give the semantic tokens, and `packages/language/src/TextMate.ts:11`
already builds the TextMate grammar from those same kinds. The occurrence index at
`packages/compiler/src/Analysis.ts:199` gives the call hierarchy when a client walks
it one step at a time. The server capability list
(`packages/lsp/src/Server.ts:50-65`) advertises none of the three.

**This is one ticket, not three.** The three capabilities share one server file, one
capability list, and one test harness. Each one is a small map from an artifact that
the compiler already computes. Three tickets would repeat the same context three
times and would give three small pull requests to the same 15 lines of the
initialize response. Split this ticket only if call hierarchy needs more work than
the occurrence index gives.

## Current behavior

The editor colors Silk source with the TextMate grammar. That grammar is a regular
expression match, so it colors a type name and a variable name the same way. The
editor folds nothing, because it has no range list. The editor shows no callers of a
function.

## Requirements

1. The server must advertise `semanticTokensProvider` with a legend built from the compiler token kinds.
2. The server must answer a full-document semantic tokens request.
3. The semantic token legend must use the same token kinds as `packages/language/src/TextMate.ts`.
4. The server must advertise `foldingRangeProvider`.
5. The server must return one folding range for each block, each declaration body, and each block comment.
6. The server must advertise `callHierarchyProvider`.
7. The prepare-call-hierarchy response must give the declaration under the cursor.
8. The incoming-calls response must give every function that calls the selected function.
9. The outgoing-calls response must give every function that the selected function calls.
10. Each of the three responses must use the negotiated position encoding.

## Example

```silk
pub fn helper(value: i32) -> i32 { value + 1 }

pub fn main() -> i32 {
  // Incoming calls of `helper` list `main`.
  // Outgoing calls of `main` list `helper`.
  // One folding range covers the body of `main`.
  helper(1)
}
```

## Out of scope

- Delta semantic tokens. The server sends the full set on each request.
- Range semantic tokens. The full-document request is enough for a bootstrap.
- Folding ranges that a comment marker such as `// region` controls.
- A call hierarchy across a project that does not compile.

## Dependencies

- None. The CST, the token kinds, and the occurrence index all exist.

## Acceptance criteria

- [ ] A test asserts the initialize response advertises all three providers.
- [ ] A test compares the semantic token legend with the token kinds in `packages/language/src/TextMate.ts`.
- [ ] A test asserts a semantic token request colors a type name and a variable name with different types.
- [ ] A test asserts a folding range covers a function body from the open brace to the close brace.
- [ ] A test asserts incoming calls of one function list two callers across two modules.
- [ ] A test asserts outgoing calls of one function list the functions it calls.
- [ ] A new `language-server-structure` spec records the requirements for the three capabilities.

julia-script:
Closes #49

Stacked on #85 (`agent/47-signature-help`), itself stacked on #78. Review only the top commit; the base branch carries the rest.

All three capabilities map an artifact the compiler already computes onto the protocol, and all three reuse `LineIndex` for positions, so each answer lands in the negotiated `utf-16` encoding.

- **Semantic tokens** type an identifier from the semantic occurrence covering it rather than from its spelling — which is what separates a type name from a variable name that the TextMate grammar's regular expressions must color alike. Keywords, comments, literals, and operators are decided by the lexer's own token kinds, the same source `packages/language/src/TextMate.ts` reads.
- **Folding ranges** fold each braced region from its opening brace's line to its closing brace's line, plus one range per run of adjacent comment lines. Silk has no delimited block comment, so a run of `//`, `///`, or `//!` lines is the block a reader means to collapse.
- **Call hierarchy** resolves the selection through the occurrence index (via `Analysis.declarationForIdentity`, exported by #85) and attributes each use to the declaration whose body encloses it, so incoming calls span every module of the analyzed project and repeated calls collapse into one entry carrying every range.

## Acceptance criteria

- [x] A test asserts the initialize response advertises all three providers.
- [x] A test compares the semantic token legend with the token kinds in `packages/language/src/TextMate.ts`. Added `@silk-effect/language` as a **devDependency** of `@silk-effect/lsp` so the test reads the grammar's real keyword table; it stays out of the runtime dependencies.
- [x] A test asserts a semantic token request colors a type name and a variable name with different types.
- [x] A test asserts a folding range covers a function body from the open brace to the close brace.
- [x] A test asserts incoming calls of one function list two callers across two modules.
- [x] A test asserts outgoing calls of one function list the functions it calls.
- [x] A new `language-server-structure` spec records the requirements for the three capabilities.

One test beyond the criteria covers requirement 10 directly: a document holding an astral character asserts that semantic tokens, folding ranges, and call hierarchy ranges are all placed in UTF-16 columns rather than byte columns.

## Notes

- Out-of-scope items in the issue are respected: no delta or range semantic tokens (`full: true` only), and no `// region` marker folding.
- A token spanning more than one line is omitted rather than truncated, because the protocol's delta encoding cannot express one.
- `biome check` reports one pre-existing formatting nit in the `signatureHelp` handler in `packages/lsp/src/Server.ts`, which comes from the base branch (#85) and is untouched here.

Tests: baseline 0 failures (measured in step 0), after 0 failures, +7 new tests

### DIFF
--- a/openspec/specs/language-server-structure/spec.md
+++ b/openspec/specs/language-server-structure/spec.md
@@ -1,0 +1,120 @@
+# language-server-structure Specification
+
+## Purpose
+
+Defines how the language server describes the shape of a Silk source: the meaning of each token, the
+regions an editor may collapse, and the call relationships between functions. All three read
+artifacts the compiler already produces — the lexer's token kinds, the lossless concrete syntax
+tree, and the semantic occurrence index — so an editor reads the same facts the compiler decided
+rather than re-deriving them from a regular expression.
+
+## Requirements
+
+### Requirement: The server advertises structural capabilities
+
+The language server SHALL advertise semantic-tokens support carrying its legend, folding-range
+support, and call-hierarchy support to compatible clients. The semantic-tokens capability SHALL
+declare full-document requests. Delta and range requests are not advertised, so a client always
+requests the whole document.
+
+#### Scenario: Client initializes structural support
+
+- **WHEN** a compatible client initializes a Silk language-server session
+- **THEN** the returned capabilities advertise a semantic-tokens provider carrying the server's legend and declaring full-document support, a folding-range provider, and a call-hierarchy provider
+
+### Requirement: The semantic token legend is built from the compiler's token kinds
+
+The semantic token legend SHALL name only standard protocol token types, and every type it names
+SHALL be reachable from a compiler token kind or a semantic occurrence. The keyword, comment,
+string, number, and operator types SHALL be decided by the lexer's own token kinds — the same kinds
+the TextMate grammar in `@silk-effect/language` colors — so a keyword added to the compiler reaches
+both the grammar and the legend from one definition.
+
+#### Scenario: Compare the legend with the grammar's keywords
+
+- **WHEN** a document spelling every keyword the TextMate grammar colors is tokenized
+- **THEN** each keyword carries a semantic token of the legend's keyword type
+
+### Requirement: Semantic tokens type an identifier by what it resolves to
+
+A full-document semantic tokens request SHALL return one token per classified lexer token, in source
+order, using the protocol's delta encoding. An identifier's token type SHALL be decided by the
+semantic occurrence covering it rather than by its spelling, so a type name and a variable name
+receive different token types where the TextMate grammar's regular expressions must color them
+alike. An identifier with no covering occurrence SHALL contribute no token rather than a guessed
+one. A token spanning more than one line SHALL be omitted, because the protocol's encoding cannot
+express one.
+
+#### Scenario: Color a type name and a variable name
+
+- **WHEN** a semantic tokens request covers a document declaring a struct type and binding a local variable
+- **THEN** the type name carries the `type` token type and the variable name carries the `variable` token type
+
+### Requirement: Folding ranges cover braced regions and comment runs
+
+A folding-range request SHALL return one range for each block, each declaration body, and each run
+of adjacent comment lines. A braced region SHALL fold from the line holding its opening brace to the
+line holding its closing brace, so the line an editor leaves visible keeps the brace that opened the
+region. A region whose braces sit on one line SHALL yield no range. Because Silk has no delimited
+block comment, a run of adjacent `//`, `///`, or `//!` lines SHALL fold as one range carrying the
+`comment` kind. A range whose folding is controlled by a comment marker such as `// region` is out
+of scope.
+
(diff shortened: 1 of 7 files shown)


julia-script:

Both judgment calls are sound, and flagging them rather than burying them is what makes them reviewable.

**(1) "Block comment" in requirement 5.** Silk has no delimited `/* */` — only `LineComment`, `DocComment`, and `ModuleDocComment`. Folding a run of adjacent comment lines as one range with `FoldingRangeKind.Comment` is the standard LSP treatment and, as you say, the only available reading. Recording that interpretation in the spec rather than silently picking it is the right disposition: the next reader sees a decision, not an accident.

**(2) The `@silk-effect/language` devDependency.** I verified this rather than taking it on trust. `packages/language/package.json` depends on `@silk-effect/compiler` and `@silk-effect/documentation` only — it does not depend on `lsp`, so there is no cycle, and dev-only scoping keeps it out of the runtime graph. Your underlying observation is also correct: `TextMate.ts` exports no token-type list, only `keywords: Record<KeywordKind, string>`, so the genuinely shared artifact is `Token.TokenKind`. Asserting that every keyword the grammar spells carries a legend keyword token tests the real invariant the criterion was reaching for. Accepted.

The extra astral-character test for requirement 10 (UTF-16 rather than byte columns across all three responses) is a good addition — and the fact that it passed without new encoding code, because `LineIndex` already handled it, is worth knowing.

### Two operational notes, both useful

**The scratchpad collision is a real finding.** `gh pr create` posting PR #85's body because a concurrent worker overwrote the shared path between your write and gh's read is a genuine hazard of parallel workers on one machine — and you caught it and corrected the body rather than leaving a wrong description on the PR. I have confirmed #88's current body is correct. Unique filenames per worker is the right fix; I am passing that on.

**Leaving the pre-existing biome nit alone was correct.** That formatting issue in the `signatureHelp` handler comes from base PR #85, and mixing a base-branch fix into this diff would have muddied both. It is already being fixed on #85 itself, so it will resolve when this stack rebases.

### Stack position

This is third in the lsp stack: #78 → #85 → #88. Merge bottom-up. Note #85's `validate` was red on the biome nit and is being fixed; #88 inherits a green base once that lands.

Nothing blocking.

PR #89, opened 2026-08-12
lexer: Use the reserved modifier slot for raw strings

### LINKED ISSUE #13: lexer: Use the reserved modifier slot for raw strings
## Context

Silk literal introductions are closed and extensible by design. The specification names
this requirement "String literal introductions are closed and extensible"
(`openspec/specs/bootstrap-lexer/spec.md:216-224`). The closed modifier vocabulary today
has two members: no modifier for text, and `b` for bytes
(`packages/compiler/src/LiteralForm.ts:33-38`). This ticket adds `r` for a raw text
literal. The extension point already exists.

## Current behavior

The lexer reserves every unknown modifier. `LiteralForm.recognize` returns nothing for
`r"..."`, then `LiteralForm.recognizeUnknown` recognizes the adjacent identifier and the
quote as one reserved introduction (`packages/compiler/src/LiteralForm.ts:81-93`). The
lexer emits an `InvalidStaticLiteral` token and the diagnostic `LEX0002`
(`packages/compiler/src/Lexer.ts:305-306`).

The specification scenario tests exactly this path with the spelling `future"value"`
(`openspec/specs/bootstrap-lexer/spec.md:236-239`). The reservation is deliberate. It
keeps the modifier slot free for a later form.

Today the user escapes every backslash:

```silk
const pattern: string = "\\d+\\.\\d+"
```

## Requirements

1. The modifier vocabulary must add `r` for a raw text literal.
2. A raw literal must accept one quote or three quotes, as the other forms do.
3. A raw literal must treat a backslash as an ordinary content byte.
4. A raw literal must close at the first unescaped run of its delimiter.
5. A raw literal must have the token kind of a text literal and the type `string`.
6. A raw literal must contain valid UTF-8, as a text literal does.
7. An unterminated raw literal must produce one lexical diagnostic.
8. The spelling `r` must stay a valid identifier when it is not adjacent to a quote.
9. The reserved-modifier diagnostic must stay for every other adjacent identifier.

## Example

```silk
const decimalPattern: string = r"\d+\.\d+"
const windowsPath: string = r"C:\Users\build\output"

const helpText: string = r"""
Usage: silk build [options]
  --target \path\to\dir
"""
```

## Out of scope

- **String interpolation.** A modifier such as `f"..."` or an embedded `${...}` form is a
  separate question. It changes the parser and the type checker, not the lexer alone. A
  later ticket must decide it.
- A raw byte-string literal such as `rb"..."`. The modifier vocabulary stays one byte
  wide in this ticket.
- A variable delimiter count, such as the `r#"..."#` form.
- Any change to the escape rules of the existing text and byte-string forms.

## Implementation note

The escape policy is already a separate field. `LiteralForm.EscapePolicy` has one member,
`Escaped` (`packages/compiler/src/LiteralForm.ts:10`), and the field sits beside the
delimiter width. The raw form adds a second member, `Raw`. This keeps the escape policy
independent from the delimiter width, as the existing type states.

`scanBoundary` must take the escape policy. Today it skips two bytes after a backslash
(`packages/compiler/src/LiteralForm.ts:119-126`). A raw literal must not skip. Without
this change, `r"path\"` closes at the wrong byte.

The `forms` array is ordered by longest introduction first
(`packages/compiler/src/LiteralForm.ts:33-38`). The new entries must keep this order, so
`r"""` must come before `r"`.

`recognizeUnknown` already skips a spelling that is in the vocabulary
(`packages/compiler/src/LiteralForm.ts:90`). After `r` joins the vocabulary, that check
covers it with no further change.

## Dependencies

None.

## Acceptance criteria

- [ ] The lexer produces one text literal token for `r"\d"` instead of `LEX0002`.
- [ ] A lexer test shows `r"..."` and `r"""..."""` as valid literals.
- [ ] A test shows that the value of `r"\n"` is two bytes, not one.
- [ ] A test shows that the value of `"\n"` stays one byte.
- [ ] A lexer test shows that `future"value"` still produces `LEX0002`.
- [ ] A lexer test shows that `r` alone stays an identifier token.
- [ ] A lexer test shows that an unterminated `r"` produces one diagnostic.
- [ ] `openspec/specs/bootstrap-lexer/spec.md:216-224` gets an amendment, because the
      requirement names the closed vocabulary as unmodified and `b`-modified only.

julia-script:
Closes #13

Stacked on #80 (`agent/63-malformed-float-exponent`).

## Acceptance criteria

- [x] The lexer produces one text literal token for `r"\d"` instead of `LEX0002`.
- [x] A lexer test shows `r"..."` and `r"""..."""` as valid literals.
- [x] A test shows that the value of `r"\n"` is two bytes, not one.
- [x] A test shows that the value of `"\n"` stays one byte.
- [x] A lexer test shows that `future"value"` still produces `LEX0002`.
- [x] A lexer test shows that `r` alone stays an identifier token.
- [x] A lexer test shows that an unterminated `r"` produces one diagnostic.
- [x] `openspec/specs/bootstrap-lexer/spec.md` gets an amendment, because the requirement names the closed vocabulary as unmodified and `b`-modified only.

## Notes

`EscapePolicy` gains a second member rather than a new axis, as the issue's implementation note describes. `scanBoundary` now takes the policy and skips the two-byte backslash advance for a raw body, so `r"path\"` closes at its own quote. `StaticText.decode` passes a raw body through while keeping multiline CRLF normalization, which belongs to the delimiter width rather than the escape policy.

Two amendments beyond the one the issue names, both because the existing text became false:

- `bootstrap-static-text` said raw behavior "SHALL require a separately recognized future modifier" — that modifier is now `r`, and the raw decoding rule is stated.
- `bootstrap-language-pressure-programs` requires the differential corpus to cover recognized literal modifiers, which now includes `r`.

`examples/language-pressure/lexer/main.silk` is updated in lockstep with a new `raw-literals` corpus case, so `LexerPressure` checks the raw boundary rule across all three engines.

Tests: baseline 0 failures (measured in step 0), after 0 failures, +8 new tests

### DIFF
--- a/examples/language-pressure/lexer/main.silk
+++ b/examples/language-pressure/lexer/main.silk
@@ -84,7 +84,8 @@ fn isLineCommentStart(source: &[u8], index: usize) -> bool {
 
 fn isLiteralStart(source: &[u8], index: usize) -> bool {
   if byteAtOrZero(source, index) == 34 { return true }
-  return hasPair(source, index, 98, 34)
+  if hasPair(source, index, 98, 34) { return true }
+  return hasPair(source, index, 114, 34)
 }
 
 fn hasTripleQuote(source: &[u8], index: usize) -> bool {
@@ -260,7 +261,8 @@ fn scanLiteral(
   modifierWidth: usize,
   delimiterWidth: usize,
   kind: u8,
-  alwaysInvalid: bool
+  alwaysInvalid: bool,
+  raw: bool
 ) -> Decision {
   let mut index = start + modifierWidth + delimiterWidth
   while index < source.length {
@@ -283,7 +285,9 @@ fn scanLiteral(
         return Decision { kind: kind, end: index + 3, invalid: false }
       }
     }
-    if byte == 92 {
+    let mut escaping = byte == 92
+    if raw { escaping = false }
+    if escaping {
       if index + 1 < source.length {
         let escaped = source[index + 1]
         if delimiterWidth == 1 {
@@ -370,14 +374,19 @@ fn nextToken(source: &[u8], start: usize) -> Decision {
   if isLiteralStart(source, start) {
     let mut kind = u8.add(7, 0)
     let mut modifierWidth = usize.add(0, 0)
+    let mut raw = false
     if byte == 98 {
       kind = u8.add(8, 0)
+      modifierWidth = usize.add(1, 0)
+    }
+    if byte == 114 {
+      raw = true
       modifierWidth = usize.add(1, 0)
     }
     let quoteStart = start + modifierWidth
     let mut delimiterWidth = usize.add(1, 0)
     if hasTripleQuote(source, quoteStart) { delimiterWidth = usize.add(3, 0) }
-    return scanLiteral(source, start, modifierWidth, delimiterWidth, kind, false)
+    return scanLiteral(source, start, modifierWidth, delimiterWidth, kind, false, raw)
   }
   if isIdentifierStart(byte) {
     let end = scanIdentifier(source, start)
@@ -390,7 +399,8 @@ fn nextToken(source: &[u8], start: usize) -> Decision {
         end - start,
         delimiterWidth,
         tokenInvalidStaticLiteral,
-        true
+        true,
+        false
       )
     }
     return Decision { kind: keywordKind(source, start, end), end: end, invalid: false }
--- a/openspec/specs/bootstrap-language-pressure-programs/spec.md
+++ b/openspec/specs/bootstrap-language-pressure-programs/spec.md
@@ -29,8 +29,8 @@ lexer-specific or token-specific operations, layouts, or branches.
 
 The Silk lexer SHALL be differentially checked against the TypeScript lexer, which remains the
 canonical implementation. The corpus SHALL cover whitespace, comments, identifiers, every current
-keyword, decimal integer and float forms, single-line and multiline text and byte literals with
-valid and malformed escapes, recognized and unknown literal modifiers, physical LF and CRLF,
+keyword, decimal integer and float forms, single-line and multiline text, byte, and raw literals
+with valid and malformed escapes, recognized and unknown literal modifiers, physical LF and CRLF,
 terminated and unterminated delimiters, every current single and compound punctuation token,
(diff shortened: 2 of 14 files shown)


julia-script:

**The corpus addition is the most important thing here, and I verified it.** The `LexerPressure` corpus contained **zero** `r"` inputs before this PR and one after — so without that case the differential test would have passed while exercising none of the new lexing. A green differential test that tests nothing is worse than a missing one, because it reads as coverage. Catching that rather than relying on the existing corpus is the difference between real and nominal verification.

**The three spec amendments are correct, not scope creep.** The issue named only `bootstrap-lexer`, but `bootstrap-static-text` stated that raw escape behavior "SHALL require a separately recognized future modifier" — that modifier is now `r`, so leaving it would have left a spec sentence that is simply false. And `bootstrap-language-pressure-programs` requires the differential corpus to cover recognized literal modifiers, which is exactly what the corpus addition satisfies. Under a `spec-change` label, keeping the specs true is the work. Keep all three.

The `TextMate.ts` follow-on was likewise unavoidable rather than optional: the grammar derives from `LiteralForm.forms`, so it picked up raw rules automatically and *needed* the escape-pattern suppression and backslash-boundary guard, or it would have highlighted raw literals wrongly. Good catch on the hardcoded `storage.modifier.byte.silk` scope name applying to every modified form — that was a latent bug in the generator, not something this issue created.

### The Example in the issue does not compile — filed as #90

I confirmed this independently. `analyzeConstant` (`Elaboration.ts:1240-1252`) guards on `'the declared type must be one primitive scalar'` and has branches for `bool` and integer/float spellings only; there is no `string` branch. So:

```silk
const decimalPattern: string = r"\d+\.\d+"   // SEM0086
```

— the motivating example for this very feature — is rejected. It predates raw strings and affects escaped literals equally, so **correctly excluded from this PR**. But it means named string constants, the most idiomatic use of raw strings (regex patterns, path fragments, format templates), are unavailable at module scope. Tracked at **#90**; using `let` bindings in the acceptance tests was the right workaround.

### Verified

Baseline clean → 141 files / 1092 tests, 0 failures; typecheck and biome clean; the regenerated `silk.tmLanguage.json` is pinned identical by a test. Nothing blocking.

Issue #90, opened 2026-08-12
type-system: a module-level const cannot declare a string type


julia-script:
## Current behavior

A module-level `const` with a `string` declared type is rejected with `SEM0086` ("the literal kind does not match string"). `analyzeConstant` (`packages/compiler/src/Elaboration.ts:1240`) accepts only primitive scalars — its first guard sets `detail = 'the declared type must be one primitive scalar'`, and the branches below it handle `bool` and integer/float spellings only. There is no `string` branch.

```silk
const decimalPattern: string = r"\d+\.\d+"   // SEM0086
```

A `let` binding with the same initializer is fine, so this is specific to `const`.

## Why this matters

This predates raw strings and affects escaped string literals identically — it is not a raw-string bug. It surfaced because the **Example block in #13 does not compile**:

```silk
const decimalPattern: string = r"\d+\.\d+"
```

That is the motivating example for raw strings in that issue, and a reader copying it gets a diagnostic. PR #89 had to write its acceptance tests with `let` bindings instead. Named string constants are also the natural way to use raw strings — regex patterns, path fragments, format templates — so the feature's most idiomatic use is currently unavailable at module scope.

## Requirements

1. A module-level `const` may declare type `string` and take a string literal initializer, escaped or raw.
2. The stored constant preserves the literal's decoded bytes, matching what the equivalent `let` binding produces.
3. Referencing such a constant works wherever a `string` value is accepted.
4. Whatever restriction remains (for example, initializer must be a literal rather than an expression) produces a diagnostic that says so, rather than the current "must be one primitive scalar".

## Acceptance criteria

- [ ] A test compiles `const pattern: string = "\\d+"` and asserts the referenced value equals the equivalent `let` binding.
- [ ] A test does the same for a raw literal, `const pattern: string = r"\d+"`.
- [ ] A three-engine acceptance test reads a module-level string constant.
- [ ] The Example block in #13 compiles as written.

blocked_by: none

---

Found by the worker implementing #13 (PR #89) and independently confirmed against `Elaboration.ts:1240-1252`. Correctly left out of that PR — it is pre-existing and orthogonal to the modifier slot.

julia-script:
🔒 Claimed by orchestrator run `run-r7x1` at 2026-08-12T22:01:35Z. Lane: type-system. Worker dispatched.

[...]

Issue #86, opened 2026-08-12
backend: Wasm stores element 0 of a Vector of move-only structs as zero


julia-script:
## Current behavior

Appending a **move-only** struct to a `Vector<T>` under the **WebAssembly backend** stores element **index 0** as all-zero. Every later index is correct. The evaluator and native/LLVM both produce the right value, so this is a three-engine disagreement in which Wasm alone is wrong — and it is a **silent wrong answer**, not a trap or a crash.

Reproduced independently on clean `main` (`654d79f`):

```silk
import silk.vector { Vector, make, append, asSlice }

struct Entry {
  value: i32
  marker: Vector<i32>
}

effect fn build() -> i32 ! OutOfMemory {
  let mut allocator = SystemAllocator.make()
  let mut values = make<Entry>()
  let entry = Entry { value: 3, marker: make<i32>() }
  let pending = append<Entry>(&mut values, move entry) |> Effect.provideMut(&mut allocator)
  let appended = run pending
  let readable = asSlice<Entry>(&values)
  return 100 + readable[0].value
}

effect fn recover(error: OutOfMemory) -> i32 { return 0 }
pub fn main() -> i32 { return run Effect.catch(build(), recover) }
```

| engine | result |
|---|---|
| evaluator | **103** ✅ |
| native / LLVM | **103** ✅ |
| wasm debug | **100** ❌ |
| wasm release | **100** ❌ |

## What narrows it

- **A move-only element type is required.** `i32`, an all-scalar struct, and a struct with a nested *Copy* struct all pass. A same-size Copy struct (`i32` + 3×`usize`) also passes, which rules out size and layout.
- **Not `Vector`-specific as the field type.** A `Bytes` field triggers it identically — any move-only field does.
- **Field order is irrelevant.** Move-only field first or second, same failure.
- **Only index 0 is affected.** Appending 1, 2, 3 reads back as 0, 2, 3. Five appends (forcing 0→4→8 growth) read 0, 2, 3, 4, 0. Lengths are always correct.
- **`asSlice` and `asMutSlice` behave the same**, so the access form does not matter.
- **Debug and release Wasm are identical** — not an optimizer artifact.

## Localization: this is a write-time defect, not a read-time one

Two independent observations place the fault in the `append` store path rather than the access path:

[...]

Issue #81, opened 2026-08-12
docs: Wire documentation:check into CI so generated pages cannot drift


julia-script:
## Current behavior

PR #77 adds `packages/compiler/scripts/generate-documentation.mjs`, which generates two of the four language documentation pages from the compiler's own models:

- `packages/language/docs/stdlib.md` — every stdlib module, built from the `@silk-effect/documentation` project model that backs `silk doc`, so each signature and description is the `///` comment on the declaration.
- `packages/language/docs/diagnostics.md` — every diagnostic code declared in `Diagnostic.ts`.

The script has a `--check` mode that fails on drift, and it was verified to fail correctly. **It is not wired into `pnpm test` or CI.**

## Why this matters

Both pages are generated, so nothing stops them going stale. Adding a stdlib module or a diagnostic code leaves the committed page describing a compiler that no longer exists, and no test or CI job notices. The failure is silent and the drift compounds — exactly the rot that made the pre-existing `apps/docs/content/` documentation describe the LLVM builder API long after that stopped being what users needed.

PR #77 already added `DocumentationExamples.test.ts`, which asserts that every stdlib module and every declared diagnostic code *appears* in the generated pages. That catches an entirely missing entry but not a changed signature, a reworded description, or a stale `///` comment.

The counts in issue #52 had themselves drifted before anyone noticed — it asked for 27 modules and 117 codes when the real numbers were 28 and 119. That is the failure mode this closes.

## Requirements

1. Wire `documentation:check` into the compiler package's `test` script, or into the `validate` CI job, so generated-page drift fails the build.
2. The failure message must name the regeneration command, so the fix is obvious from the CI log alone.

## Acceptance criteria

- [ ] Adding a diagnostic code or a stdlib module without regenerating fails `pnpm --filter @silk-effect/compiler test` (or `validate`).
- [ ] The failure output names the exact command that regenerates the pages.
- [ ] A clean checkout with no changes passes.

blocked_by: none

---

Filed by orchestrator run-b4c2 as a follow-up to #77, where the generator was added but deliberately left unwired. Small and self-contained — a reasonable `good first issue`.

julia-script:
🔒 Claimed by orchestrator run `run-r7x1` at 2026-08-12T22:00:26Z. Lane: tooling. Worker dispatched.

**This issue just demonstrated its own value.** While you were being dispatched, `main` went red at `8221f5a` on exactly the drift this issue is about:

```
[...]

PR #92, opened 2026-08-12
ownership: a generic effect body leaks any owner held across a fallible run

### LINKED ISSUE #68: ownership: a generic effect body leaks any owner held across a fallible run
Isolated while implementing #27 (PR #67). **Pre-existing.** PR #67 removes `provideWith` from the blast radius but deliberately does not fix the underlying classification, which is still open.

## The defect

`fallibleRunSites` (`packages/compiler/src/Ownership.ts:839-852`) classifies a run as able to propagate only when its effect type has `failures.length > 0`.

Inside a **generic** combinator the caller's failures arrive as a row *parameter*: `failures` is empty while `failureParameters` is not. The run is therefore treated as infallible, no `Propagation` exit is published for it, and owners live at that run get no releases attached. After specialization substitutes a non-empty row and the run really can fail, those owners are simply gone.

**Any generic effect body that holds an affine owner across a fallible run leaks it.** `provideWith` was the one combinator in the stdlib that did.

## Measured evidence

With a provider owning a heap block, the allocation trace on `main` before PR #67:

| scenario | trace |
| --- | --- |
| two successful executions | `Acquire, Release, Acquire, Release` |
| one failing execution | `Acquire` |
| retry, three failing attempts | `Acquire, Acquire, Acquire` |

Every provider acquired by an execution that ended in typed failure leaked.

The mechanism was confirmed by bisecting the generic parameters of a user-space replica: the same body with a **concrete** failure row releases correctly; with a failure **row parameter** it leaks. The lowered MIR for the two is identical apart from the instance's type arguments — the only difference is the missing propagation release list on the `RunEffectValue`.

## Why the obvious fix is its own ticket

Counting an open failure row as fallible is sound and does fix `provideWith`. It was tried and reverted, because it publishes propagation cleanup at generic run sites across the whole codebase and immediately hits an unimplemented backend path:

> `WasmBackend.ts:1217` — "Wasm cleanup does not yet lower hook-bearing union cases"

That took the suite from the 3 pre-existing host-triple failures to **14**, across `LexerPressure`, `StackVmPressure`, `MultiAffineReturn`, `SynchronousEffectCost`, and two determinism suites.

So this needs two pieces, and they may warrant splitting:

1. **Wasm backend:** lower cleanup for hook-bearing union cases (prerequisite).
2. **Ownership:** treat a run whose effect type carries a non-empty `failureParameters` as fallible, so propagation exits and release lists are published for generic bodies.

## Why it matters beyond `provideWith`

The stdlib currently has only one combinator in this shape, so the practical blast radius is small *today*. But the rule is a trap for any future generic effect combinator holding an owner across a run — including `Effect.ensuring` (#26), whose whole purpose is running a finalizer on every exit. Landing #26 on top of this classification would reintroduce exactly this leak.

## Acceptance criteria

- [ ] A generic effect body holding an owner across a run that fails with a specialized non-empty failure row releases that owner.
- [ ] An allocation-trace test asserts acquire count equals release count for the generic case, mirroring the `provideWith` tests added in PR #67.
- [ ] The Wasm backend lowers cleanup for hook-bearing union cases, or the ownership change is gated until it does.
- [ ] `LexerPressure`, `StackVmPressure`, `MultiAffineReturn`, `SynchronousEffectCost`, and the determinism suites stay green.
- [ ] PR #67's `provideWith` restructuring can be simplified back toward the straightforward shape once this lands (optional follow-up, not required).

julia-script:
Closes #68

## What was wrong

The issue's diagnosis is confirmed, and it is half of the fix.

`fallibleRunSites` (`Ownership.ts`) classified a run as able to propagate only when its effect type had `failures.length > 0`. Inside a generic combinator the caller's failures arrive as a row *parameter* — `failures` is empty while `failureParameters` is not — so the run counted as infallible, no `Propagation` exit was published, and any owner live at that run was stranded once specialization substituted a row that really fails.

Measured on `main` with a generic body that acquires a provider and holds it across a run whose failure channel is a row parameter:

| scenario | trace |
| --- | --- |
| one failing execution | `Acquire` |
| retry, three failing attempts | `Acquire, Acquire, Acquire` |

which reproduces the issue's table exactly.

## Why the obvious fix double-released, and what the real predicate is

Counting an open failure row as fallible is necessary but **not sufficient** — on its own it traps, exactly as the earlier investigation measured on `GenericRunReleaseAcceptance`'s `holding`.

The reason is a second, independent defect that the first one was masking. The propagation exit is published *before* the statement is walked, so the live set still holds every owner the run's own operands move away. For `run f(move owned)`, ownership therefore published a release of `owned` in the **caller**, even though `owned` had just been moved into the callee.

That caller release was always wrong. It was merely invisible while the callee published no release of its own. Fix the callee, and the pre-existing caller release becomes a double free.

So the actual predicate the issue asked for — what distinguishes `provideWith` (leaks) from `holding` (does not) — is not a property of the run's failure row at all. It is whether the owner is *acquired inside* the body (genuinely stranded, needs the release) or *moved in through the run's own operands* (already owned by the callee, must not be released again). Excluding the run's consumed operands from its propagation exit handles both, at the one predicate every caller routes through.

`holding` is untouched by this PR's behaviour and its test stays green; the release it always had now comes from the callee rather than the caller.

## Wasm backend

Making ownership correct reaches the path that previously threw at `WasmBackend.ts:1217` ("Wasm cleanup does not yet lower hook-bearing union cases"). Both cleanup walkers now lower `UnionCleanup`: each hook-bearing case runs guarded on the union's own tag (`i32.load` at the union base, compare to the case ordinal, run that case's cleanup under `ifElse` at the payload offset), matching the LLVM backend's semantics. It is exercised by the new tests rather than landing untested.

## Acceptance criteria

- [x] A generic effect body holding an owner across a run that fails with a specialized non-empty failure row releases that owner.
- [x] An allocation-trace test asserts acquire count equals release count for the generic case, mirroring the `provideWith` tests added in PR #67.
- [x] The Wasm backend lowers cleanup for hook-bearing union cases, or the ownership change is gated until it does.
- [x] `LexerPressure`, `StackVmPressure`, `MultiAffineReturn`, `SynchronousEffectCost`, and the determinism suites stay green.
- [ ] PR #67's `provideWith` restructuring can be simplified back toward the straightforward shape once this lands (optional follow-up, not required — left alone here).

## Tests

`GenericRunAcquiredReleaseAcceptance.test.ts` covers the shape that actually leaks — an owner *acquired inside* the generic body — on the failing path, the succeeding path, and across three retried attempts, asserting the interleaved acquire/release trace and that the counts balance, on both the evaluator and Wasm.

Verified red before green: with the source fix reverted, the failing case traces `[Acquire]` and the retried case `[Acquire, Acquire, Acquire]`.

Tests: baseline 0 failures (1079 passing, measured in step 0), after 0 failures (1082 passing), +3 new tests. Full `pnpm check` green.

### DIFF
--- a/packages/compiler/src/Ownership.ts
+++ b/packages/compiler/src/Ownership.ts
@@ -835,6 +835,11 @@ const deferredBlocks = (
 /**
  * Run sites in this expression that can propagate a typed failure, stopping at effect blocks:
  * a deferred body's runs propagate out of its own compiled function, not out of this one.
+ *
+ * A run is fallible when its effect type carries failures OR an open failure row: inside a
+ * generic body the caller's failures arrive as a row *parameter*, so `failures` is empty while
+ * `failureParameters` is not, and specialization can substitute a row that really does fail.
+ * Reading `failures` alone leaves every owner live at such a run without a release.
  */
 const fallibleRunSites = (
   expression: Hir.Expression,
@@ -847,8 +852,24 @@ const fallibleRunSites = (
     typeof subjectType === 'object' &&
     subjectType !== null &&
     (subjectType as { readonly _tag?: string })._tag === 'EffectType' &&
-    (subjectType as Type.Effect).failures.length > 0
+    ((subjectType as Type.Effect).failures.length > 0 ||
+      (subjectType as Type.Effect).failureParameters.length > 0)
   return fallible ? [expression, ...nested] : nested
+}
+
+/**
+ * Owner sites this expression consumes by move, keyed like the liveness set.
+ *
+ * A propagation exit is published before the statement is walked, so the live set still holds
+ * every owner the run's own operands move away — `run f(move owned)` would otherwise release
+ * `owned` here as well as in the callee that now owns it. Excluding the operands keeps the
+ * exit to the owners that genuinely survive the run and are stranded by the failure.
+ */
+const consumedSites = (expression: Hir.Expression): ReadonlyArray<string> => {
+  const nested = Hir.expressionChildren(expression).flatMap(consumedSites)
+  if (expression._tag !== 'Move') return nested
+  const site = useSite(expression.subject)
+  return site === undefined ? nested : [siteKey(site), ...nested]
 }
 
 interface LoanAnalysis {
@@ -1975,11 +1996,12 @@ const checkFunction = (
       // still live at that point must release on the way out. The exit is keyed by the run
       // expression's span so lowering can attach the releases to the run operation.
       for (const run of statementRootExpressions(statement).flatMap(fallibleRunSites)) {
+        const consumed = new Set(consumedSites(run.subject))
         exits.push(
           Object.freeze({
             kind: 'Propagation' as const,
             span: run.span,
-            sites: frameSitesInnerFirst(frames, live),
+            sites: frameSitesInnerFirst(frames, live).filter((site) => !consumed.has(site)),
           }),
         )
       }
--- a/packages/compiler/src/WasmBackend.ts
+++ b/packages/compiler/src/WasmBackend.ts
@@ -1212,9 +1212,29 @@ const emitOperation = (
             walk(plan_.element, byteOffset + index * representation.stride),
           ).flat()
         }
-        case 'UnionCleanup':
+        case 'UnionCleanup': {
           if (plan_.cases.every((entry) => !planHasHook(entry.cleanup))) return []
-          throw new RangeError('Wasm cleanup does not yet lower hook-bearing union cases')
+          const entry = LayoutPlan.entry(memory.plan, plan_.type)
+          const representation = entry?.representation
+          if (representation?._tag !== 'Union') return []
+          // Only the live case owns the payload, so each hook-bearing case runs guarded on the
+          // union's own tag. The tag sits at the union's base and the payload at its offset.
+          return plan_.cases.flatMap((caseEntry) =>
+            planHasHook(caseEntry.cleanup)
+              ? [
+                  ...frameAddress(planned.offset + byteOffset),
+                  Instr.memoryAccess('i32.load', memory.memory),
+                  Instr.i32Const(caseEntry.ordinal),
+                  Instr.op('i32.eq'),
+                  Instr.ifElse(
+                    Instr.emptyBlockType,
+                    walk(caseEntry.cleanup, byteOffset + representation.payloadOffset),
+                    [],
+                  ),
+                ]
+              : [],
+          )
+        }
         default:
           return []
       }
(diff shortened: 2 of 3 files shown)


julia-script:
## Review — orchestrator run-b4c2: verified, approving

I reproduced the red-before-green claim rather than trusting it. Reverting `Ownership.ts` and `WasmBackend.ts` to `origin/main` while keeping the new tests:

```
× releases an owner acquired inside a generic body across a failing run
  expected [ 'AllocationAcquire' ] to deeply equal [ 'AllocationAcquire', 'AllocationRelease' ]

× gives every retried attempt of a generic body its own released owner
  actual:   [ 'AllocationAcquire', 'AllocationAcquire', 'AllocationAcquire' ]
  expected: [ …with three releases ]
```

Three acquires, zero releases. That is the leak, exactly as the issue's table predicted. With the fix restored: **35/35 green** across the new suite and `Ownership.test.ts`.

### The analysis here is the valuable part

The issue's diagnosis was correct **and insufficient**, and finding that out before shipping is what makes this PR trustworthy. Counting an open failure row as fallible is necessary but on its own **converts a leak into a double free** — because the propagation exit is published before the statement is walked, so the live set still holds owners that the run's own operands move away. For `run f(move owned)`, ownership published a release of `owned` in the *caller* despite `owned` having just been moved into the callee.

That caller release was **always wrong**; it was invisible only because the callee published no release of its own. Fixing the callee alone would have surfaced a latent double free as a new bug, and the natural conclusion would have been "the fix is wrong" rather than "a second defect was hiding behind the first."

The resulting predicate is also the right one conceptually: what separates `provideWith` (leaks) from `holding` (does not) is **not** a property of the run's failure row — it is whether the owner is *acquired inside* the body (stranded, needs release) or *moved in through the run's operands* (already the callee's, must not be released again). Excluding the run's consumed operands from its propagation exit handles both at the single predicate every caller routes through. That is a root-cause fix, not a patch on the reported path.

### On the Wasm cleanup work

Making ownership correct reaches the path that previously threw at `WasmBackend.ts:1217` ("Wasm cleanup does not yet lower hook-bearing union cases"). Lowering `UnionCleanup` in both walkers — guarded on the union's own tag, matching LLVM's semantics — is unavoidable scope rather than creep: without it the corrected ownership would crash the backend. Landing it exercised by the new tests rather than untested is the right call.

### Notes

- The unticked criterion (simplifying PR #67's `provideWith` restructuring) is correctly left alone — it is explicitly optional and belongs in its own change.
- Verifying that `holding` stays green, with its release now coming from the callee rather than the caller, is the check that proves the predicate change did not just move the bug.
- Baseline 0 → 0 with 1079 → 1082 passing, full `pnpm check` green, matches my own measurement of this repo.

Nothing blocking. This closes the leak that #26 (`Effect.ensuring`) was waiting on — that issue can now be reconsidered on a correct classification.

julia-script:
## Re-verified against updated `main` — still good to merge

`main` moved substantially since this branched: the nine run-b4c2 PRs merged, plus fixes for #86, #81, and #90. Since #86's fix (`1fcff63`, "reload only the frame roots a call can write through") also edits `WasmBackend.ts`, this PR's `UnionCleanup` lowering overlaps it, so I re-checked rather than trusting the mergeable flag.

- Merging current `origin/main` into this branch: **no conflicts**.
- On the merged tree, `GenericRunAcquiredReleaseAcceptance`, `VectorAcceptance`, and `Ownership`: **58/58 passing**.

The two Wasm changes are complementary rather than competing — #86 fixed the frame-root reload on the store path, this one adds cleanup lowering for hook-bearing union cases. No rebase needed.

Still the last open PR from run-b4c2, and still the one I'd merge first: it closes a real leak, and I verified its regression tests red before the fix (three acquires, zero releases) and green after.

Issue #102, opened 2026-08-13
docs: terminal-session test is flaky — output-consumer count races (expected +0 to equal 1)


julia-script:
## Evidence

`apps/docs/test/terminal-session.test.ts:154` — `shares one output consumer and follows default atom lifetime in a fresh registry` — failed `validate` on PR #101 (run 31653206003) with `AssertionError: expected +0 to equal 1`, on a diff that touches only the compiler, lexer-pressure corpus, and documentation pages.

Reproduced as flaky locally: **1 failure in 5 runs** on the PR merge tree, 0 in 3 on clean `main`, same command (`vitest run test/terminal-session.test.ts`), same machine, no source change between runs.

The assertion counts registered output consumers immediately after session setup, which races whatever async registration the atom lifetime does. A count of `0` then means "not yet", not "never".

## Why it matters

This test just cost a green PR a full CI round. A flaky gate on `validate` taxes every merge and trains people to re-run rather than read failures — the worst property a CI gate can have.

## Requirements

1. Find the race — either await the registration the test depends on, or make the consumer registration synchronous in a fresh registry.
2. The fix must not paper over it with a retry or a sleep; the test should be deterministic by construction.
3. If the race is real product behavior (a window where the consumer genuinely is not yet attached), that is a bug worth its own issue, and this test should pin the fixed behavior.

## Acceptance criteria

- [ ] 50 consecutive local runs pass.
- [ ] The fix is an ordering/synchronization change, not a timeout increase.

---

Filed by orchestrator run `run-r7x1` while triaging PR #101's CI failure.

Issue #43, opened 2026-08-12
stdlib: Decide whether OsString and PlatformPath are necessary


julia-script:
## Context

This is a discussion ticket. It asks a question and does not describe work. No one must
write code for it until the question has an answer.

Silk has one path type today: `Path` in `filesystem.silk:17`. `Path` holds UTF-8 bytes and
gives a borrowed `string` view (`filesystem.silk:309`). This is a portable design, and it is
correct for every case the standard library serves now. The question is whether it stays
correct.

## The question

Some platforms accept a file name that is not valid UTF-8. A Unix file name is a byte
sequence with no encoding guarantee. A Windows file name is a UTF-16 sequence that can hold
an unpaired surrogate. Neither case converts to a Silk `string` without a loss.

Rust answers this with `OsString` and `PathBuf`. Those types add real cost: a second text
type, a conversion at every boundary, and a fallible view for every read. The question is
whether Silk must pay that cost.

## Why the answer waits for ChildProcess

The filesystem does not force the decision. A program that reads a file it wrote, or a file
that a tool made, sees valid UTF-8 in practice. A program can also refuse an invalid name
and return the existing `invalidPath` failure (`filesystem.silk:120`).

Process arguments do force the decision. A child process receives an argument vector from
the parent, and the parent did not choose those bytes. An environment variable has the same
problem. If Silk cannot pass through an argument it received, then a Silk program cannot
wrap another program. #29 adds the ChildProcess service, so the decision belongs
there.

## The cases that decide it

Each case below must get a yes or a no answer before the decision closes.

1. A Silk program lists a directory that holds a file name with an invalid UTF-8 byte. Must
   the program see that entry, or is a failure acceptable?
2. A Silk program receives its own command-line arguments. Must it pass an argument through
   to a child process without a change?
3. A Silk program reads an environment variable that is not valid UTF-8. Must it read the
   exact bytes?
4. A Silk program builds a path from a name that a network protocol supplied. Must the
   program reject an invalid name, or store it?
5. Does the self-hosted compiler need any of these four cases? If it does not, the decision
   can wait longer.

## The options

1. **Keep `Path` only.** A name that is not valid UTF-8 gives the `invalidPath` failure.
   This is the cheapest option and it is the current design.
2. **Add a byte-level escape.** `Path` gets a constructor and a view that work on raw bytes.
   This adds no new type and covers the pass-through case.
3. **Add `OsString` and `PlatformPath`.** This is the complete answer and the expensive one.
[...]

Issue #29, opened 2026-08-12
stdlib: add a child-process service to spawn and wait for a program


julia-script:
## Context

A self-hosted Silk compiler must invoke Clang and the linker itself. Silk cannot run
another program today. `OsRuntime.ts` has no spawn, fork, or exec function of any kind.
This is a hard stage-2 self-hosting blocker.

## Current behavior

The native runtime offers 10 OS intrinsics, and every one of them is a file-system
operation. A grep for spawn, fork, execve, and child across `OsRuntime.ts` returns no
match. The `os_filesystem` module is the only host capability that reaches the OS.

`CONTEXT.md:343` already sketches the contract:

> The narrow bootstrap host capability for synchronously executing a program from
> structured input and returning structured output.

That entry names each part of the shape: requests give "an executable path, ordered
OS-string arguments, optional explicit working directory, exact environment, and closed
standard input"; results "own captured output and distinguish exit codes from signal
termination"; a nonzero exit "is result data, while failure to start, wait, or capture is
typed process failure"; and the service "never interprets a shell command string".

## Requirements

1. The stdlib must declare a `ChildProcess` service with one synchronous execute
   operation.
2. The request must carry an executable path, ordered arguments, an optional working
   directory, and an exact environment.
3. The service must close the child's standard input.
4. The result must own the captured standard output and the captured standard error.
5. The result must distinguish a normal exit code from termination by a signal.
6. A nonzero exit code must be result data, and must not be a typed failure.
7. A failure to start, to wait, or to capture must be a typed process failure.
8. The service must never interpret a shell command string.
9. The compiler must add the unsafe OS intrinsics that the native provider needs, because
   `OsRuntime.ts` has none today.
10. This work must add a new spec.

## Example

```silk
pub struct ProcessRequest {
  program: Path
  arguments: Vector
  workingDirectory: Option
  environment: Vector
}

pub type ProcessOutcome =
  Exited { code: i32 output: Bytes errors: Bytes }
  | Signaled { signal: i32 output: Bytes errors: Bytes }

pub service ChildProcess {
  effect fn execute(self: &mut Self, request: &ProcessRequest)
    -> ProcessOutcome ! ProcessFailure
}
```

## Out of scope

- A shell. The service never interprets a command string.
- Asynchronous or concurrent processes. The execute operation blocks.
- Pipes, streaming output, and interactive child input.
- Process groups, signals sent to the child, and job control.
- `OsString` and `PlatformPath`. Decide these when this ticket lands
[...]

Issue #28, opened 2026-08-12
stdlib: add a standard input service for reading bytes


julia-script:
## Context

`StandardStreams` writes only. Silk cannot read a byte from standard input. The current
exclusion is scoped to that one service: it "does not imply terminal control, color
detection, flushing, logging, or interactive input" (`CONTEXT.md:352-357`). That wording
points at a separate input service rather than an extension of `StandardStreams`.

## Current behavior

`StandardStreams` offers one operation, `writeAll`. Its `destination` parameter is a
`bool`: `stdout()` returns `false` and `stderr()` returns `true` (`core.silk:45-49`).

## Two shaping constraints

1. **The destination is a `bool`.** `core.silk:45-49` gives exactly two values, and
   neither one can name a read direction. Adding a read to `StandardStreams` means either
   a third value in a two-value type, or a second parameter that every write call must
   also pass.
2. **The write contract is all-or-nothing.** `writeAll` must "either commit the complete
   sequence or return a typed stream failure" (`bootstrap-standard-streams:22`). A read is
   inherently partial: the host can give fewer bytes than the caller asked for, and it can
   report end of input. A read therefore needs its own outcome shape.

Both constraints point to the same answer: a separate `StandardInput` service.

## Requirements

1. The stdlib must declare a `StandardInput` service that is separate from
   `StandardStreams`.
2. The read operation must return the count of bytes it wrote into the caller's buffer.
3. The read operation must report end of input as data, and not as a typed failure.
4. The read operation must report a host error as a typed failure.
5. The `StandardStreams` service must not change.
6. A native provider must read through the existing OS intrinsic pattern that
   `os_filesystem` uses.
7. This work must add a new spec, because `bootstrap-standard-streams` covers writes only.

## Example

```silk
pub service StandardInput {
  effect fn read(self: &mut Self, buffer: &mut [u8]) -> ReadOutcome ! StreamFailure
}

pub type ReadOutcome = Filled { count: usize } | EndOfInput
```

## Out of scope

- Terminal control, raw mode, and color detection.
- Line editing and prompt handling.
- A change to `StandardStreams` or to its `bool` destination.
- Non-blocking or asynchronous reads.

## Dependencies

- None.

## Acceptance criteria

- [ ] A test reads a known byte sequence from a test provider and asserts the count.
- [ ] A test asserts a read at the end of input returns `EndOfInput` and not a failure.
- [ ] A test asserts a partial read returns the true count and not the buffer length.
[...]

PR #111, opened 2026-08-13
stdlib: add a HostInput service for argv, environment, and cwd

### LINKED ISSUE #30: stdlib: add a HostInput service for argv, environment, and cwd
## Context

A self-hosted Silk compiler must receive source paths and flags from its command line.
Silk cannot read a command line, an environment variable, or the working directory. The
native entry point takes no arguments: `ToolchainPlan.ts:196` emits `int main(void)`. This
is a stage-2 self-hosting blocker.

## Current behavior

`ToolchainPlan.ts:196` emits this C entry point:

```c
int main(void) {
  const int tag = silk_main();
  if (tag == 0) return 0;
  ...
}
```

The signature is `main(void)`, so `argc` and `argv` never reach the program. Silk `main`
is public, non-generic, and takes zero parameters. There is no service that gives a
program its arguments.

## Why this touches the entry specification

The native entry shape must change to `main(int argc, char **argv)` so the runtime can
hold the arguments. Silk `main` itself keeps zero parameters, and the program reads the
arguments through a service. This changes the emitted entry point, so the work touches
`bootstrap-entry-termination`.

## Requirements

1. The stdlib must declare a `HostInput` service.
2. The service must give the ordered command-line arguments.
3. The service must give the value of a named environment variable.
4. The service must give the current working directory.
5. The native entry point must change from `main(void)` to a form that receives `argc` and
   `argv`.
6. Silk `main` must keep its zero-parameter, empty-requirement-row shape.
7. The exit-code contract must not change: 0 for success, 1 for a typed failure, 2 for a
   failed stderr write.
8. A program that does not use `HostInput` must behave exactly as it does today.
9. This work must amend `bootstrap-entry-termination` through the openspec process.

## Example

```silk
pub service HostInput {
  effect fn arguments(self: &Self) -> Vector ! OutOfMemory ? &mut Allocator
  effect fn variable(self: &Self, name: &String) -> Option ! OutOfMemory
  effect fn workingDirectory(self: &Self) -> Path ! HostInputFailure
}
```

## Out of scope

- Setting an environment variable or changing the working directory. This service reads
  only.
- `OsString` and `PlatformPath`. Decide these with #29
  (`FEATURE-MATRIX.md:448`).
- Argument parsing and flag grammar. These belong above the service boundary.
- The Wasm entry point, unless the target needs it.

## Dependencies

- #29 (`ChildProcess`). Both need the same OS-string decision. That decision is #43,
  which is open and states it waits for #29. The three are interlocked, so none of them
  can close until a maintainer settles `OsString` and `PlatformPath`.

## Acceptance criteria

- [ ] A test runs a native program with 3 arguments and asserts the service returns them
      in order.
- [ ] A test asserts a missing environment variable returns `None` and not a failure.
- [ ] A test asserts a program that does not use `HostInput` keeps the same exit codes.
- [ ] An `EffectEntry` test asserts the new entry shape still returns 0, 1, and 2
      correctly.
- [ ] An openspec change amends `bootstrap-entry-termination`.

julia-script:
Closes #30

## What this adds

A portable read-only `HostInput` service — `argumentCount`, `argument(index)`, `variable(name)`, `workingDirectory` — plus an `arguments` collector and the checked `text` view above it. It reads only: no operation sets a variable or changes the directory, per the issue's own scope.

Following the shape #28 (PR #97) established: portable contract module `silk/host_input`, native provider module `silk/os_host_input`, unsafe intrinsics mirroring `Intrinsic.osFileRead`'s `Option<usize>` + reason + native-code shape, a separately injected evaluator host, and Wasm rejected through the existing native-only restriction.

### The OsString question (#43, option 2)

Every value is the raw bytes the process received, unchanged. `text(values: &[u8]) -> Result<string, InvalidUtf8>` layers the checked, fallible view on top and never replaces or discards bytes it cannot decode. No new text type.

The working directory returns `Bytes`, not `Path`: `Path` is normalized UTF-8 confined to a *provider* root that never consults a process working directory, so returning a host-absolute location through it would either lie about the namespace or discard values a program is entitled to read. A program that wants a `Path` builds one through the existing checked constructors. This is the one place the implementation departs from the issue's illustrative example, and `design.md` records why.

An index at or past the argument count and an unset name are `None` — absence is an ordinary answer, the same reasoning that makes `EndOfInput` outcome data in `StandardInput`. `HostInputFailure` means only that the host could not answer at all.

### The entry point

The native shim's entry becomes `main(int argc, char **argv)` and stores both for the host-input runtime before calling `silk_main`. Silk `main` keeps its zero-parameter, empty-requirement-row shape, so arguments reach a program through a replaceable service rather than through a signature that could not be replaced in a test. Every termination status, report byte, and cleanup path is unchanged; a program that reads no host input pays two stores and no syscall. Direct Wasm is untouched — no shim, no entry parameters, no invented import.

`Some(n)` at the intrinsic boundary reports the value's *complete* byte length and copies the prefix that fits, so an undersized buffer costs one exactly sized second pass instead of a separate buffer-too-small protocol. The environment is scanned by raw bytes over the process block rather than through `getenv`, so an undecodable name or value is matched and returned exactly as received.

## Engine coverage

OS intrinsics target Evaluator + LLVM only, so three-engine criteria read here as evaluator/native parity plus the Wasm rejection test, as #97 did:

- **Evaluator** — a pure in-source provider and the injected memory host, including a blocked evaluation naming `MissingHostInput` when no host is supplied.
- **Native** — the lookups lower to their own reachable-only runtime symbols (`@silk_os_host_*_v1`, and no filesystem symbol), and a compiled binary run with `alpha beta gamma` and `SILK_EXIT_MARK=K` reads the real argv in order, the real environment, and the real cwd, then derives exit status 42 from the marker byte.
- **Wasm** — a reachable lookup on `wasm32-unknown-unknown` is rejected with `SEM0093`.

## Acceptance criteria

- [x] A test runs a native program with 3 arguments and asserts the service returns them in order — the compiled-binary test asserts `argumentCount` is 4 and that the collected sequence is program name, `alpha`, `beta`, `gamma`, in order.
- [x] A test asserts a missing environment variable returns `None` and not a failure — asserted through the in-source provider, the memory host, and the compiled binary; the surviving recovery branch proves the failure channel was never entered.
- [x] A test asserts a program that does not use `HostInput` keeps the same exit codes — `EffectEntry` now runs the reporting and success binaries *with* arguments and asserts `1` with identical report bytes, `2` on a closed stderr, and `0`.
- [x] An `EffectEntry` test asserts the new entry shape still returns 0, 1, and 2 correctly — above, plus a `NativeToolchain` test pinning `main(int argc, char **argv)` and the retained command line in both termination shapes.
- [x] An openspec change amends `bootstrap-entry-termination` — `openspec/changes/add-host-input-service/`, with the new `bootstrap-host-input` capability and deltas for entry termination, the intrinsic boundary, evaluation, and the stdlib; synced into `openspec/specs/`.

Regenerated `packages/language/docs/stdlib.md` and the checked intrinsic inventory; added the glossary entry and the stdlib README section.

## Tests

`Tests: baseline 4 failures, after 4 failures, +11 new tests`

The 4 are the known-environmental ones, identical before and after: `compiler-cli` `FormatCommand`/`FormatWorkflow` (3, chmod vs root) and `packages/wasm` `Extended.test.ts` "threads address types" (1). `packages/compiler` is 1164 passed + native acceptance 1 passed; `apps/docs` 89 passed, including the flaky `terminal-session` (#102).

New: 10 in `packages/compiler/test/HostInput.test.ts` and 1 in `NativeToolchain.test.ts`; `EffectEntry.test.ts` extends an existing test rather than adding one. Verified the regression guard fails pre-change: with `ToolchainPlan.ts` reverted, the new `NativeToolchain` test fails with `expected '/* silk-effect bootstrap runtime shim…' to include 'int main(int argc, char **argv) {'`.

## Note for merge sequencing

#29 (`ChildProcess`) is in flight in parallel and also touches `stdlib/manifest.json` and the generated stdlib docs, so expect a rebase there.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GZ8crzQ1kWrWARQWonuJBy)_


julia-script:
Merged `origin/main` (through #110, the ChildProcess service) into `agent/30-host-input` — pushed as `9d61097`.

Both PRs add an OS service by the same pattern, so all 14 conflicts were "both sides added" rather than "the two disagree". Nothing was dropped and nothing was redesigned; where the two services touch a common list, ChildProcess is ordered first and HostInput second, consistently.

### How each conflict was resolved

| File | Conflict | Resolution |
| --- | --- | --- |
| `apps/docs/app/labs/row/flow-model.ts` | both added a `BlockedReason` case in two switches | kept both cases in both switches |
| `apps/docs/app/labs/row/project-backend.tsx` | same, one switch | kept both cases |
| `openspec/specs/bootstrap-evaluation/spec.md` | both appended an injected-host requirement | kept both requirement blocks |
| `openspec/specs/bootstrap-intrinsic-boundary/spec.md` | both appended an admitted-primitives requirement | kept both |
| `openspec/specs/bootstrap-silk-stdlib/spec.md` | both appended a portable/native module-split requirement | kept both |
| `packages/compiler/package.json` | both added a subpath export next to `./Stdlib` | kept `./ChildProcess` and `./HostInput` |
| `packages/compiler/src/Hir.ts` | both extended the OS operation union | kept all six identities |
| `packages/compiler/src/Intrinsic.ts` | both inserted `osBuiltin` entries after `standardInputRead`, and both changed the `Os` consumer mapping | kept all six builtins; folded main's `processExecute`/`processCapture` cases into this branch's `osConsumer` helper (the helper was introduced here to keep the nested ternary readable), so the mapping is one function with all cases |
| `packages/compiler/src/OsRuntime.ts` | both added runtime symbols, C helper blocks, and implementation entries at the same anchors | kept all six symbols, both helper blocks, all six implementations; main's added `<poll.h>`/`<sys/wait.h>` includes auto-merged |
| `packages/compiler/src/BootstrapEvaluation.ts` | both added an import, a `BlockedReason` member, state fields, `Options` fields, state construction, and an intrinsic-dispatch branch | kept both throughout; `osProcessExecute`/`osProcessCapture` dispatch precedes the `osHost*` branch, and `processCaptures` initialization sits alongside the `hostInput` spread |
| `packages/compiler/stdlib/README.md` | both appended a service description | kept both |
| `packages/compiler/test/IntrinsicCatalog.test.ts` | both appended accepted-source wrappers to the same fixture string | kept all six wrapper functions |
| `packages/language/docs/stdlib.md` | generated | **regenerated** with `documentation:generate`; now lists 34 modules including `silk/host_input` and `silk/os_host_input` alongside `silk/os_child_process` |
| `packages/compiler/test/fixtures/intrinsic-inventory.json` | generated snapshot of `Intrinsic.inventory()` | **regenerated** from the built `Intrinsic` module and formatted with Biome; diff versus main is purely the four `osHost*` entries |

`Stdlib.generated.ts` and `stdlib/manifest.json` auto-merged and were confirmed unchanged by `stdlib:generate` during the build.

No design decision was needed: the two services never disagree, they only both extend the same lists.

### Verification

- `pnpm install --frozen-lockfile && pnpm -r build` — clean.
- `pnpm --filter @silk-effect/compiler test` — 152 files, 1199 tests passed (includes `stdlib:check` and `documentation:check`). `ChildProcess.test.ts`, `HostInput.test.ts`, `IntrinsicCatalog.test.ts`, `EffectEntry.test.ts`, `NativeToolchain.test.ts` also re-run together: 47 tests passed.
- `npx biome check .` — 545 files, clean.
- `pnpm typecheck` — 18/18 tasks clean (this is where a missed `BlockedReason` case would surface).
- Other packages: `llvm` 59, `lsp` 78, `documentation` 4, `docs` 89 — all passed.
- Known-ignorable local failures, unchanged and unrelated: `compiler-cli` FormatWorkflow/FormatCommand (3 tests, chmod as root) and `packages/wasm` `Extended.test.ts` "threads address types". The `apps/docs` terminal-session flake (#102) did not reproduce this run.

---

Issue #100, opened 2026-08-12
backend: a call through a function-typed parameter does not lower (InvalidCallShape)


julia-script:
## Current behavior (measured on `main` at `a8c3926`)

A plain `fn` may declare a function-typed parameter and call it. Semantic analysis accepts it with **zero diagnostics**. MIR verification then rejects it:

```silk
fn double(value: i32) -> i32 { return value * 2 }

fn apply(transform: once fn(i32) -> i32, value: i32) -> i32 {
  return transform(value)
}

pub fn main() -> i32 { return apply(double, 21) }
```

```
diagnostics: []
Blocked: InvalidMir — InvalidCallShape
  call …apply does not match its logical contract
  (0:fn(i32) -> i32->i32, 1:i32->i32; result=i32->i32)
```

Note the parameter's type as the violation renders it: **`fn(i32) -> i32->i32`**. The function type appears to be spliced into the contract rather than carried as a single parameter type, which is a concrete lead on the cause.

## Why this matters

Accepting a program through the entire front end and failing in the verifier is the worst failure shape available: no diagnostic names the construct, no span points at it, and the user sees an internal-sounding `InvalidMir` instead of "this is not supported yet". Either it should lower, or it should be refused with a real diagnostic at the declaration or the call.

It also blocks real work. **#35** (`Option`/`Result` combinators) is written as a Silk-source-only task and is unsatisfiable without this; its recorded blocker — "function-typed parameters do not exist anywhere in the compiler" — is false, and this is the actual prerequisite.

The effect combinators are not a counterexample: `effects.silk` declares `once fn` arms on `map`/`flatMap`/`catchAll`, but those are `effect fn` and their arms are invoked by the effect interpreter, never through this path.

**Likely neighbouring gap:** #25 was skipped in run-s1k9 after measuring the same shape of failure — `Vector<…>` has no layout, passes semantic analysis, dies at MIR verification. Whoever takes this should check whether the two share a root cause before treating them separately.

## Requirements

1. A call through a function-typed parameter in a non-effect `fn` must lower and execute, or be rejected by a diagnostic naming the construct.
2. If it lowers, the evaluator, LLVM, and Wasm must agree on the result.
3. If it is instead refused, the refusal must be a diagnostic with a span, not an `InvalidMir` violation.

## Decide before implementing

[...]

Issue #79, opened 2026-08-12
effects: Effect.catch is a whole-row alias for catchAll, not the member-selective operation the spec requires


julia-script:
## Current behavior

`openspec/specs/bootstrap-flow-functions/spec.md:47-49` requires:

> `Effect.catch<E>(effect, handler)` and `effect |> Effect.catch<E>(handler)` SHALL handle only an `E` present in the protected row, remove that member when coverage is complete, union the handler's failures, bypass the handler on success, and propagate nonmatching members unchanged.

What ships in `packages/compiler/stdlib/silk/effects.silk:119-127` is an unconditional alias for `catchAll`:

```silk
/// Recovers every typed failure with another Effect.
pub effect fn catch<A, !E, !F, ?R, ?S>(
  self: once Effect<A ! E ? R>,
  onFailure: once fn(Row<!E>) -> Effect<A ! F ? S>
) -> A ! F ? R | S {
  return run catchAll(move self, move onFailure)
}
```

- The `E` in `catch<A, !E, !F, ?R, ?S>` is the **protected row**, not a selected member — `Effect.catch<E>(handler)` cannot be written today.
- The whole row `!E` is discarded and replaced by `!F`. There is no residual row; the handler receives the entire `Row<!E>`.
- Its doc comment at `effects.silk:118` is **identical** to `catchAll`'s at `:112` — "Recovers every typed failure."

`openspec/specs/bootstrap-semantic-facts/spec.md:864` also expects facts recording "the protected row, selected member, handler row, empty residual row" — a fact shape nothing currently produces.

## Evidence there is no selective path anywhere

- Every call site in the repo passes two positional arguments with no type argument: `test/Elaboration.test.ts:137`, `test/BootstrapEvaluation.test.ts:318,386,477`, `test/Ownership.test.ts:608,627`, `test/FileSystemAcceptance.test.ts:339-341`.
- `Diagnostic.invalidEffectHandler` (`SEM0067`), whose message is literally `Invalid Effect.catch handler:`, is referenced only at `Elaboration.ts:4814,4818` — inside `analyzeEffectResult` for `Intrinsic.effectResult` arity checking, never for member coverage.

## Why this matters

Selective catch is the difference between recovering one known failure and erasing a whole typed failure row. Today, a user who wants to handle `NotFound` and let everything else propagate cannot express it: `catch` swallows the entire row, and the type system reports the result as having no remaining failures. That is a soundness-shaped gap in the failure-tracking story, not a convenience gap.

[...]

Issue #35, opened 2026-08-12
stdlib: Add Option and Result combinators


julia-script:
## Context

`Option` and `Result` are pure data today. `option.silk` declares 3 structs and
2 functions: `none` and `some`. `result.silk` declares 3 structs and 2 functions: `succeed`
and `failResult`. Each file is 22 lines. Neither module has a combinator, so every use site
must write a `match` statement.

## Current behavior

A caller that wants to change the value inside an `Option` must open the union:

```silk
let doubled = match move maybe {
  Some { value } => Option.some(value + value)
  None nothing => Option.none()
}
```

A caller that wants a default value must write the same shape again. This code repeats at
every use site. The compiler itself hits this in the standard library, because
`String.fromUtf8` returns `Result` (`string.silk:52`) and
`String.nextScalar` returns `Option` (`string.silk:189`).

The named structs are transparent. The compiler represents `Option` as the structural
union `Some | None`, so a combinator is ordinary Silk source over that union.

## Requirements

1. `option.silk` must add `map`, `flatMap`, `unwrapOr`, `isSome`, and `isNone`.
2. `result.silk` must add `map`, `mapError`, `flatMap`, `unwrapOr`, `isSuccess`, and `isFailure`.
3. `map` must take the value by move and must return a new `Option` or `Result`.
4. `unwrapOr` must consume the input and must return an owned value.
5. `unwrapOr` must take the fallback value by move and must consume it only for an absent value.
6. `isSome`, `isNone`, `isSuccess`, and `isFailure` must take a shared borrow and must return `bool`.
7. Each function must work with a move-only value type, so no function may require `Copy`.
8. Each function must drop exactly one of the two arms, so no value leaks and no value drops twice.
9. Each function must be ordinary Silk source with no new intrinsic.
10. No function may allocate, so no function may declare an effect row.

## Example

```silk
/// Applies a function to a present value and keeps an absent value absent.
pub fn map(self: Option, transform: fn(T) -> U) -> Option {
  return match move self {
    Some { value } => some(transform(move value))
    None nothing => none()
  }
}

/// Returns the present value, or the fallback value when the option is absent.
pub fn unwrapOr(self: Option, fallback: T) -> T {
  return match move self {
    Some { value } => move value
    None nothing => move fallback
  }
}
```

## Out of scope

- A combinator that declares an effect row. These functions stay pure.
- A conversion between `Option` and `Result`.
- An iteration protocol over `Option`.
[...]

Issue #25, opened 2026-08-12
effects: add sequential Effect.zip and Effect.all


julia-script:
## Context

`stdlib/silk/effects.silk` has no combinator that runs two or more Effects and collects
every result. Callers write nested `flatMap` calls by hand. Sequential `zip` and `all`
need no concurrency, so they are ordinary Silk additions. The spec requires combinators to
be "canonical ordinary Silk declarations" (`bootstrap-flow-functions:246-248`), which is
what these are.

## Current behavior

A caller nests `flatMap` to combine two Effects:

```silk
let pair = Effect.flatMap(move first, combineWithSecond)
```

The `race` and `timeout` combinators stay absent because they need concurrency
(`bootstrap-flow-functions:277-278`). `zip` and `all` do not.

## Requirements

1. `Effect.zip` must run the first Effect, then the second Effect, in that order.
2. `Effect.zip` must return a struct that holds both success values.
3. `Effect.zip` must stop at the first typed failure and must not run the second Effect.
4. `Effect.all` must run every Effect of a `Vector` in order and must collect the success
   values into a `Vector`.
5. `Effect.all` must stop at the first typed failure.
6. Both combinators must union the failure rows and the requirement rows of every input.
7. Both combinators must be ordinary Silk declarations with no intrinsic.

## Example

```silk
pub struct Pair<A, B> { first: A second: B }

/// Runs two Effects in order and holds both results.
pub effect fn zip<A, B, !E, !F, ?R, ?S>(
  self: once Effect<A ! E ? R>,
  other: once Effect<B ! F ? S>
) -> Pair<A, B> ! E | F ? R | S {
  let first = run self
  let second = run other
  return Pair<A, B> { first: move first, second: move second }
}
```

## Out of scope

- Parallel execution. Both combinators run in order.
- `race` and `timeout`. These need concurrency and stay By design absent.
- Error accumulation. The first typed failure stops the sequence.

## Dependencies

- `Effect.all` needs a `Vector` of Effects. Confirm that `Vector<Effect<...>>` is
  expressible before work starts.

## Acceptance criteria

- [ ] A test asserts `zip` runs the two Effects in declaration order.
- [ ] A test asserts `zip` does not run the second Effect after the first one fails.
- [ ] A test asserts `all` collects every success value in order.
- [ ] A test asserts both combinators compile with no intrinsic.

julia-script:
[...]

PR #119, opened 2026-08-13
effects: analyze member-selective Effect.catch at the effect seam

julia-script:
Advances #79. **Does not close it** — the analysis half of member-selective `catch` is complete and tested; no engine lowers the residual dispatch yet, so the selector form is not executable. Writing it now says so, at the call site, under a stable code. Details and the remaining work are below and on the issue.

Implements the decision recorded on #79: Design 2 (a third compiler primitive at the effect seam) with spelling (i) (`Effect.catch<E>(handler)`).

## What this adds

`Effect.catch<E>(protected, handler)` is recognized at the effect seam. The seam validates the selector against the protected row, checks the handler against the selected member, computes the residual — the protected row minus that member — and types the result as the residual unioned with the handler&#39;s failures. The four rows `bootstrap-semantic-facts/spec.md:864` describes (protected, selected, handler, residual) are recorded on the `EffectCatch` fact.

The form **without** a selector is untouched: it stays the whole-row alias in `effects.silk`, so all 196 existing `Effect.catch` call sites keep their meaning and their lowering. Only the new syntax routes to the new code.

## The non-executable form now reports SEM0097

Requested in review, and the reason this PR is no longer a fourth *passes-analysis-dies-later* instance.

Before this, `Effect.catch<A>(risky(flag), recoverA)` analyzed with **zero diagnostics** and then failed in the backend as:

```
InvalidMir → Violation InvalidEffectOperation
  in silk/effects.result$effect$2$907, region 0
  "effect result runner, channel data, tags, or calling shapes disagree"
```

A generated stdlib function the user never wrote, and no span on their source at all.

`SEM0097` is new: *a construct the front end analyzes fully but no engine can lower yet.* It is parameterized by construct name — `<construct> is analysis-only: it type-checks, but no engine lowers it yet, so a program that uses it cannot be built` — so the next analysis-only operation reuses the code rather than inventing another. It follows the shape of the nearest existing precedent, `SEM0093 intrinsicTargetUnavailable`.

### Where it is raised, and why

**In semantic analysis** (`Elaboration.ts`, at the end of `analyzeEffectCatch`), not at lowering.

The review allowed lowering as a fallback if a semantic diagnostic would wrongly reject the analysis-only tests. It does not, because the diagnostic is pushed **after** the seam has finished typing the expression and **without** touching the `valid` flag. Nothing about the analysis is suppressed: the result type, the residual row, and all four semantic facts are produced exactly as before. The `records the protected, selected, handler, and residual rows as facts` test reads the `EffectCatch` HIR node and passes unchanged.

That buys the thing lowering could not: the error appears in the editor while the line is being written, which is precisely what was missing in #25, #35 and #100. Emission still stops — `Pipeline.prepare` and `Analysis.codegen` both reject a snapshot carrying errors before instance discovery — so the build now halts on a source diagnostic instead of reaching MIR verification.

The span starts at the `Effect` token rather than using the call node&#39;s span directly, because the node span opens on the call&#39;s leading trivia and an editor would underline the preceding whitespace.

### Reported, not papered over: three existing tests changed

The review asked that any tension be reported rather than smoothed away. There is one, and it is small.

Three analysis tests assert exact diagnostic-code lists on programs that write the selector form, so those lists grew:

| Test | Before | After |
| --- | --- | --- |
| recovers one member and leaves the other in the result failure row | `[]` | `['SEM0097']` |
| reports the unrecovered member as still present in the result row | `['SEM0066']` | `['SEM0066', 'SEM0097']` |
| catch and catchAll produce different result types on the same input | `['SEM0066']` | `['SEM0066', 'SEM0097']` |

No assertion was loosened — all three remain exact `deepEqual`s, and the two message assertions are now exact `deepEqual`s over both messages rather than the single-element lists they were. What each test concludes about the analysis is unchanged: the residual survives, `B` still reaches the caller, `catch` still differs from `catchAll` at the same run site.

The first row is the interesting one. `[]` used to mean *the analysis accepts this program*; it now reads `['SEM0097']`, which means the same thing plus *and it will not build*. Those two statements are independent, and the diagnostic sits beside a clean analysis rather than replacing one. If a reviewer would rather the analysis tests stayed at zero diagnostics, the alternative is raising at lowering and losing the editor report — that is the trade, and this PR takes the editor.

### New tests (3)

- the code, severity, message and span for the direct form, asserting the span covers exactly `Effect.catch<A>(risky(flag), recoverA)` — not the enclosing `run`, not the statement, not a stdlib function;
- the same for the pipelined form `risky(flag) |> Effect.catch<A>(recoverA)`, spanning the operation rather than the whole pipeline;
- that `Analysis.codegen` fails with `CodegenUnavailable` carrying `SEM0097`, pinning that emission now stops on the source diagnostic instead of on the MIR violation quoted above.

## Scope change from the decision, reported before implementing

The decision named two prerequisites — partial explicit type arguments, and a failure-row argument form at call sites — sequenced as their own change first. **Neither is needed under Design 2**, and this PR contains neither. Both were consequences of the Design 1 premise that `catch` stays an ordinary generic function; Design 2 removes that premise. The seam runs before `analyzeCallTypeArguments`, so it never reaches the arity gate and reads its own type arguments; and the selector is an ordinary nominal, not a row argument. Verified by running it — full evidence in the [scope-finding comment](https://github.com/julia-script/silk-effect/issues/79#issuecomment-5275212699).

Partial explicit type arguments remain a real gap, orthogonal to this issue, and are now filed as #140 with a standalone reproduction.

## Accepted architectural cost

`reference.md` §5.4 is amended from &#34;exactly two primitives&#34; to three, with the reason stated: the failure row cannot express what the requirement row already can — `Requirement.capability` is `Nominal | Parameter`, `Type.Effect.failures` is `ReadonlyArray<Nominal>` — so there is no value type for a partially-open row and no residual binder for `match`, and a Silk `catch` is unwritable rather than merely awkward. Recorded as a deliberate exception, with row-polymorphic values named as the path back to two primitives. The rule is unchanged for every other operation at this seam.

One asymmetry worth flagging in review: `Effect.catch` is the first seam operation users spell **directly**. `Intrinsic.ts:1004` exposes only `string` and `Intrinsic` as actors, so `intrinsicOperationTarget` only ever matches `Intrinsic.`-qualified call heads; `result` and `bindRequirement` could stay behind Silk wrappers because their signatures are expressible. `catch`&#39;s is not, so recognition is an explicit predicate on the qualified spelling under the `Effect` namespace.

## Criteria

- [x] A test recovers one member of a two-member row and asserts the other still appears in the result failure row
- [x] A test asserts `catch` and `catchAll` produce different result types on the same two-member input
- [x] `catch` and `catchAll` no longer share an identical doc comment
- [x] Reference documentation matches shipped behavior, including what does not yet execute
- [x] Semantic facts record protected row, selected member, handler row, residual row
- [x] Regression test pinning today&#39;s failure mode (SEM0012)
- [x] Writing the non-executable form reports a real diagnostic with a stable code and a call-site span
- [ ] **The unhandled member propagates at runtime on all three engines** — not done; see below

## Not done: residual dispatch lowering

`Lower.ts` emits no body for `EffectCatch`. Analysis, typing, and facts are complete, so tooling reports the right types; the build is refused with SEM0097 rather than failing obscurely. The doc comment, §5.4 and the diagnostic all say this explicitly.

The remaining work, with the integration points found while mapping it: `EffectCatch` needs a recipe case in `Lower.ts` alongside `EffectResult` (`:1571`, handled under `Run` at `:1578`), emitting the protected reification, a tag test against the selected member&#39;s union ordinal, the handler call on match, and a narrowing re-raise of the residual otherwise. `RunEffectValue` already carries `tagMappings` mapping a callee&#39;s failure tags onto the caller&#39;s propagation outcome — that is the mechanism the residual re-raise wants, but selective catch needs it applied to a proper subset while the selected member branches away, which is new semantics for `Evaluator`, `Wasm`, and `LLVM`. Also note the sealed inventory (`Intrinsic.ts:1052`) requires every registered operation to have a non-empty target set, so the operation should be registered there only once it has a real execution surface — it deliberately is not registered in this PR. When it is, SEM0097 comes out of `analyzeEffectCatch` in the same change.

## Merge and tests

`origin/main` merged in (through #131) with no conflicts. `packages/language/docs/stdlib.md` was regenerated rather than hand-merged, as were `diagnostics.md` (SEM0097 row) and `Stdlib.generated.ts`.

Full suite after the merge, per package:

| Package | Result |
| --- | --- |
| compiler | 167 files, **1340 passed, 0 failed** (includes the 11 in `SelectiveCatch.test.ts`) |
| documentation / lsp / language / llvm / platform-webcontainer / docs | all passed |
| compiler-cli | 3 failed — FormatCommand/FormatWorkflow, chmod-vs-root, known-local |
| wasm | 1 failed — `threads address types through memory instructions`, known-local |

Exactly the known-ignorable set and nothing else. `biome check` clean, `pnpm typecheck` clean across 18 tasks, and the generated-documentation drift check passes (`documentation:check` runs as part of `pnpm --filter @silk-effect/compiler test`).

https://claude.ai/code/session_019U2QLhw6Y4DD8NS5GatMyx


julia-script:
## Review: one change required before merge — a real diagnostic for the non-executable form

Decided by Julia, 2026-08-13. The analysis work is accepted; the scope finding is accepted; **the failure mode is not**.

As it stands, `Effect.catch<E>(handler)` type-checks and then fails at build because `Lower.ts` emits no body for `EffectCatch`. That is the *passes-analysis-dies-later* shape this codebase has now paid for three separate times — **#25** (`Vector<Effect<…>>` has no layout), **#35** (blocked for weeks on a false premise, real cause a lowering gap), and **#100** (`InvalidCallShape` after zero diagnostics). Each cost a worker a full investigation to discover the front end had accepted something the back end could never build.

Shipping a fourth instance **deliberately** is the one version of this we should not do.

**Required change:** writing the selector form must produce a real diagnostic — a stable code, a message naming the construct, and a span pointing at the call — saying selective `catch` is not yet executable. Not an `InvalidMir` violation, not a verifier crash, not a silent build failure. A user who writes it should be told, at their call site, that the feature is analysis-only today.

With that in place this merges as documented-partial, and the lowering follow-up gets filed and dispatched immediately.

**Accepted as reported, no action needed:**

- **The scope finding.** The decision named two prerequisites — partial explicit type arguments and a failure-row argument form — and you established that neither is needed under Design 2, because the seam runs at `Elaboration.ts:5688-5700`, ahead of `analyzeCallTypeArguments` at `:5705`, and the selector is an ordinary nominal rather than a row argument. That correction is on the orchestrator, not on you: the sequencing instruction was reasoned from Design 1's premise and you checked it by running it instead of taking it on faith. Please do file partial explicit type arguments separately, since it remains a real gap.
- **The §5.4 amendment** to three primitives, with the failure/requirement asymmetry stated as the reason and row-polymorphic values named as the path back.
- **The asymmetry flag** — that `Effect.catch` is the first seam operation users spell directly, so recognition is a predicate on the qualified spelling rather than the `Intrinsic.`-qualified path `intrinsicOperationTarget` matches.
- **Leaving the operation out of the sealed inventory** until it has an execution surface.

---

julia-script:
## Required change done: `SEM0097`, raised in semantic analysis

Pushed as f521029, on top of a clean merge of current `origin/main` (through #131). PR body updated.

**First, the failure mode you described, confirmed by running it.** Before this commit, a program containing `Effect.catch<A>(risky(flag), recoverA)` analyzed with zero diagnostics and then produced:

```
InvalidMir → Violation InvalidEffectOperation
  in silk/effects.result$effect$2$907, region 0
  "effect result runner, channel data, tags, or calling shapes disagree"
```

A generated stdlib function, and no span on anything the user wrote. Exactly the fourth instance.

### The diagnostic

`SEM0097` — *a construct the front end analyzes fully but no engine can lower yet*:

> `<construct> is analysis-only: it type-checks, but no engine lowers it yet, so a program that uses it cannot be built`

Severity `error`, phase `semantic`. Parameterized by construct name rather than hardcoded to this operation, so the next analysis-only seam reuses the code instead of minting another; it follows `SEM0093 intrinsicTargetUnavailable`, which is the nearest existing precedent for "reachable but no execution surface". Added to `diagnostics.md` via `documentation:generate`, and named in both the `catch` doc comment and §5.4.

### Where it is raised

**Semantic analysis** — `Elaboration.ts`, at the end of `analyzeEffectCatch` — not lowering.

You allowed lowering as a fallback if a semantic diagnostic would wrongly reject the analysis-only tests. It does not, because the diagnostic is pushed *after* the seam has finished typing the expression and *without* touching the `valid` flag. Nothing is suppressed: the result type, the residual row, and all four semantic facts are produced exactly as before — the facts test reads the `EffectCatch` HIR node and passes unchanged. Analysis still concludes everything it concluded; the diagnostic is a second, independent statement sitting beside it.

That buys what lowering could not: the error shows up in the editor while the line is being written, which is the thing that was missing in #25, #35 and #100. The build still stops — `Pipeline.prepare` and `Analysis.codegen` both reject a snapshot carrying errors before instance discovery — so emission now halts on a source diagnostic instead of reaching MIR verification.

The span starts at the `Effect` token rather than reusing the call node's span, because the node span opens on leading trivia and an editor would underline the whitespace before the call.

### The tension, reported rather than papered over

Three existing analysis tests assert exact code lists on programs that write the selector form, so those lists grew:

| Test | Before | After |
| --- | --- | --- |
| recovers one member and leaves the other in the result failure row | `[]` | `['SEM0097']` |
| reports the unrecovered member as still present in the result row | `['SEM0066']` | `['SEM0066', 'SEM0097']` |
| catch and catchAll produce different result types on the same input | `['SEM0066']` | `['SEM0066', 'SEM0097']` |

Nothing was loosened to accommodate this — all three stay exact `deepEqual`s, and the two message assertions are now exact `deepEqual`s over both messages where they were single-element lists before. What each test concludes is unchanged: the residual survives, `B` still reaches the caller, `catch` still differs from `catchAll` at the same run site.

The first row is the one worth your eye. `[]` used to mean *analysis accepts this program*; `['SEM0097']` means that plus *and it will not build*. I think that is the honest state and the reason to prefer semantic analysis. If you would rather the analysis tests stayed at zero diagnostics, the alternative is raising at lowering and giving up the editor report — that is the trade, and I took the editor.

### New tests (3)

- code, severity, message and span for the direct form, asserting the span covers exactly `Effect.catch<A>(risky(flag), recoverA)` — not the enclosing `run`, not the statement, not a stdlib function;
- the same for the pipelined form, spanning the operation rather than the whole pipeline;
- that `Analysis.codegen` fails with `CodegenUnavailable` carrying `SEM0097`, pinning that emission stops on the source diagnostic rather than on the MIR violation quoted at the top.

`SelectiveCatch.test.ts` is 11 tests, all passing.

### Tests after the merge

compiler **1340 passed / 0 failed** across 167 files; documentation, lsp, language, llvm, platform-webcontainer and docs all green. The only failures anywhere are the known-local set — 3 compiler-cli FormatCommand/FormatWorkflow chmod-vs-root, and 1 wasm `threads address types through memory instructions`. `biome check` clean, `pnpm typecheck` clean, generated-doc drift check passes.

`stdlib.md` was regenerated, never hand-merged.

### Also filed

Partial explicit type arguments is now #140, with a standalone reproduction: `pair<i32>(1, true)` on `fn pair<A, B>` reports `SEM0051 pair expects 2 type arguments, received 1`, while both `pair(1, true)` and `pair<i32, bool>(1, true)` analyze clean — so the gate at `Elaboration.ts:3438` is genuinely all-or-nothing and `B` is inferable in exactly the way the fully-inferred case already proves.

---

Issue #129, opened 2026-08-13
type-system: a user-defined type cannot conform to an interface


julia-script:
## Current behavior

`DeclarationIndex.ts:3414-3419` accepts an interface witness **only** when the mapping target is a two-segment `Intrinsic.*` path. Every other target fails with `is incompatible with`.

The consequence: **no user-defined type can conform to any interface.** Every type that can conform is a built-in scalar.

Found by the #36 worker (PR #126) while trying to write that ticket's own acceptance criteria.

## Why it matters more than it looks

An interface whose only possible implementors are the twelve built-in scalars is not really an interface. Concretely, today:

- **`Order` can only order scalars.** Since every scalar is `Copy` and its equality is identity, equal elements are indistinguishable — so **sort stability is unobservable**, and a move-only element type cannot exist at all.
- **`HashKey` (#34) could only ever key a `HashMap` by a scalar**, which defeats most of the point of having one.

**The dangerous part is what this does to test suites.** PR #126 could not write two of its acceptance criteria — "equal elements keep their input order" and "a sort over a move-only element type with no leak" — because no conforming element type can witness either. Those requirements *are* met by the implementation (a generic body is checked with `T` treated as non-Copy, so the compiler already rejects reading an element out of a borrowed place), but nothing proves it. That is a gap that reads as coverage until someone checks, which is the same shape as PR #84's decorative move-only assertions that hid the #86 Wasm bug.

## This is a separate axis from #118

- **#118** is *call surface*: an operation whose name no operator spells cannot be called from a generic body.
- **This issue** is *witness admissibility*: who is allowed to conform in the first place.

`lessThan` is operator-spelled and reachable — PR #126's tests demonstrate that — so #126 was dispatchable while #34 is not. But both interfaces are capped by this issue regardless of #118.

## Requirements

1. A user-defined nominal type may declare a conformance to an interface, mapping each operation to one of its own functions.
2. Specialization checks that witness exactly as it checks an `Intrinsic.*` one — complete coverage, with a diagnostic naming any unmapped operation (the coverage check #103 added must apply unchanged).
3. Existing `Intrinsic.*` witnesses keep working, and `Integer<T>` is unaffected.
4. A generic bounded by an interface may be specialized at a user type, and the call reaches that type's mapped function.
5. The spec records what may serve as a witness target, through the openspec process.

## Out of scope

- #118's call-surface question. Assume operator-spelled operations only; the two changes compose.
[...]

Issue #103, opened 2026-08-13
type-system: accept an interface with multiple operations as a generic bound


julia-script:
## Context

Split out as the prerequisite that #34 and #36 both wait on, per the decision on #34 (Julia, 2026-08-13).

The generic bound form today is a single bare spelling: `DeclarationIndex.ts:51` declares `readonly bound?: string`. The only interface the stdlib ships is `Integer<T>` (`numeric.silk:5`), which declares **one** operation. Nothing has ever used an interface with two operations as a bound.

Two consumers are already waiting:
- **#34** needs `HashKey` — an equivalence operation and a hash operation on one bound.
- **#36** needs `Order` — and its own text notes it "needs the same bound form that `HashKey` needs."

## Requirements

1. An interface may declare more than one operation and be named as a generic bound.
2. A generic body may call every operation of the bound on a bound-typed value.
3. Specialization checks the witness completely: a type argument missing any operation of the bound is rejected with a diagnostic naming the missing operation.
4. This is type-system and resolution work only — no runtime representation change; the three engines are unaffected except through what specialization admits.
5. The spec records the bound form through the openspec process.

## Out of scope

- `HashKey` and `Order` themselves — they stay in #34/#36.
- Multiple bounds on one parameter (`T: A + B`), unless it falls out for free; if it does, say so.
- Any dyn/vtable story. Bounds are monomorphization-time only.

## Acceptance criteria

- [ ] A test declares a two-operation interface, bounds a generic by it, and calls both operations.
- [ ] A test shows a type argument missing one operation is rejected, naming the operation.
- [ ] `Integer<T>` keeps working unchanged as the single-operation case.
- [ ] An openspec change records the bound form.

## Dependencies

None. Needed by #34 and #36.

---

Filed from the #34 decision by orchestrator run `run-r7x1`.

Issue #36, opened 2026-08-12
stdlib: Add a deterministic sort over Vector


julia-script:
## Context

`Vector` has no sort. The standard library therefore has no ordered operation of any
kind. A user cannot binary-search a `Vector` either, because a binary search needs a
sorted sequence first. The TypeScript compiler sorts in several places, such as the row
members in `Type.ts` and the symbol list in `Backend.ts`. A self-hosted compiler needs the
same ordering to make its output deterministic.

## Current behavior

A user who needs an ordered sequence must write the sort again in the user program. The
user must also write a comparison for every element type, because Silk has no comparison
interface. The only interface in the standard library is `Integer` in `numeric.silk`,
and it declares one operation: `add`.

A linear scan is the only lookup today, so a keyed program pays O(n) for each read.

## Requirements

1. A new `Ordering` type must record the three results: less, equal, and greater.
2. A new `Order` interface must declare one three-way comparison operation.
3. The standard library must ship an `Order` witness for each integer type.
4. `Vector.sort` must order the elements in place over `&mut Vector`.
5. The sort must be stable, so two equal elements must keep their input order.
6. The sort must be deterministic: the same input must always give the same output.
7. The sort must give the same output on the evaluator, on LLVM, and on Wasm.
8. The sort must accept a move-only element type and must not require `Copy`.
9. The sort must move each element exactly once per exchange, so no element leaks or drops twice.
10. `Vector.binarySearch` must return `Option` for a sorted vector.
11. `Vector.binarySearch` must use the same `Order` interface as the sort.
12. The implementation must be ordinary Silk source with no compiler phase change.

## Example

```silk
pub struct Ordering {
  code: i32
}

pub interface Order {
  fn compare(left: &T, right: &T) -> Ordering
}

/// Orders the elements in place. Equal elements keep their input order.
pub fn sort(self: &mut Vector) -> ()

/// Returns the index of a matching element in a sorted vector.
pub fn binarySearch(self: &Vector, target: &T) -> Option
```

## Out of scope

- An unstable sort and a parallel sort.
- A sort that takes a comparison function value instead of an interface witness.
- A sort over `Bytes` or over a slice that a `Vector` does not own.
- A sort for `HashMap` or `HashSet`.
- A locale-sensitive comparison for text. See #42.

## Dependencies

- The generic bounds decision from #34. `DeclarationIndex.ts:51` declares
  `readonly bound?: string`, which is one unparameterized bound. `Order` needs the same
[...]

PR #141, opened 2026-08-13
type-system: give a bound's non-operator operations a call surface

### LINKED ISSUE #118: type-system: give a bound's non-operator operations a call surface
## Context

The prerequisite #34 needs, surfaced by PR #107 (issue #103) and recommended there rather than smuggled into #34.

#103 landed multi-operation interface bounds: a bound may now declare several operations, a witness is checked for **coverage** rather than existence, and a bound may name an interface another module declares. But it left one gap, which PR #107 stated explicitly rather than working around:

> An operation whose name an operator spells (`add`, `subtract`, `lessThan`, `equals`) is callable in a generic body through that operator, and every such operation of a bound now is. **An interface operation whose name no operator spells still has no call surface**: interfaces are not actors, so `Bound.operation(value)` resolves to a public function of the module declaring `Bound` (which is why `Integer.add(40, 2)` resolves today), not to the bound's operation.

## Why it blocks #34

`HashKey` needs **`hash`**. No operator spells it. So a generic bounded by `HashKey` can reach the equivalence half through `equals` and cannot reach the hash half at all — which is the whole point of the bound. #36 (`Order`) is unaffected, because comparison *is* operator-spelled.

## The decision this needs

**How a bound's operation is spelled at a call site.** The services parallel suggests `Bound.operation(args)`. Whatever is chosen, note the constraint PR #107's `design.md` records: unlike an operator, this must consult **the witness's selected operation at specialization**, rather than reusing the width-neutral scalar lowering an operator already gets.

Do not implement before that spelling is chosen.

### Decision (Julia, 2026-08-13)

**`Bound.operation(args)`** — qualified through the bound's own name, parallel to a service operation. That spelling already parses; it resolves to the wrong thing. The work is making resolution prefer the bound when the receiver names a bound of the declaration being elaborated.

## Requirements

1. An operation declared by a bound is callable in a generic body regardless of whether an operator spells its name.
2. The call resolves to the witness's implementation for the specialized type argument, not to a same-named public function of the interface's declaring module.
3. A name collision between a bound operation and a module-level function of the same name resolves unambiguously, with a diagnostic if it cannot.
4. Existing operator-spelled operations keep working exactly as #103 left them.
5. The spec records the call form through the openspec process.

## Acceptance criteria

- [x] The spelling is chosen and recorded here before implementation.
- [x] A test declares an interface with a non-operator-spelled operation, bounds a generic by it, calls that operation, and evaluates end to end.
- [x] A test shows the call reaches two different witnesses for two different type arguments.
- [x] A test covers requirement 3's collision case.
- [x] `Integer` and the #103 tests still pass unchanged.

## Dependencies

#103 (landed, PR #107). Blocks #34.

---

Filed by orchestrator run `run-r7x1`, from PR #107's recommendation.

julia-script:
Closes #118

## What

An operation whose name an operator spells was callable through that operator; one whose name no operator spells was reachable from nowhere. `Bound.operation(value)` parsed, but resolved to a public function of the module declaring the interface — which is why `Integer.add(40, 2)` resolves today.

The spelling is settled: **`Bound.operation(args)`**, qualified through the bound's own name, parallel to a service operation ([decision recorded on the issue](https://github.com/julia-script/silk-effect/issues/118#issuecomment-5283424914)).

```silk
pub interface Mixer<T> {
  fn mix(left: T, right: T) -> T
}

impl Mixer<i32> for i32 { mix: Intrinsic.i32WrappingAdd }
impl Mixer<u8> for u8 { mix: Intrinsic.u8SaturatingAdd }

pub fn blend<T: Mixer>(left: T, right: T) -> T { return Mixer.mix(left, right) }
```

One body, two specializations, two instructions: `blend<i32>` wraps, `blend<u8>` saturates.

## How

**Resolution.** A two-segment receiver naming an interface is resolved against the bounds of the declaration being elaborated. When one of its type parameters is bound by that interface *and* the bound's recorded contract declares the member, the bound takes the spelling. The preference is deliberately narrow: a member no bound declares, and a body with no such bound, resolve exactly where they did before. The match is on the bound's canonical capability, not on the spelling it was written with, so an aliased import is the same bound. A receiver naming an interface that bounds two of one declaration's parameters answers to no single parameter and is reported — `SEM0097`, naming the parameters it is ambiguous across.

**Contract.** The interface's own contract with its type parameter replaced by the bounded one, checked once over the canonical parameter, before any concrete argument exists. `mix(left: T, right: T) -> T` over the bound's `T`; `ranksBelow(left: T, right: T) -> bool` keeps its `bool`. No type arguments of its own.

**Specialization.** This is the constraint PR #107's `design.md` recorded. An operator carries its compiler-known operation from elaboration because the specialized operand type alone selects the instruction — `Add` over `u8` and over `i32` are one width-neutral lowering. A bound operation has no such property: two providers may map one operation to two unrelated instructions. So the new `BoundOperationCall` HIR node records the *question* — which operation, of which interface, over which bounded parameter — and carries no operation code and no intrinsic identity. Lowering, the phase that holds the substitution, reads the conformance the specialization admitted, takes the compiler-known operation it maps, and continues as the ordinary builtin call it now is. The reachable-intrinsic report reads the same witness, per instance, so target availability sees each specialization's own operation.

Recording the question rather than an answer is also what keeps the node stable under #129: when witness admissibility widens, the node does not change — only what lowering finds at the end of the conformance does.

## Criteria

- [x] The spelling is chosen and recorded on the issue before implementation.
- [x] A test declares an interface with a non-operator-spelled operation, bounds a generic by it, calls that operation, and evaluates end to end.
- [x] A test shows the call reaches two different witnesses for two different type arguments — wrapping for `i32`, saturating for `u8`, neither the other's width-neutral form.
- [x] Tests cover requirement 3's collision case: the bound preferred inside a bounded body, `Integer.add(40, 2)` unchanged outside one, and the two-parameter receiver reported.
- [x] `Integer` and the #103 tests pass unchanged, with `numeric.silk` untouched.
- [x] The spec records the call form through openspec (`give-bound-operations-a-call-surface`).

## Out of scope

`HashKey` (#34) and `Order` (#36) themselves; witness admissibility (#129), so the tests necessarily use scalar witnesses; a bound operation as a first-class value.

## Tests

`Tests: baseline 1 failure (LexerPressure timeout under parallel load, passes alone), after 0 failures, +8 new tests` — 1329 → 1337 in `packages/compiler`. The five behavioural additions were verified to fail before the change; the two guard tests (`Integer.add` outside a bounded body, an operation the interface never declares) pass before and after by design.

Also run: `turbo run typecheck` clean, `pnpm test:scripts` 8/8, `lsp` 78/78, `llvm` 59/59, `documentation` 4/4, `language` 28/28. Pre-existing local failures unrelated to this change and identical at baseline: `packages/wasm` `Extended.test.ts` "threads address types", `compiler-cli` `FormatCommand`/`FormatWorkflow` (chmod vs root).

New diagnostic code `SEM0097` — `packages/language/docs/diagnostics.md` regenerated via `documentation:generate`, and the language reference documents the call form.

https://claude.ai/code/session_01NNNUW72WRYcjUbZewZya4a

---
_Generated by [Claude Code](https://claude.ai/code/session_01NNNUW72WRYcjUbZewZya4a)_

### DIFF
--- a/openspec/changes/give-bound-operations-a-call-surface/design.md
+++ b/openspec/changes/give-bound-operations-a-call-surface/design.md
@@ -1,0 +1,90 @@
+## Context
+
+`accept-multi-operation-interface-bounds` left one gap deliberately and recorded why: an interface
+is not an actor, so `Bound.operation(value)` resolves to a public function of the module declaring
+the interface rather than to the bound's operation. It also recorded the constraint any answer must
+satisfy — unlike an operator, a bound operation must consult the witness's selected operation at
+specialization rather than reuse the width-neutral scalar lowering an operator already has.
+
+The spelling is settled: `Bound.operation(args)`, qualified through the bound's own name, parallel
+to a service operation. What follows is how that spelling resolves and how it reaches the engines.
+
+## Goals / Non-Goals
+
+Goals:
+
+- Every operation a bound declares is callable in the generic body, at the contract the interface
+  declares for it, checked once over the canonical parameter.
+- The call reaches the witness's implementation for the specialized type argument.
+- A collision with a module-level function of the same name resolves by a stated rule, and a
+  receiver that answers to no single parameter is reported.
+
+Non-Goals: `HashKey` and `Order` themselves, witness admissibility (#129), a bound operation as a
+first-class value, type arguments on a bound operation, and any runtime representation change.
+
+## Decisions
+
+### The bound takes the spelling, and only the names it declares
+
+Resolution of a two-segment receiver already asks what the qualifier names. When it names an
+interface, the new question comes first: does the declaration being elaborated bound one of its
+type parameters by that interface, and does that bound's recorded contract declare this member? Only
+both answers together take the spelling. A member the bound does not declare, and a body with no
+such bound, fall through to exactly the resolution they had before — which is what keeps
+`Integer.add(40, 2)` resolving to `silk/numeric`'s public `add` outside a bounded body, and keeps
+every other member of an interface's module reachable inside one.
+
+Inside a bounded body the bound wins. `Integer.add(left, right)` in a body bounded by `Integer`
+names the bound's operation, not the module function it shadows. This is the resolution the whole
+change exists to produce: qualifying through the bound's name and getting the declaring module's
+function instead is the defect being fixed, so preferring the module function there would leave the
+bound's operation with no spelling at all. The module function stays reachable through its module.
+
+The match is on the bound's recorded canonical capability, not on the spelling the bound was
+written with, so an interface imported under an alias is the same bound under either name.
+
+### A receiver that names two parameters is reported, not guessed
+
+One declaration may bound two of its parameters by one interface. The receiver is the bound's name,
+so `Bound.operation(...)` then names no single parameter, and each parameter has its own witness.
+The call is reported — `SEM0097`, naming the parameters it is ambiguous across — rather than
+resolved to the first parameter. Nothing here forecloses a receiver that names the parameter
+instead; it forecloses only guessing.
+
+### The contract is the interface's own, over the bounded parameter
+
+The interface writes its contract over its own type parameter. A bound applies that interface to one
+parameter of the bounded declaration, so the operation's contract in the body is the declared one
+with the interface's parameter substituted — `mix(left: T, right: T) -> T` over the bound's `T`,
+`ranksBelow(left: T, right: T) -> bool` keeping its `bool`. That is the same contract the conformance
+check already holds every witness to, which is what lets the body be checked once, over the
+canonical parameter, before any concrete argument exists.
+
+A bound operation carries no type arguments of its own: the only type the call varies over is the
+bounded parameter, and that one is supplied by the specialization of the declaration the body
+belongs to.
+
+### The call records the question; specialization answers it
+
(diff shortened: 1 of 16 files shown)


julia-script:
CI is green — `validate` and `webcontainer-browser` both pass, plus `release:candidate`.

The first `validate` attempt failed on one test, and it was not this diff. `OccurrencePerformance` — `keeps representative multi-module occurrence storage and lookup within budget` — measured 1131ms against a 1000ms wall-clock budget. It was the only failure in 1337 tests, and re-running the same commit unchanged passed.

Measured before pushing anything: 415/406/407ms at this commit, 434/407/420ms at its parent — a wash, marginally faster. It also can't be affected in principle: the timed region calls only `Analysis.semanticOccurrenceAt`, a binary search over a prebuilt index, while the one line this PR adds to `SemanticOccurrence` runs during index construction, outside the timing — and the fixture has no interfaces or bounds, so the added branch is never taken.

Filed as #143 rather than fixed here: the budget has ~2.4x headroom over a ~410ms measurement, which a shared runner can lose, and per #102 the fix wants to be a structural guard rather than a raised number. That's a change to someone else's guard and doesn't belong in this diff.

---

Issue #130, opened 2026-08-13
backend: clang 18.1 crashes in Register Coalescer compiling removeDirectoryRecursively at -O2


julia-script:
## Current behavior

Compiling `silk/filesystem`'s `removeDirectoryRecursively` through the native backend at `-O2` crashes clang inside its own optimizer:

```
Running pass 'Register Coalescer' on function '@silk_silk_filesystem_remove…'
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/
```

**A release-profile program that calls `TemporaryDirectory.release` therefore fails to build.** Debug builds are fine, and the function is correct on the evaluator and correct natively at `-O0` — the populated-tree test removes a real nested tree on a real filesystem and asserts it is gone.

Found by the #31 worker (PR #128), which ships with its two native tests pinned to `profile: 'debug'` and a comment at each site explaining why.

## What has already been ruled out

The worker established this is a backend defect rather than something the module can express differently:

- Reproduces with **three different formulations** of the directory walk — two vectors plus an index; one vector with borrow-matches; owned entries popped and consumed.
- Reproduces with the same function written in **an ordinary user module**, outside the stdlib.

So changing the Silk source is not a workaround, and the crash does not depend on anything specific to `silk/filesystem`.

## What is not yet known

Nobody has narrowed it further. Open questions for whoever takes this:

1. **Which clang versions.** Only 18.1 is confirmed. If a newer clang is clean, pinning or documenting a minimum version may be the whole fix.
2. **Which IR construct triggers it.** A minimal reproduction would make an upstream report actionable, and might reveal an emission pattern the backend can avoid.
3. **Whether the emitted IR is itself unusual.** A Register Coalescer crash is upstream's bug regardless, but if the backend emits something needlessly hostile — very long live ranges, unusual block structure — changing the emission may sidestep it without waiting on LLVM.

## A separate observation from the same investigation, not the same bug

While bisecting, the worker found that a **larger native program traps at `-O0`** once enough owners are live across the release. That was not isolated, is not introduced by #128, and may or may not be related. Worth reproducing before assuming either way — if it is real and separate, it wants its own issue.

## Requirements

1. Determine whether this reproduces on clang versions other than 18.1.
2. Produce a minimal IR reproduction, or establish that the emitted IR is unremarkable and the bug is purely upstream.
3. Either avoid the trigger in emission, or document a supported-toolchain constraint and report upstream.
[...]

Issue #114, opened 2026-08-13
backend: Slot.take and Slot.copy read 4 bytes for a u8 element on Wasm


julia-script:
## Current behavior

`Slot.take` and `Slot.copy` on a `u8` element read **4 bytes** on the Wasm backend. `WasmBackend` emits the generic `i32.load` for slot lanes, where `RawBufferRead` correctly uses the lane mnemonic for the element width.

Silent wrong answer, not a trap.

## Reproduction

Write `7u8` at index 0 and index 3 of a `RawBuffer<u8>`, take both, and sum them:

| engine | result |
|---|---|
| evaluator | `14` |
| native LLVM | `14` |
| **Wasm** | **`117440526`** |

`117440526` is `0x07000007` — the load at index 0 picked up the byte written at index 3 in the same 4-byte window, which is exactly what a 4-byte load of a 1-byte lane produces.

Found by the #21 worker while building the bulk-memory intrinsics, and **reproduced with none of that branch's code** — this is on `main` today, independent of #113.

## Why it matters

Same class as #86 (the `Vector` element-0 zeroing fixed in PR #95): a Wasm-only silent miscompilation where the evaluator and LLVM are both correct. Any `u8`-element storage read through a slot is affected, which includes byte buffers — the most common element type in the standard library.

It also means three-engine parity tests only catch this when they assert on **values**, not on completion. Worth checking whether existing byte-level tests assert strongly enough.

## Requirements

1. Slot lane loads use the element's lane mnemonic, matching what `RawBufferRead` already does.
2. The three engines agree for every scalar element width, `u8`/`i8` and `u16`/`i16` included — the same bug shape is plausible at 2 bytes and should be checked rather than assumed absent.
3. A regression test asserts values across all three engines, verified failing on Wasm before the fix.

## Acceptance criteria

- [ ] The reproduction above gives `14` on all three engines.
- [ ] A test covers `Slot.take` and `Slot.copy` for each scalar element width.
- [ ] The regression test is verified to fail on Wasm pre-fix.

---

Filed by orchestrator run `run-r7x1`, from a defect the #21 worker found and reported rather than worked around.

PR #149, opened 2026-08-13
stdlib: add HashMap, HashSet, HashKey, and HashSeed

### LINKED ISSUE #34: stdlib: Add HashMap, HashSet, HashKey, and HashSeed

julia-script:
Closes #34.

Picks up where the draft stopped. The confirmation this PR opened with found a gap that had to close first; #155 (PR #157) closed it, and the collections are here.

## The two things asked for before resuming

**Rebased** onto `main`, which has since moved again — now onto `88bc41c`, carrying #151 and #153 as well as #157, #119, #148, #154 and #156.

**Tripwire inverted.** `HashKeyBoundForm.test.ts`'s seventh case asserted the broken behaviour on purpose, and #157 correctly left it alone rather than reaching into this branch. It now asserts the working outcome and is **kept, not deleted** — it is the exact path every `HashKey.hash` call over a user-defined key takes. The other six cases are unchanged and still pass.

No diagnostic was added, so `SEM0097`–`SEM0101` are untouched.

## The collections

`hash.silk` declares `HashSeed`, the `HashKey<T>` contract, a mixer witnesses can fold through, and `Word`. `hash_map.silk` and `hash_set.silk` are open addressing with linear probing over two buffers of one width — the entries, initialized exactly where the state byte says occupied, and the states, which are vacant, occupied, or removed. Ordinary Silk over `Allocator` and typed storage throughout.

The recorded decisions all stood: the equivalence is `equals` because `==` spells it, the seed is an operand of the hash rather than key state, and every hash is `u64`.

### Two language facts shaped the code more than any decision did

**A bound is not carried into a nested generic call.** A `K: HashKey` body cannot hand its own `K` to another `K: HashKey` function — the constraint `vector.silk` already records for its sort. Everything needing a witness has to be written in the body that needs it, so `insert`, `contains`, `indexOf`, `get` and `remove` each carry their own probe loop, and `HashSet` is the table written out again rather than a map with empty values.

Storing each entry's hash beside its key is what stops this spreading. Growth rehomes an entry under the hash it was placed with instead of hashing it again, so migration, placement and release stay ordinary unbounded code that the five bounded operations can call.

**A scalar type cannot witness `HashKey` at all.** A conformance whose provider is a scalar admits a sealed compiler operation and nothing else, and no compiler operation hashes — nor may one be added. So an integer key is a declared type. `Word` is the standard library's, and the constraint is documented on it rather than left for a caller to trip over. This is the same argument the draft gave for why scalar keys were never a viable partial path; it survives #157 unchanged.

### Where three-engine agreement is actually won

A hash is `u64`; a bucket index is `usize`, which is 64 bits natively and 32 under WebAssembly. Narrowing first and reducing second would put a key in different buckets on different engines — the map would still answer every lookup correctly while presenting its entries in two different orders, which a test that only checked lookups would never catch. The remainder is taken against the width in `u64`, and only the result narrows.

## Acceptance

All five criteria, plus the two singled out as most likely to be skipped.

**Seeded order, in every engine.** Each program folds the keys in bucket order, checks the fold against the order that seed is known to give, and answers 42 only on a match; that answer is required from the evaluator, from WebAssembly, and from native LLVM. A second seed gives a second order, so the seed is what fixes it rather than riding along unused. Two maps built with one seed on either side of an unrelated allocation agree, so the order is not an address's. The fold is compared inside the program because a native run reports its answer as an exit status and only the low byte survives — worth knowing before writing a test that returns a wide value from native.

**Move-only ownership**, on keys as well as values, because a map is generic over its key. Four programs assert the exact number of allocations acquired and that every one was released: on removal, on overwrite, on the map's own drop while non-empty, and across the growth that moves every entry once. Stating the expected acquire count rather than deriving it matters — an insert that stopped taking ownership would balance trivially and pass. Each asserts it *completed* first, because a second release traps rather than completing.

**Requirement 9** is checked twice. The MIR of a program that inserts, grows, looks up and removes names no hash operation, and the hash it computes is an ordinary `Call` to the witness's own Silk function, itself lowered from `hash.silk`. Wider: the intrinsic catalogue contains no hash, and since analysis, the HIR, the MIR, the evaluator and both backends all draw from that one catalogue, that is six statements at once. An interface of the same two operations under a name nothing could recognize answers identically, so nothing special-cases the spelling.

Also covered: a failed growth leaving the map exactly as it was, and probing across the marks a removal leaves — the classic linear-probe failure, where a removal that vacates its slot outright cuts the run carrying later keys past it.

## One thing found and not taken

An `effect fn` whose result struct carries any 64-bit field fails WebAssembly emission (`Expected i32 on the stack but found i64`). It is bisected to a reproduction naming no collection — it needs no allocation, no ownership, no bound, and works from an ordinary function — and reproduces on `main` without this branch. Reported on #34 ([comment](https://github.com/julia-script/silk-effect/issues/34#issuecomment-5286541060)) for its own ticket rather than fixed here. It cost the tests nothing: the move-only key's tag is an `i32` its witness widens, and a comment records why.

## Verification

- `pnpm --filter @silk-effect/compiler test` — **1423 passed**, 0 failed, plus native acceptance.
- `pnpm biome check .` — clean, 578 files. `pnpm typecheck` — clean, 18/18.
- `documentation:check` and `stdlib:check` clean; `stdlib.md` regenerated, never hand-edited.
- Every other workspace package green: `documentation`, `lsp`, `llvm`, `language`, `docs`.
- `pnpm check` still fails in `@silk-effect/wasm` (`Extended.test.ts > threads address types through memory instructions`, #160) and in `@silk-effect/compiler-cli` (three format tests that `chmod` a path and expect to be refused, which root is not — #159). Both are known and pre-existing; this diff touches neither package.


julia-script:
## Interim review — the shape is right; two acceptance criteria are still open

Reviewed through `3b920ad`. This is going well and I want to name what is right before what is left, because both matter.

### What is already right

- **`threeEngineValue` asserts values on all three engines** — evaluator, `codegenWasm`, and a real `Driver.compile`. Not "it compiled." That is the standard #86, #114 and #130 were all missed by.
- **Requirement 9 holds structurally, not by assertion.** The only file the branch touches under `packages/compiler/src` is the *generated* `Stdlib.generated.ts`. No compiler, MIR, evaluator or backend change exists to review. That is the strongest possible form of that guarantee.
- **`insert` returns `Option<V>`**, so an overwritten value is handed back to the caller rather than silently dropped — the honest way to discharge that ownership.
- **The determinism suite goes past the obvious case.** *"presents one order for two collections built with one seed, whatever was allocated between"* guards against allocation-address leakage into iteration order, which is exactly the way a seeded-order claim usually turns out to be false. Pairing it with *"a different order under a different seed"* means the test would fail if the seed were being ignored entirely — a same-seed test alone would not.
- **`leaves the map intact when the growth allocation fails`** covers the `OutOfMemory` path rather than assuming it.

### Outstanding — tasks 6.3 and 6.4

Both are unticked in your own `tasks.md`, so I take it these are known and in flight. Recording them so nothing depends on memory:

**6.3 — a test showing the MIR for a `HashMap` program contains no hash operation.** Requirement 9 currently holds because no compiler file changed, which is excellent evidence *today* and no evidence at all against a future change. The test is what makes it durable.

**6.4 — correct cleanup when the map owns move-only values, asserting acquires and releases.** This is the one I flagged twice as most likely to be skipped, and it is the last one standing. The repo already has the tooling — `AllocationMetricsAcceptance.test.ts` uses an `impl Drop` ledger and acquire/release counts, and `OwnedAllocation*.test.ts` and `WasmHeapReclaim.test.ts` use the same pattern. Please reuse it rather than building something new.

Your own spec text names the three cases that matter, and all three need to be in the test: **on removal, on overwrite, and on the map's own drop while non-empty**. The middle and last are the ones that get missed — a map dropped while still holding entries has to walk its slots, and `releaseFull` calling `Slot.dropValue` looks correct by reading, but "looks correct by reading" is what a Drop bug survives.

Exactly once, not at-least-once: a double release is as much a defect as a leak, and only the counts show the difference.

### Not blocking

Take the PR out of draft when 6.3 and 6.4 land and it closes #34. `main` has moved since you branched — #153 landed `SEM0099`/`SEM0100`, so if you add any diagnostic, `SEM0097`–`SEM0101` are all taken and #156's guard will reject a reused code.

---

Issue #145, opened 2026-08-13
backend: verify emitted LLVM IR and Wasm modules in CI


julia-script:
## Context

Neither backend checks its own output. Invalid IR goes straight to clang, and what happens next depends on which clang and which optimization level you happened to use — a crash, a `SIGILL`, or a silent wrong answer.

This is not hypothetical. **Three separate silent-wrong-answer defects were found in one session**, all in the backends, none by a test that was looking for them:

| Issue | Defect | How it presented |
|---|---|---|
| #86 (fixed, PR #95) | Wasm stored element 0 of a `Vector` of move-only structs as zero | wrong values, evaluator and LLVM correct |
| #114 (fixed, PR #123) | Wasm read 4 bytes for a `u8` lane; **8-byte lanes never emitted a module at all** | wrong values, and total failure for `u64`/`i64`/`f32`/`f64` elements |
| #130 (fixed, PR #144) | LLVM backend emitted **50 dominance violations** — invalid SSA | presented as a "clang bug" in Register Coalescer |

Each was found by someone chasing an unrelated symptom. Each was invisible to a test asserting on *completion* rather than on *values*.

**#130 is the clearest case for this issue.** The fix's author ran one command:

```
$ opt -passes=verify program.bc
Instruction does not dominate all uses!
  %25949 = load i32, ptr %25948, align 4
  ... 50 of these, all in @silk_..._removeDirectoryRecursively_...
```

Ubuntu's clang is built with assertions off, so it does **not** run the verifier on `-x ir` input. The invalid module went straight to codegen. Both reachable clang versions crashed, in *different passes* — 18.1.3 in Register Coalescer, 19.1.1 in correlated-propagation — which is why a minimum-version constraint would have been actively harmful: it would have hidden a real miscompile behind a toolchain pin.

## What to add

1. **LLVM: run the verifier on every emitted module.** `opt -passes=verify` on the bitcode, or the equivalent in-process. The command that found #130 is exactly the check that would have prevented it.
2. **Wasm: validate every emitted module.** `WebAssembly.validate` at minimum; a stricter validator if one is already a dependency. Note #114's 8-byte-lane case produced a module that could not be instantiated at all — validation catches that class immediately.
3. **Wire both into `pnpm check`,** so it runs on PRs *and* on pushes to `main`. #94 established that a check which only runs on PRs misses defects that appear on the merge result.

## Where it should run

[...]

PR #150, opened 2026-08-13
llvm: make IrText.render output valid, round-trippable LLVM assembly

### LINKED ISSUE #146: llvm: IrText.render output is not valid LLVM assembly and does not round-trip
## Current behavior

`IrText.render` produces text that `llvm-as` rejects. Two defects found so far:

- **Duplicate value names** — the same `%N` is emitted for more than one value.
- **`store i32 null`** — `null` is written where the type is not a pointer.

Found by the #130 worker (PR #144) while trying to run the LLVM verifier against the backend's output.

## Why it matters, and why it is P2 rather than P1

It breaks nothing at runtime today. The compile path uses **bitcode**, not the rendered text, so no shipped artifact is affected.

The problem is that **the human-readable view of the IR is not the same thing as what actually gets compiled** — which is exactly the property you want when debugging a miscompilation, and exactly when it fails you. In #130 the worker wanted to run `opt -passes=verify` on the emitted module, could not use the renderer's output because `llvm-as` rejected it, and had to route through bitcode instead.

**The golden tests cannot catch this.** They compare rendered text against stored rendered text — text to text — so a systematic renderer flaw compares equal to itself and passes indefinitely. That is the same shape as PR #84's move-only assertions that never read their value: a check that looks like coverage and asserts nothing about correctness.

## Requirements

1. `IrText.render` output is accepted by `llvm-as`.
2. Value names are unique within a function.
3. Constants are rendered with a spelling valid for their type — `null` only where the type is a pointer.
4. A test asserts the property rather than the text: render a module, feed it to `llvm-as`, require it to assemble. Ideally assemble it and compare against the bitcode the same module produces, so the two paths cannot drift.

## Acceptance criteria

- [ ] A representative module's rendered text assembles with `llvm-as` without error.
- [ ] The test covers a module exercising `Drop` hooks and union cleanup — the shapes where #130's duplicate names appeared.
- [ ] Existing golden tests still pass, or are updated with the corrected rendering and the diff explained.

## Related

- #130 / PR #144 — where this was found; the invalid-SSA bug whose diagnosis this obstructed.
- #145 — verifying emitted modules in CI. If that lands first, it should verify **bitcode**, since the text path is unreliable until this is fixed.

---

Filed by orchestrator run `run-r7x1`, from an incidental finding in PR #144.

julia-script:
Closes #146.

`IrText.render` produced text that `llvm-as` rejects, so the human-readable view of a module was not the thing being compiled.

## Defects

The two known ones from #130 / PR #144, plus two more found by rendering a spread of modules and feeding each to `llvm-as` rather than by fixing the two known spellings and stopping:

1. **Duplicate local names.** `localIdentifier`/`blockIdentifier` emitted the requested name verbatim, so two values asking for the same name collided — and so did a block label and an instruction result of the same name, since LLVM gives values and blocks a single symbol table per function. Names are now allocated once per function body: the requested spelling is kept when free and disambiguated with a `.N` suffix otherwise, in value-then-block index order so the mapping is a pure function of the snapshot. Resolved forward references are skipped, since they render as the value they resolve to and never define a name.

2. **`store i32 null`.** `Constant.nullValue`, `Constant.none`, and `Constant.zero` all encode to the *same* bitcode record (`CST_CODE_NULL`), but the renderer echoed whichever constructor the front end happened to call. The spelling now comes from the type, which is the only thing that can be correct: `null` for pointers, `none` for tokens, `0` for integers, `0.0` for half/bfloat/float/double, hexadecimal zero for `x86_fp80`/`fp128`/`ppc_fp128` (which reject decimal literals outright), `zeroinitializer` for aggregates.

3. **`!DIFile` missing `directory`.** Absent operands were dropped from the field list, but the textual parser requires both `filename` and `directory`. Both are now always printed, empty when the operand is null — which is how bitcode reads them back. This is why *every* `debug`-mode module failed to assemble, not just the ones with `Drop` hooks.

4. **`!DICompileUnit` disagreeing with its own bitcode.** The renderer omitted `splitDebugInlining`, whose textual default is `true`, while `Bitcode.encode` writes the operand as `false`. Both paths parsed; they just described different modules. Found by the round-trip comparison, not by `llvm-as`.

## The test

Requirement 4 is the real deliverable, so it is worth saying what it does and does not assert.

`packages/llvm/test/IrTextRoundTrip.test.ts` covers the shapes at the constructor level: two values and a block all asking for one name, every scalar and aggregate lane zeroed through `Constant.nullValue`, a `DIFile` with neither operand.

`packages/compiler/test/LlvmIrRoundTrip.test.ts` covers them as they actually arise, in `debug` and `release`:
- a `Drop` hook over an owned `Allocation`;
- the boxed tree from `BoxHeapIndirection`, whose `Leaf | Branch` union costs a conditional cleanup around `impl Drop for Box` — the shape #130's duplicate names appeared in.

Both render, run `llvm-as`, run `opt -passes=verify`, and then require the assembled module to equal the one `Bitcode.encode` produces, so the two paths cannot drift apart again.

**Comparing the two paths needs both sides through one door**, and the two normalizations are load-bearing:
- The textual parser resolves an *omitted* `align` against the module's data layout as it reads and bakes the result in, while bitcode leaves the operand unspecified for its consumer to resolve. The emitted modules carry no `target datalayout`, so the two resolve differently (4 vs 8 for `i64`). Each side is therefore materialized through `llvm-as` before comparison. An explicitly wrong alignment still fails the comparison; only unspecified-vs-unspecified is laundered.
- `strip` then renumbers locals, because `Bitcode.encode` writes no function-level value symbol table, so names reach a consumer only along the textual path.

I checked the tests bite by reverting each of the four fixes in turn with the others in place. Each mutation fails at least one test; the `packages/llvm` suite catches all four, and the compiler suite catches the three that reach it (the name mutation fails exactly the two union-cleanup cases). Worth noting because the compiler package resolves `@silk-effect/llvm` from `dist`, so a mutation check there is silently vacuous without a rebuild — the same shape of non-assertion this issue is about.

## Finding the toolchain, and refusing to skip

The first CI run died with `spawnSync llvm-as ENOENT`. `llvm-as`, `llvm-dis`, and `opt` ship in LLVM's tools package rather than alongside `clang`, and the Linux distributions leave them under a versioned prefix without symlinking them into `/usr/bin` — so a runner that compiles Silk programs fine has none of them by bare name.

#151 had already hit this with `opt` and solved it by looking where the distributions put it; its green CI *with the hard-fail guard active* is the evidence that the prefixes are right, since a green run there is only possible if `opt` was actually found. Rather than copy that resolution into two more files, it now lives in **`test/support/llvmToolchain.ts`** — the repository-root `test/support` directory that already holds the cross-package `raise.ts` — and takes a tool name instead of being written once per tool. All three call sites share it: both round-trip suites and #151's `ModuleVerification.test.ts`, whose inline copy is deleted here. Overrides follow #151's naming, derived from the tool: `SILK_TEST_OPT`, `SILK_TEST_LLVM_AS`, `SILK_TEST_LLVM_DIS`.

The obvious alternative — skip when the tool is missing — is the one thing that cannot happen here. A round-trip test that skips is a golden test comparing text against itself, which is the failure mode #146 exists to remove; #151's silent skip is precisely why its cross-check had never once run in CI without anyone noticing. So the helper keeps that suite's shape and now enforces it for all three: a missing tool skips locally with a printed reason, and hard-fails under `CI`, answered by installing LLVM rather than by a quieter test.

## Golden diff

Two lines, both in `packages/llvm/fixtures`, both text-only — the `.bc.hex` fixtures and their recorded sha256 are byte-identical, which is the point:

```diff
 core/representative.ll
-  %negated = sub i32 zeroinitializer, %xor
+  %negated = sub i32 0, %xor

 metadata/representative.ll
-!22 = distinct !DICompileUnit(…, emissionKind: FullDebug, globals: !21)
+!22 = distinct !DICompileUnit(…, emissionKind: FullDebug, globals: !21, splitDebugInlining: false)
```

`sub i32 zeroinitializer` does assemble, so that line was not itself a defect; it changes because the fix derives every zero spelling from the type rather than from the constructor, which removes the class instead of the one spelling that failed. The `!DICompileUnit` line is defect 4. All five `fixtures:verify` suites pass. No compiler goldens changed.

`packages/llvm/tsconfig.test.json` gains `"types": ["node"]` and the package gains `@types/node`, so the test can shell out to the toolchain — the same test-only widening `packages/compiler` already carries, with the reason in a comment.

## Verification

Rebased onto `main` at 841659c, after #151, #157, #156, #154, #148, and #119 landed. The rebase was clean and nothing in the renderer change moved.

`pnpm check` passes except for two failures that are pre-existing on `main` and untouched by this branch: `@silk-effect/wasm > test/Extended.test.ts > threads address types through memory instructions` (#160) and the two `@silk-effect/compiler-cli > test/FormatWorkflow.test.ts` permission cases (#159, which fail only because the agent runs as root and root ignores `chmod`). Turbo cancels siblings when one task fails, so `@silk-effect/llvm` and `@silk-effect/compiler` were re-run standalone: green in full, 66 and 1392 tests plus the native acceptance run. `pnpm typecheck` and `biome check` are clean across the repo.

I also exercised both branches of the new guard directly — missing tool under `CI` throws with the override variable named, missing tool locally returns a skip — since the environment this runs in has the tools and would otherwise never take either path.

Beyond the gate, while hunting: every program in `packages/compiler/test/support/corpus.ts` (66) plus the two above, across `debug`/`release` × `aarch64-apple-darwin`/`wasm32-unknown-unknown`, rendered → `llvm-as` → `opt -passes=verify`. All pass. Before this change every `debug` module failed and the `Drop` module failed in both modes.

## Left alone, deliberately

Three things this surfaced that are out of scope here and would each be a separate change:

- **The emitted modules carry no `target datalayout`.** That is why the text and bitcode paths resolve unspecified alignments differently, and it means any consumer resolves them from its own defaults. A backend fix, and it would move the bitcode digest goldens.
- **The compiler's debug metadata has no `Debug Info Version` module flag**, so LLVM discards the whole debug graph on load — visible as `ignoring debug info with an invalid version (0)` in the test output. The `DIFile` fix is still what makes those modules assemble at all, since the rejection happens at parse time.
- **`Bitcode.encode` writes no function-level value symbol table**, so local names exist only in the rendered text. A second, smaller text-vs-bitcode gap, and one that makes the two paths harder to compare.

## Coordination

#151 (which verifies **bitcode** in CI) has merged, and this branch now builds on it rather than around it: the toolchain resolution it introduced is extracted to `test/support/llvmToolchain.ts` and its call site updated, so there is one place to fix when a distribution moves the tools or a newer LLVM needs adding to the search list.

`packages/llvm`'s `fixtures:verify` — which already runs `llvm-as` on rendered output — is still not wired into `test` or `check`, which is why the invalid `zeroinitializer` spelling sat in a fixture unnoticed. Wiring it in still belongs in that lane.

### DIFF
--- a/packages/compiler/test/LlvmIrRoundTrip.test.ts
+++ b/packages/compiler/test/LlvmIrRoundTrip.test.ts
@@ -4,6 +4,7 @@ import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterAll, assert, it } from '@effect/vitest'
 import * as Effect from 'effect/Effect'
+import { llvmToolchain } from '../../../test/support/llvmToolchain.js'
 import * as Analysis from '../src/Analysis.js'
 import type * as Backend from '../src/Backend.js'
 
@@ -13,7 +14,7 @@ const ascii = (value: string): Uint8Array =>
 const directory = mkdtempSync(join(tmpdir(), 'silk-llvm-ir-round-trip-'))
 afterAll(() => rmSync(directory, { recursive: true, force: true }))
 
-const executable = (name: string, variable: string): string => process.env[variable] ?? name
+const toolchain = llvmToolchain(['llvm-as', 'llvm-dis', 'opt'], 'the compiler IR round-trip check')
 
 const write = (name: string, contents: Uint8Array | string): string => {
   const path = join(directory, name)
@@ -23,13 +24,13 @@ const write = (name: string, contents: Uint8Array | string): string => {
 
 const llvmAs = (name: string, source: string): Uint8Array => {
   const assembled = join(directory, name)
-  execFileSync(executable('llvm-as', 'LLVM_AS'), [source, '-o', assembled])
+  execFileSync(toolchain.command('llvm-as'), [source, '-o', assembled])
   return readFileSync(assembled)
 }
 
 const llvmDis = (name: string, bitcode: Uint8Array): string => {
   const output = join(directory, name)
-  execFileSync(executable('llvm-dis', 'LLVM_DIS'), [write(`${name}.bc`, bitcode), '-o', output])
+  execFileSync(toolchain.command('llvm-dis'), [write(`${name}.bc`, bitcode), '-o', output])
   // `llvm-dis` names the module after whichever file it read.
   return readFileSync(output, 'utf8')
     .split('\n')
@@ -52,7 +53,7 @@ const canonical = (name: string, bitcode: Uint8Array): string => {
     write(`${name}.dis.ll`, llvmDis(`${name}.dis`, bitcode)),
   )
   const stripped = join(directory, `${name}.stripped.bc`)
-  execFileSync(executable('opt', 'LLVM_OPT'), [
+  execFileSync(toolchain.command('opt'), [
     '-passes=strip',
     write(`${name}.materialized.input.bc`, materialized),
     '-o',
@@ -70,7 +71,7 @@ const canonical = (name: string, bitcode: Uint8Array): string => {
  */
 const assemble = (name: string, ir: string): string => {
   const assembled = llvmAs(`${name}.assembled.bc`, write(`${name}.source.ll`, ir))
-  execFileSync(executable('opt', 'LLVM_OPT'), [
+  execFileSync(toolchain.command('opt'), [
     '-passes=verify',
     write(`${name}.verify.bc`, assembled),
     '-disable-output',
@@ -200,6 +201,7 @@ for (const mode of ['debug', 'release'] as const) {
       `renders assemblable LLVM IR for the ${name} program in ${mode}`,
       () =>
         Effect.gen(function* () {
+          if (toolchain.unavailable()) return
           const artifact = yield* emit(
             `round-trip/${name}`,
             source,
--- a/packages/compiler/test/LlvmIrRoundTrip.test.ts
+++ b/packages/compiler/test/LlvmIrRoundTrip.test.ts
@@ -1,0 +1,222 @@
+import { execFileSync } from 'node:child_process'
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterAll, assert, it } from '@effect/vitest'
+import * as Effect from 'effect/Effect'
+import * as Analysis from '../src/Analysis.js'
+import type * as Backend from '../src/Backend.js'
+
+const ascii = (value: string): Uint8Array =>
+  Uint8Array.from(value, (character) => character.charCodeAt(0))
+
+const directory = mkdtempSync(join(tmpdir(), 'silk-llvm-ir-round-trip-'))
+afterAll(() => rmSync(directory, { recursive: true, force: true }))
+
+const executable = (name: string, variable: string): string => process.env[variable] ?? name
+
+const write = (name: string, contents: Uint8Array | string): string => {
+  const path = join(directory, name)
+  writeFileSync(path, contents)
+  return path
+}
+
+const llvmAs = (name: string, source: string): Uint8Array => {
+  const assembled = join(directory, name)
+  execFileSync(executable('llvm-as', 'LLVM_AS'), [source, '-o', assembled])
+  return readFileSync(assembled)
+}
+
+const llvmDis = (name: string, bitcode: Uint8Array): string => {
+  const output = join(directory, name)
+  execFileSync(executable('llvm-dis', 'LLVM_DIS'), [write(`${name}.bc`, bitcode), '-o', output])
+  // `llvm-dis` names the module after whichever file it read.
+  return readFileSync(output, 'utf8')
+    .split('\n')
+    .filter((line) => !line.startsWith('; ModuleID ='))
+    .join('\n')
+}
+
+/**
+ * Reduces a module to the form the two emission paths can be compared in.
+ *
+ * Both sides go through `llvm-as` first. The textual parser resolves an omitted `align` against the
+ * module's data layout as it reads, while bitcode leaves the operand unspecified for its consumer
+ * to resolve, so only modules that entered through the same door are comparable. `strip` then
+ * renumbers locals, because `Bitcode.encode` writes no function-level value symbol table and names
+ * therefore survive only along the textual path.
+ */
+const canonical = (name: string, bitcode: Uint8Array): string => {
+  const materialized = llvmAs(
+    `${name}.materialized.bc`,
+    write(`${name}.dis.ll`, llvmDis(`${name}.dis`, bitcode)),
+  )
+  const stripped = join(directory, `${name}.stripped.bc`)
+  execFileSync(executable('opt', 'LLVM_OPT'), [
+    '-passes=strip',
+    write(`${name}.materialized.input.bc`, materialized),
+    '-o',
+    stripped,
+  ])
+  return llvmDis(`${name}.stripped.ll`, readFileSync(stripped))
+}
+
+/**
+ * Assembles rendered IR and reduces it the same way.
+ *
+ * The golden tests compare rendered text against stored rendered text, so a systematic renderer
+ * flaw compares equal to itself and passes forever. `llvm-as` is the check that cannot: either the
+ * text the backend prints is the assembly language, or it is not.
+ */
+const assemble = (name: string, ir: string): string => {
+  const assembled = llvmAs(`${name}.assembled.bc`, write(`${name}.source.ll`, ir))
+  execFileSync(executable('opt', 'LLVM_OPT'), [
+    '-passes=verify',
+    write(`${name}.verify.bc`, assembled),
+    '-disable-output',
+  ])
+  return canonical(`${name}-text`, assembled)
+}
+
+const emit = Effect.fnUntraced(function* (
+  name: string,
+  text: string,
+  request: Backend.CodegenRequest,
+  target: string,
+) {
+  const snapshot = yield* Analysis.ofSourceRealized(name, ascii(text), target)
+  assert.deepEqual(Analysis.diagnostics(snapshot), [])
+  return yield* Analysis.codegen(snapshot, request)
+})
+
+/** A guard whose `Drop` hook runs before its owned `Allocation` field releases. */
+const dropHook = `struct Guard {
+  tag: i32
+  storage: Allocation
+}
+
+impl Drop for Guard {
+  fn drop(self: &mut Guard) -> () { return () }
+}
+
+effect fn build() -> i32 ! OutOfMemory {
+  let mut allocator = SystemAllocator.make()
+  let layout = Layout.of<[i32; 2]>()
+  let recipe = Allocator.allocate(move layout) |> Effect.provideMut(&mut allocator)
+  let allocation = run recipe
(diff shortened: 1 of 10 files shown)


julia-script:
## CI red: `llvm-as` is not on `PATH` on the runner

```
Error: spawnSync llvm-as ENOENT
 ❯ llvmAs test/IrTextRoundTrip.test.ts:48:3
 ❯ assemble test/IrTextRoundTrip.test.ts:97:21
Failed: @silk-effect/llvm#test
```

Both new suites die the same way. `typecheck` and `biome` are clean, and `@silk-effect/llvm` tests pass locally where `llvm-as` exists — I verified both on your branch, so nothing is wrong with the renderer fix or the tests themselves.

**The runner has `clang` but not the LLVM tools on `PATH`.** The native acceptance tests pass, so a compiler is present; `opt`, `llvm-as` and `llvm-dis` are a separate distribution package and are typically installed under `/usr/lib/llvm-<version>/bin/` without being symlinked into `/usr/bin`. `.github/workflows/ci.yml` never installs them.

### This is the same problem PR #151 hit an hour ago

#145's worker found `opt` missing the same way and fixed it by resolving the tool where distributions actually put it — `agent/145-verify-emitted-modules`, commit `1d52987`:

```ts
const opt = [
  process.env.SILK_TEST_OPT,
  'opt',
  '/opt/homebrew/opt/llvm/bin/opt',
  '/usr/local/opt/llvm/bin/opt',
  ...['21', '20', '19', '18', '17', '16'].map((version) => `/usr/lib/llvm-${version}/bin/opt`),
].find((candidate) => candidate !== undefined && spawnSync(candidate, ['--version']).status === 0)
```

Do the same for `llvm-as` and `llvm-dis`. If the tools turn out genuinely absent rather than merely unsymlinked, the alternative is installing them in the workflow — but try resolution first, since #151's fix implies they are present.

**Two packages now need identical tool resolution**, so please factor it rather than copy it — a shared test helper taking a tool name, honoring a `SILK_TEST_<TOOL>` override. Say in your PR body where you put it, and coordinate with `agent/145-verify-emitted-modules` if you touch anything it also touches.

### One thing your PR gets right that #151 does not

**Your test fails loudly when the tool is missing. #151's skips silently and reports green.** That difference is exactly why I have been holding #151 for a CI hard-fail guard — and your red build is the proof: `opt` had been silently absent in every CI run since #151's cross-check was written, and nobody could tell. Yours announced the problem in 61 seconds.

So do not "fix" this by making the round-trip test skip when `llvm-as` is missing. Resolve the tool; if it still cannot be found **in CI**, fail. A round-trip test that skips is a golden test that compares text to itself — the exact failure mode #146 exists to remove.

---

julia-script:
## One addition to `llvmToolchain` — a `console.log` skip reason is invisible

The shared helper is exactly what I asked for: override → `PATH` → distribution prefixes, an actionable CI error naming the env vars, and a log line reporting which binaries it resolved. Extracting it to `test/support/llvmToolchain.ts` and moving #151's now-merged call site onto it is the right shape.

One thing to change, and it comes from a measurement in PR #163 rather than from taste. Its author tested whether a printed skip reason is actually visible:

```
$ vitest run test/diag.test.ts               # console.log in the test body
 Test Files  1 passed (1)
      Tests  1 passed (1)                    # ← the log appears nowhere
```

**The default reporter this repo runs prints neither log output nor annotations.** So this branch of your helper:

```ts
} else if (process.env.CI === undefined) {
  console.log(`${missing.join(', ')} was not found; skipping ${purpose}`)
}
```

produces a fully green local run that says nothing at all. That is the same shape as the failure we just spent an hour removing: #151's `opt` cross-check reported green for its entire life while never running. Hard-failing in CI fixes the CI half — which your helper does correctly — but a contributor running locally still gets a green suite with a silently absent cross-check, and no signal that the strongest test they have did not execute.

**Adopt #163's mechanism:** put the reason in the **skipped test's name**, so every reporter that lists tests shows it, and let the skip raise a non-zero skipped count in the summary. Their form, for reference:

```
 ↓ the host validates a module threading address types through memory instructions — skipped:
   this host's WebAssembly rejects a mixed 64/32-address memory.copy … (see issue #161)

      Tests  8 passed | 1 skipped (9)
```

The run then reports that something did not happen, instead of reporting success.

Note the same `console.log` pattern is live at `packages/compiler/test/ModuleVerification.test.ts:55` — code merged in #151. **You are already refactoring that file onto the helper**, so please fix it there rather than leaving it for a follow-up; the two changes would otherwise collide.

Keep the `console.log` line that reports *which* binaries were resolved on a success — that one is genuinely just informational and costs nothing.

Nothing else outstanding from my side. CI is still running on `751573c`.

---

julia-script:
## CI green — the tools were there, just unsymlinked

Rebased onto `main` @ `841659c` (clean) and pushed `751573c`. Both checks pass, and the log confirms all three suites resolved real binaries and **ran** rather than passing vacuously:

```
the IR round-trip check is using /usr/lib/llvm-18/bin/llvm-as, /usr/lib/llvm-18/bin/llvm-dis, /usr/lib/llvm-18/bin/opt
  ✓ test/IrTextRoundTrip.test.ts (2 tests) 3081ms

the compiler IR round-trip check is using /usr/lib/llvm-18/bin/llvm-as, /usr/lib/llvm-18/bin/llvm-dis, /usr/lib/llvm-18/bin/opt
  ✓ test/LlvmIrRoundTrip.test.ts (4 tests) 10997ms

the LLVM verifier cross-check is using /usr/lib/llvm-18/bin/opt
  ✓ test/ModuleVerification.test.ts (1 test) 29955ms
```

Your diagnosis was exactly right: LLVM 18 is on the runner, unsymlinked, and no workflow change was needed.

### Where the helper went

`test/support/llvmToolchain.ts` — the repository-root `test/support` directory that already holds the cross-package `raise.ts`. It takes a tool *name* rather than being written once per tool, and derives the override from it: `SILK_TEST_OPT`, `SILK_TEST_LLVM_AS`, `SILK_TEST_LLVM_DIS`. Three call sites share it — both round-trip suites and #151's `ModuleVerification.test.ts`, whose inline copy is deleted here. One place to fix when a distribution moves the tools or a newer LLVM needs adding to the list.

### The skip policy is now enforced, not re-decided

I did not convert anything to a skip. The helper owns the policy #151 established — printed reason and skip locally, **hard-fail under `CI`** naming the missing tools and their override variables — so it applies to all three suites instead of being each suite's decision. That also closes the gap you found: #151's cross-check can no longer go silently unrun.

Since this environment has the tools and would never take either branch, I exercised both directly against a nonexistent tool: `CI=1` throws with `SILK_TEST_…` named, unset `CI` returns a skip.

### Gate

`pnpm typecheck` and `biome check` clean repo-wide. Locally `pnpm check` leaves only #160 (`wasm` `Extended.test.ts`) and #159 (`compiler-cli` `FormatWorkflow`, root ignoring `chmod`) — worth noting that **neither reproduces on the GitHub runner**: `wasm` and `compiler-cli` are green above, so both look specific to the agent container. Turbo cancels siblings when one task fails, so I re-ran `@silk-effect/llvm` (66 tests) and `@silk-effect/compiler` (1392 + native acceptance) standalone locally as well: both green.

PR body updated with a new *Finding the toolchain, and refusing to skip* section and a rewritten Coordination section, since #151 has merged and this now builds on it rather than around it.

---

PR #151, opened 2026-08-13
feat(backend): verify every emitted LLVM and Wasm module

### LINKED ISSUE #145: backend: verify emitted LLVM IR and Wasm modules in CI

julia-script:
Closes #145

Both backends now check their own output before it leaves `emit`, so every test that already emits a module is a verifier test.

- **LLVM** — `@silk-effect/llvm/Verify` is the in-process form of `opt -passes=verify` for the invariants a builder can break on its own: SSA dominance of every use by its definition, phi incoming values dominating their incoming edges, phis grouped at the top of their block, and in-range branch destinations. It reads the committed function bodies, which are exactly what `Bitcode.encode` serializes — not `IrText.render` output, which does not round-trip (#146).
- **Wasm** — `@silk-effect/wasm/Validate` puts the encoded module through the host's own implementation. `WebAssembly.validate` is the fast answer; a rejected module is compiled a second time to recover the host's diagnostic, which names the offending function.
- Both report through one `BackendError` reason, `InvalidModule`, carrying `{ function, message, detail }` per violation.
- Wasm body validation already happened inside the builder (`Func.define`), but its failure named only the operation. It is now wrapped at the backend boundary so it names the function too.

**Why in-process rather than shelling out to `opt`.** The issue allows either. In-process always runs — no toolchain dependency, no per-module process spawn, and it can name blocks and values from the builder's own naming. `opt` is used as the cross-check instead: `ModuleVerification.test.ts` emits every corpus program and runs `opt -passes=verify` on the bitcode. It looks for the binary on `PATH`, under `SILK_TEST_OPT`, and in the versioned LLVM prefixes (`opt` ships in LLVM's tools package, not alongside `clang`), prints which one answered, and **fails in CI when it finds none** — a cross-check that turns itself off when its reference implementation is missing reports success in exactly the situation where it verified nothing.

### What the in-process verifier does not cover

An empty result is not the claim that `opt` would have nothing to say. Deliberately outside it: operand and result type agreement, call-site agreement with the callee's signature and calling convention, attribute and metadata well-formedness, and intrinsic signatures. The `opt` cross-check is what covers that gap over real emitted modules — which is why it must not be skippable.

### Public API

`@silk-effect/llvm/Verify` and `@silk-effect/wasm/Validate` are intentionally public actors, with the compatibility obligations that carries. `@silk-effect/compiler` imports both across a package boundary, and the packages' internal-only mechanism is `src/internal/**`, which is not reachable from another package — `Bitcode`, `IrText`, `Binary` and `WatText` are public for the same reason. AGENTS.md: *"For a public actor, add its explicit package subpath export and prefer the deep import."* The `release-candidate/validate.test.ts` update is therefore a deliberate surface change, not a workaround.

## Acceptance criteria

- [x] Reverting PR #144's fix makes CI fail with a dominance violation naming `removeDirectoryRecursively`, not a clang crash
- [x] Reverting PR #123's fix makes Wasm validation fail for an 8-byte lane element
- [x] A clean tree passes both checks
- [x] The checks are part of `pnpm check` — they run inside `Backend.emit`, which `turbo run test` exercises; `pnpm check` runs that, and CI already runs `pnpm check` on `pull_request` and on pushes to `main`. No workflow change was needed for the always-on checks.

## Revert experiment 1 — PR #144 (LLVM)

Commenting out the five `reloadMutableRoots` calls at the emitter's own joins, then `vitest run test/TemporaryDirectoryAcceptance.test.ts`:

```
BackendFailed: LLVM emitted an invalid module for temporary-directory/native-tree (50 violation(s)):
Instruction does not dominate all uses! (in @silk_silk_filesystem_removeDirectoryRecursively_effect_73_23769_provided_54__15_…)
  %reload4_0_294 defined in %propagation_release80_u1_drop
  used by Store in %propagation_release80_u1_next
Instruction does not dominate all uses! (in @silk_silk_filesystem_removeDirectoryRecursively_effect_73_23769_provided_54__15_…)
  %reload4_1_294 defined in %propagation_release80_u1_drop
  used by Store in %propagation_release80_u1_next
...
```

Three tests fail this way (`native` 110 violations, `native-tree` 50, `native-many` 110). The count, the function, and the value name all match PR #144's report — `%reload4_0_294` is the same value that PR's write-up names.

Cross-checked against the tool. Dumping that module's bitcode and running the command from the issue:

```
$ opt -passes=verify -disable-output temporary_directory_native_tree.bc
Instruction does not dominate all uses!
  %25949 = load i32, ptr %25948, align 4
  store i32 %25949, ptr %26175, align 4
...
$ opt -passes=verify -disable-output temporary_directory_native_tree.bc 2>&1 | grep -c "does not dominate"
50
```

50 from `opt`, 50 from the in-process verifier, on the same module — and `%25949 = load i32, ptr %25948` is the exact instruction quoted in #130.

On the clean tree, `opt` accepts every corpus module the LLVM backend emits (the new `ModuleVerification.test.ts`), and the in-process verifier reports nothing on any module emitted anywhere in the suite.

## Revert experiment 2 — PR #123 (Wasm)

Reverting the three lane-width sites (`loadAt(address, offset, lane)` → `loadAt(address, offset)` in two places, `laneStoreMnemonic(...)` → `'i32.store'` in one), then `vitest run test/SlotLaneWidth.test.ts -t u64`:

```
BackendError: WebAssembly emitted an invalid body for slot-lane-width/u64 (1 violation(s)):
Expected i64 on the stack but found i32 (in silk_silk_slot_copy__9_73696c6b2f736c6f74_4_636f7079_11_6275696c74696e3a753634_…)
```

**One honest qualification.** The 8-byte-lane case is rejected *before* a module exists: the wasm builder's own body validator fails at `Func.define`, so host validation never sees it — which is precisely #114's "never emitted a module at all". Without this PR the same revert failed with `Func.define: Expected i64 on the stack but found i32` and no function name at all; the failure now names the body that produced it, which is what requirement 3 asks for. Host validation covers the class the builder's validator structurally cannot: a bug in the *encoder*, where the builder's view is valid and the bytes are not.

## The CI guard, and the `opt` question it settles

`ModuleVerification.test.ts` fails in CI when `opt` cannot be found, rather than skipping green. Both branches are exercised, not assumed — with the candidate list temporarily replaced by a bogus path:

```
$ CI=true pnpm exec vitest run test/ModuleVerification.test.ts
AssertionError: opt was not found in CI, so the LLVM verifier cross-check cannot run. …
 Test Files  1 failed (1)

$ pnpm exec vitest run test/ModuleVerification.test.ts     # no CI
 Test Files  1 passed (1)
```

**CI is green on `484df80`, the commit that adds the guard** — so the cross-check ran against a real `opt` on the runner, since its absence would have failed the build. `ubuntu-latest` has the binary; it is simply not always the bare `opt` on `PATH`, which the prefix search handles. No LLVM install is needed in the workflow.

Whether the runs before `1d52987` found `opt` or skipped silently is not established — those job logs were unreadable from this sandbox (`curl: (56) CONNECT tunnel failed, response 403`), and the versioned-prefix search was defensive rather than diagnostic. The guard makes the question unable to recur: a runner without LLVM tools now produces a red build instead of a green check that verified nothing.

## Tests

Baseline 0 real failures, after 0 real failures, +11 new tests.

Per package, on this branch:

| package | result |
|---|---|
| `compiler` | 170 files, 1369 tests, all passing (+ the serial `DriverNativeAcceptance` run, 1 test) |
| `llvm` | 14 files, 64 tests, all passing |
| `wasm` | 9 files, 82 tests, 1 failing — `Extended` *"threads address types through memory instructions"*, the documented pre-existing failure |
| `compiler-cli` | 16 files, 92 tests, 3 failing — `FormatCommand` / `FormatWorkflow`, the documented chmod-vs-root failures |
| `lsp` | 78 tests, all passing |
| `documentation` | 4 tests, all passing |
| `docs` | 89 tests, all passing |
| `release:candidate` | 7 tests, all passing |

The two failing sets are the documented-ignorable ones and are unrelated to this change; both pass in CI, where the job does not run as root. CI on `484df80` is green.

New tests:

- `packages/llvm/test/Verify.test.ts` — 5: a well-formed diamond, a value escaping its arm into the join (asserting the exact violation text and `@escapes` as the function), a valid phi, a phi reading an arm across the wrong edge, and a body-less declaration.
- `packages/wasm/test/Validate.test.ts` — 5: hand-built valid and invalid modules, a builder-encoded module, function-name recovery from the host's diagnostic (`#0` when unnamed, `answer` when the `name` section names it), and bytes that are not a module.
- `packages/compiler/test/ModuleVerification.test.ts` — 1: `opt -passes=verify` agrees with the in-process verifier across the corpus.

Two follow-up commits pinned the new `Verify` / `Validate` subpath exports in `release-candidate/validate.test.ts`, which asserts the exact export and barrel-namespace lists.

## Findings

No violations were found in any module beyond the two known ones. Every module the suite emits — LLVM and wasm, across the corpus and every acceptance suite — passes both checks on a clean tree.


julia-script:
## Review — strong PR, one thing to change before merge

The evidence here is the best of the run so far. Both revert experiments were actually run, and the `opt` cross-check makes the first one conclusive rather than merely consistent: **50 dominance violations from the in-process verifier and 50 from `opt`, on the same module**, with `%25949 = load i32, ptr %25948` matching the instruction quoted in #130. That is the difference between "my verifier reports something" and "my verifier reports what the reference implementation reports." The honest qualification on the Wasm side — that the 8-byte lane is rejected at `Func.define` before a module exists, so host validation never sees it — is exactly the kind of thing that usually gets quietly rounded up to a checked box.

Choosing in-process verification with `opt` as a cross-check, rather than shelling out per module, is the right call and better than what the issue asked for.

### The one change: the cross-check can silently disable itself

`ModuleVerification.test.ts:21` is:

```ts
const optAvailable = spawnSync('opt', ['--version'], { encoding: 'utf8' }).status === 0
```

and on a miss the test logs a note and returns green.

That is the failure mode this PR exists to eliminate. It is the same shape as the golden tests in #146 — *"they compare rendered text against stored rendered text, so a systematic flaw compares equal to itself and passes indefinitely"* — and the same shape as PR #84's move-only assertions that never read their value. A cross-check that turns itself off when its reference implementation is missing is a check that reports success in precisely the situation where it has verified nothing.

It matters concretely here: **`.github/workflows` never installs LLVM.** The repo takes clang from whatever the `ubuntu-latest` image happens to provide, and `opt` is frequently not symlinked into `/usr/bin` on that image even when clang is. So this test may well be skipping on every CI run right now — and neither of us can tell from a green check mark, which is the whole problem.

Please make the skip conditional on not being in CI:

```ts
if (!optAvailable) {
  if (process.env.CI) throw new Error('opt is not on PATH in CI; the LLVM verifier cross-check cannot be skipped here')
  console.log('opt is not on PATH; skipping the LLVM verifier cross-check')
  return
}
```

Locally it stays a convenience skip. In CI it becomes a hard failure, so the check is either running or loudly absent — never silently neither. If it turns out `opt` genuinely is not on the runner, that is a one-line `apt-get install` in the workflow, and better discovered now than the next time someone trusts this test.

Also please say in the PR body **whether `opt` was present in this PR's own CI run** — reading it out of the job log settles it directly.

### Not blocking

The in-process verifier covers dominance, phi placement, phi incoming edges, and branch destinations. Worth a sentence in the PR body on what it deliberately does *not* cover versus `opt -passes=verify`, so the next person reading a green in-process check knows its boundary — type well-formedness, calling-convention agreement, and metadata validity are the obvious candidates. The cross-check test is what covers the gap today, which is another reason it must not be skippable.

Happy to merge as soon as the CI guard is in.

---

julia-script:
## CI is red — the release-candidate export surface, not your verifiers

`validate` failed on `38560617` ([job 94561882117](https://github.com/julia-script/silk-effect/actions/runs/31734233753/job/94561882117)). Two assertions in `release-candidate/validate.test.ts`, at `:107` and `:1177`:

```
AssertionError: expected [ '.', './Binary', './Builder', …(17) ] to deeply equal [ '.', './Binary', './Builder', …(16) ]

@@ -14,8 +14,9 @@
    "./Type",
    "./ValType",
+   "./Validate",
    "./WasmError",
    "./WatText",
```

That test pins the **exact public export surface** of each package, and this PR adds two new public entry points — `@silk-effect/wasm/Validate` and (at `:107`) the `@silk-effect/llvm` equivalent. Nothing is wrong with the verifiers; the release gate simply has to be told the surface grew.

Worth a moment's thought rather than a reflexive list update, since this is a **public API decision**: `Verify` and `Validate` need to be importable by `@silk-effect/compiler`, so exporting them is right — but they then become part of the released surface of both packages, with the compatibility obligations that carries. If you intend them as internal-but-cross-package, the repo may have a convention for that worth following instead. Say which you chose in the PR body.

## Still outstanding from my earlier review

1. **The `opt` skip must be a hard failure in CI.** Unchanged and still blocking — see [my earlier comment](https://github.com/julia-script/silk-effect/pull/151#issuecomment-5285287199).
2. **Whether `opt` was actually present in this run.** I could not settle it from the job log tail, and it is the single fact that decides whether your cross-check has ever run in CI. Please grep your own run's log for `opt is not on PATH` and report the answer in the PR body. If it *is* absent, add the LLVM install to the workflow — that turns the cross-check from decorative into real, which is the difference between this PR catching the next #130 and not.

Everything else in the review stands: the revert experiments and the 50-vs-50 agreement with `opt` are the strongest evidence in the run so far.

---

julia-script:
## `1d52987` answers the question — and confirms the cross-check had never run in CI

"find opt where the distributions put it" is the right fix, and searching `/usr/lib/llvm-<version>/bin/opt` is exactly where `ubuntu-latest` keeps it. But the change itself is the answer to my question: **you would not have needed it if `opt` had been on `PATH`.** So the `opt` cross-check — the thing that makes the #144 revert experiment conclusive rather than merely self-consistent — was silently skipping on every CI run since it was written, reporting green.

That is the finding, and it is worth stating plainly in the PR body rather than leaving implicit in a commit message. It is also the single best argument for the change I asked for, which `1d52987` did **not** include:

```ts
if (opt === undefined) {
  console.log('opt was not found; skipping the LLVM verifier cross-check')
  return
}
```

Better detection reduces the chance of a silent skip. It does not remove it. A runner image that drops LLVM tools, a version outside the `16..21` list, a container without them — each puts you straight back to a green check mark that verified nothing, and the last hour is direct evidence that nobody notices when that happens. Adding `console.log('cross-checking … against ${opt}')` makes it *observable*, which is a real improvement, but observable-if-someone-reads-the-log is not the same as enforced.

Please add the hard failure in CI:

```ts
if (opt === undefined) {
  if (process.env.CI)
    throw new Error('opt was not found in CI; the LLVM verifier cross-check cannot be skipped here')
  console.log('opt was not found; skipping the LLVM verifier cross-check')
  return
}
```

Local runs keep the convenience skip. CI gets a loud failure instead of a quiet lie. If a future image genuinely lacks the tools, that is a one-line install in the workflow — and you want to find out from a red build, not from the next miscompile that slips through.

This is the last blocker from my side. The export-surface fix in `release-candidate/validate.test.ts` looks right; say in the PR body whether you treated `Verify`/`Validate` as intentionally public API.

---

julia-script:
## CI guard added — `484df80`

The blocker is in. `ModuleVerification.test.ts` now fails in CI when `opt` cannot be found, and keeps the convenience skip locally:

```ts
if (opt === undefined) {
  if (process.env.CI !== undefined) {
    return assert.fail(
      'opt was not found in CI, so the LLVM verifier cross-check cannot run. Install LLVM in the workflow, or point SILK_TEST_OPT at the binary.',
    )
  }
  console.log('opt was not found; skipping the LLVM verifier cross-check')
  return
}
```

You are right that this is the same shape as #146's goldens, and better detection is not a substitute for enforcement. Both branches are exercised rather than assumed — I temporarily replaced the candidate list with a bogus path and ran it twice:

```
$ CI=true pnpm exec vitest run test/ModuleVerification.test.ts
AssertionError: opt was not found in CI, so the LLVM verifier cross-check cannot run. …
 Test Files  1 failed (1)

$ pnpm exec vitest run test/ModuleVerification.test.ts     # no CI
 Test Files  1 passed (1)
```

## On whether `opt` was present in this PR's CI run — I do not know, and I should not have implied otherwise

I have to correct the inference in your last comment, because it credits me with a finding I did not make. `1d52987` was **defensive, not diagnostic**: I added the prefix search because I could not read the job log, not because I had established `opt` was missing. My attempt to fetch it was blocked by this sandbox's egress proxy:

```
$ curl -sS -o ci.log "https://productionresultssa8.blob.core.windows.net/.../job-logs.txt?..."
curl: (56) CONNECT tunnel failed, response 403
```

and `get_job_logs` only returns a tail, which lands in the `pnpm release:candidate` failure rather than the compiler test output. So "the cross-check had never run in CI" is plausible but **unconfirmed**, and I am not going to assert it in the PR body as though I had checked.

The guard settles it on the next run, which is a better answer than either of us reading a log: **green ⇒ `opt` was found**, and the run prints which binary answered (`cross-checking the LLVM verifier against …`). **Red ⇒ it is genuinely absent**, and I will add the LLVM install to `.github/workflows/ci.yml` and re-push. Either way the ambiguity ends with this run rather than persisting behind a skip. I will report the outcome here and put the confirmed answer in the PR body.

## `Verify` and `Validate` are intentionally public

Deliberate, and it is the only option consistent with the repo. `@silk-effect/compiler` imports them across a package boundary, and the packages' internal-only mechanism is `src/internal/**`, which is not reachable from another package — `Bitcode`, `IrText`, `Binary` and `WatText` are all public for the same reason. AGENTS.md states it directly: *"For a public actor, add its explicit package subpath export and prefer the deep import."* So both are public actors carrying the usual compatibility obligations, which is why the `release-candidate` update is a list edit rather than a workaround. Recorded in the PR body.

## Non-blocking item taken

`Verify.verify`'s doc now names its boundary, so a green in-process result is not misread as the full verifier's silence:

> An empty result means the module is well-formed by those invariants, not that `opt` would have nothing to say. Deliberately outside this check: operand and result type agreement, call-site agreement with the callee's signature and calling convention, attribute and metadata well-formedness, and intrinsic signatures. Those are covered by the cross-check against `opt -passes=verify` over emitted modules, not here.

Which is, as you note, another reason the cross-check must not be skippable — it is the only thing covering that gap.

---

julia-script:
## CI green on `484df80` — with the guard active, which answers it

[Run 31739718047](https://github.com/julia-script/silk-effect/actions/runs/31739718047) passed on `484df804`, the commit that makes a missing `opt` a hard failure in CI. That is the answer, and it is a stronger one than reading a log line: **the cross-check ran against a real `opt` on the runner.** Had it not found one, the guard would have failed the build — that branch is exercised, not assumed.

So no LLVM install is needed in the workflow. `ubuntu-latest` has the binary; it simply is not always the bare `opt` on `PATH`, which is what the prefix search handles.

### What remains unknown, and why it no longer matters

Whether the pre-`1d52987` runs found `opt` on `PATH` or skipped silently is still not established — I could not read those job logs, and I am not going to infer it backwards from the fact that I added a fallback. What changed is that the question can no longer recur: from here, a runner without LLVM tools produces a red build rather than a green check mark that verified nothing. That is the property you asked for, and it is now enforced rather than observed.

The PR body records the confirmed result and drops the unresolved-question section.

### Full state

| | |
|---|---|
| CI on `484df80` | green — `validate` and `webcontainer-browser` |
| `opt` cross-check in CI | ran, against the runner's LLVM tools |
| Local suite | 0 real failures; only the documented `wasm` `Extended` and `compiler-cli` Format chmod-vs-root failures, both green in CI |
| `pnpm release:candidate` | 7/7 |
| Revert experiments | both run, output in the PR body |

Ready for merge from my side. The three review items are in: the CI guard (`484df80`), the public-API answer, and the documented boundary of the in-process verifier.

---

PR #156, opened 2026-08-13
test(compiler): fail when two diagnostic constants share a stable code

### LINKED ISSUE #152: tooling: nothing enforces that stable diagnostic codes are unique

julia-script:
Closes #152

## The gap

A stable diagnostic code is a public contract — it appears in `diagnostics.md`, in user-facing messages, and in tests that assert on it by string. Nothing enforced that two constants cannot claim the same one.

- **The guard test did not check uniqueness.** `DocumentationExamples.test.ts:82` is `assert.include(diagnostics, \`${code}\`, …)`, a per-code *containment* check. Two constants holding `'SEM0097'` produce two iterations that both find the string present, and both pass.
- **The generator did not reject a duplicate.** `generate-documentation.mjs` keys `documented` by code, so a second constant claiming a taken code silently displaces the first. The published index would document one meaning while the compiler reports two. (The issue predicted two rows; the map keying makes the real behaviour a silent drop, which is quieter still.)

Concurrent branches each take "the next free code" against a `main` that lacks the other's, so a collision is only ever visible on the merge result. Git surfaces it only when both branches happen to insert at the same line — the `SEM0097` case. `SEM0098` had no shared line, and would have merged clean.

## The change

**`test/DocumentationExamples.test.ts`** — a new test asserts every `export const <name>Code` holds a distinct value, and reports each colliding code with **both** constant names. The parse is now shared with the existing `diagnostics.md` containment check rather than duplicated.

It also pins that every stable-code literal in `Diagnostic.ts` is reachable through the declaration pattern, since a literal the pattern cannot see is a code the uniqueness check cannot cover.

**`scripts/generate-documentation.mjs`** — collects constants that claim an already-taken code and exits non-zero naming both, instead of rendering the index. Both pages now render before either is written, so a rejected index leaves the documentation as it was rather than half-regenerated.

## Requirement 3 — meaningful on the merge result

No CI change was needed. `.github/workflows/ci.yml` runs `pnpm check` on `push: branches: [main]`, and `pnpm check` reaches both guards: the compiler package's `test` script runs `documentation:check` and then the vitest suite. A collision that merges clean is caught by the first `main` build after the second branch lands — which is the only place it can be caught, since it requires both branches to exist.

## Verification

Recorded in a follow-up comment: the guard is exercised against a deliberate duplicate placed in a **different region** of `Diagnostic.ts` from the constant it collides with — a `LEX`-region constant taking `SEM0097` — since the same-region case is the one git already covers.

## Out of scope

The issue floats allocating codes from a checked-in registry, or deriving them from constant names. Neither is implemented here, per the issue's own note that they only matter if collisions recur once a guard exists.

No diagnostic was renumbered; the existing `SEM0097` / `SEM0098` split is untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FSP7na8nv6TfhSSBBbLWTS)_

### DIFF
--- a/packages/compiler/scripts/generate-documentation.mjs
+++ b/packages/compiler/scripts/generate-documentation.mjs
@@ -126,6 +126,7 @@ const diagnosticsPage = () => {
   const source = readFileSync(diagnosticSource, 'utf8')
 
   const documented = new Map()
+  const collisions = new Map()
   // `[^*]` in the comment body keeps the match from reaching back past an earlier `*/`.
   const declaration =
     /(?:\/\*\*((?:[^*]|\*(?!\/))*)\*\/\s*)?export const (\w+)Code = '([A-Z]{3}[0-9]{4})' as const/g
@@ -134,7 +135,24 @@ const diagnosticsPage = () => {
       .replace(/^\s*\*\s?/gm, '')
       .replace(/\s+/g, ' ')
       .trim()
+    // Keyed by code, so a second constant claiming a taken code would silently displace the first
+    // and this page would document one meaning while the compiler reports two. Refuse instead: a
+    // published index that resolves a collision by dropping a diagnostic is worse than no index.
+    const taken = documented.get(match[3])
+    if (taken !== undefined)
+      collisions.set(match[3], [
+        ...(collisions.get(match[3]) ?? [`${taken.name}Code`]),
+        `${match[2]}Code`,
+      ])
     documented.set(match[3], { name: match[2], comment })
+  }
+
+  if (collisions.size > 0) {
+    console.error('Duplicate stable diagnostic codes in src/Diagnostic.ts:')
+    for (const [code, names] of collisions)
+      console.error(`  ${code} is held by ${names.join(' and ')}`)
+    console.error('A stable code identifies one condition. Renumber the newer constant.')
+    process.exit(1)
   }
 
   // The constructor bodies carry the user-visible message template, which is the most precise
@@ -231,5 +249,10 @@ const write = (name, contents) => {
   writeFileSync(destination, contents)
 }
 
-write('stdlib.md', await stdlibPage())
-write('diagnostics.md', diagnosticsPage())
+// Both pages are rendered before either is written, so a rejected diagnostic index leaves the
+// documentation exactly as it was rather than half-regenerated.
+const stdlib = await stdlibPage()
+const diagnostics = diagnosticsPage()
+
+write('stdlib.md', stdlib)
+write('diagnostics.md', diagnostics)
--- a/packages/compiler/test/DocumentationExamples.test.ts
+++ b/packages/compiler/test/DocumentationExamples.test.ts
@@ -5,6 +5,7 @@ import { assert, it } from '@effect/vitest'
 import * as Effect from 'effect/Effect'
 import * as Analysis from '../src/Analysis.js'
 import * as Stdlib from '../src/Stdlib.js'
+import { raise } from './support/raise.js'
 
 const ascii = (value: string): Uint8Array =>
   Uint8Array.from(value, (character) => character.charCodeAt(0))
@@ -57,6 +58,26 @@ const documents = readdirSync(documentationRoot)
 
 const blocks = documents.filter((entry) => !generated.has(entry)).flatMap(blocksOf)
 
+const diagnosticText = readFileSync(diagnosticSource, 'utf8')
+
+interface Declaration {
+  readonly name: string
+  readonly code: string
+}
+
+// Every stable code is declared as `export const <name>Code = '<CODE>' as const`. Reading the
+// source rather than the module keeps the constant name, which a collision report has to carry.
+const declarations: ReadonlyArray<Declaration> = [
+  ...diagnosticText.matchAll(/export const (\w+Code) = '([A-Z]{3}[0-9]{4})' as const/g),
+].map((match) => ({
+  name: match[1] ?? raise('the declaration pattern must capture a constant name'),
+  code: match[2] ?? raise('the declaration pattern must capture a stable code'),
+}))
+
+const codeLiterals = [...diagnosticText.matchAll(/'([A-Z]{3}[0-9]{4})' as const/g)].map(
+  (match) => match[1] ?? raise('the literal pattern must capture a stable code'),
+)
+
 it('finds Silk examples in the tutorial and the reference', () => {
   assert.isTrue(
     documents.includes('tutorial.md'),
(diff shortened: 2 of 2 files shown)


julia-script:
## Verification against a deliberate duplicate

The probe is placed in a **different region** of `Diagnostic.ts` from the constant it collides with — the same-region case is the one git already covers, so it proves nothing. `collidingProbeCode` is declared in the `LEX` block at line 38, immediately after `LEX0007`; the constant it collides with, `ambiguousBoundOperationCode`, is at line 199. **161 lines apart, no shared line, no textual conflict** — this is the `SEM0098` shape, the one that merges clean and ships.

```diff
--- a/packages/compiler/src/Diagnostic.ts
+++ b/packages/compiler/src/Diagnostic.ts
@@ -33,6 +33,9 @@
 /** Stable code for a character literal that denotes a number of scalars other than one. */
 export const characterLiteralScalarCountCode = 'LEX0007' as const
+
+/** Deliberate collision for issue #152 verification: a different region of the file. */
+export const collidingProbeCode = 'SEM0097' as const
```

```console
$ grep -n "collidingProbeCode\|ambiguousBoundOperationCode = " packages/compiler/src/Diagnostic.ts
38:export const collidingProbeCode = 'SEM0097' as const
199:export const ambiguousBoundOperationCode = 'SEM0097' as const
```

### The test fails, naming both constants

```console
$ pnpm --filter @silk-effect/compiler exec vitest run test/DocumentationExamples.test.ts -t 'distinct stable code'

 ❯ test/DocumentationExamples.test.ts (27 tests | 1 failed | 26 skipped) 11ms
   × gives every diagnostic constant a distinct stable code 9ms

⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/DocumentationExamples.test.ts > gives every diagnostic constant a distinct stable code
AssertionError: each stable diagnostic code must belong to exactly one constant; renumber the newer one: expected [ Array(1) ] to deeply equal []

- Expected
+ Received

- []
+ [
+   "SEM0097 is held by collidingProbeCode and ambiguousBoundOperationCode",
+ ]

 Test Files  1 failed (1)
      Tests  1 failed | 26 skipped (27)
```

### The generator refuses, in both modes

```console
$ pnpm --filter @silk-effect/compiler documentation:check
Duplicate stable diagnostic codes in src/Diagnostic.ts:
  SEM0097 is held by collidingProbeCode and ambiguousBoundOperationCode
A stable code identifies one condition. Renumber the newer constant.
Exit status 1

$ pnpm --filter @silk-effect/compiler documentation:generate
Duplicate stable diagnostic codes in src/Diagnostic.ts:
  SEM0097 is held by collidingProbeCode and ambiguousBoundOperationCode
A stable code identifies one condition. Renumber the newer constant.
Exit status 1

$ node scripts/generate-documentation.mjs --check >/dev/null 2>&1; echo "check exit=$?"
check exit=1
$ node scripts/generate-documentation.mjs >/dev/null 2>&1; echo "generate exit=$?"
generate exit=1

$ git status --short packages/language/docs/
```

The empty `git status` is the point of rendering both pages before writing either: the rejected run left `diagnostics.md` and `stdlib.md` exactly as they were, rather than writing a fresh `stdlib.md` and then bailing.

### Clean run after reverting the probe

```console
$ git checkout packages/compiler/src/Diagnostic.ts
$ grep -c collidingProbeCode packages/compiler/src/Diagnostic.ts
0

$ pnpm --filter @silk-effect/compiler test
$ node scripts/generate-stdlib.mjs --check
$ node scripts/generate-documentation.mjs --check
$ vitest run --passWithNoTests --exclude test/DriverNativeAcceptance.test.ts
 Test Files  169 passed (169)
      Tests  1369 passed (1369)
$ vitest run test/DriverNativeAcceptance.test.ts --passWithNoTests --maxWorkers=1
 Test Files  1 passed (1)
      Tests  1 passed (1)

$ pnpm biome check .
Checked 564 files in 1290ms. No fixes applied.

$ pnpm --filter @silk-effect/compiler typecheck
$ node scripts/generate-stdlib.mjs --check
$ tsc -p tsconfig.json --noEmit && tsc -p tsconfig.test.json
```

`documentation:check` is the first step of the compiler package's `test` script, so the clean generator run above is part of that green suite. The existing `SEM0097` / `SEM0098` split is untouched, and `diagnostics.md` is byte-identical to `main` — no diagnostic was renumbered by this PR.

### On the full gate

`pnpm check` is **red in this sandbox on pristine `origin/main`**, before any of my changes — `@silk-effect/wasm#test` fails at `test/Extended.test.ts > threads address types through memory instructions`, with knock-on failures in `llvm`, `lsp`, `compiler-cli`, `docs` and `documentation`. I confirmed this by running the full gate in a separate worktree checked out at `origin/main`: identical outcome, `21 successful, 30 total, Failed: @silk-effect/wasm#test`. These look environmental rather than real (the `compiler-cli` cases are the permission/storage-failure simulations, which do not fail when the suite runs as root).

`@silk-effect/compiler` — the only package this PR touches — is fully green on this branch, as shown above. CI on GitHub is the authority for the rest.

---

PR #157, opened 2026-08-13
fix(compiler): lower a bound operation against a source witness

### LINKED ISSUE #155: generics: a non-operator bound operation over a source witness passes analysis and lowers to nothing
## Current behavior

A `Bound.operation(args)` call whose witness is a **function of the provider's own actor** (a source witness, admitted by #129 → PR #142) is analyzed cleanly and then **lowers to nothing**. The specialized instance fails MIR validation with `InvalidCallShape`, and **no diagnostic is reported**.

`packages/compiler/src/Lower.ts:2619` lowers a bound operation by reading the intrinsic the witness names:

```ts
if (selected?.rule._tag !== 'BuiltinRule') return undefined
```

A source witness is read only from the **operator** path, through `lowerInterfaceWitnessCall`. So the four combinations a generic body can reach are covered unevenly:

| | operator-spelled (e.g. `==`) | non-operator (`Bound.operation(…)`) |
|---|---|---|
| **intrinsic witness** (#118 → PR #141) | ✅ | ✅ |
| **source witness** (#129 → PR #142) | ✅ | ❌ **does not lower** |

Confirmed by test on `agent/34-hash-collections` — `packages/compiler/test/HashKeyBoundForm.test.ts` (PR #149, draft), where the failing combination is **asserted rather than skipped**, so it starts passing the moment this is fixed.

## How it got here, and the lesson

**PR #141 and PR #142 landed within minutes of each other, and neither wired the other's case.** Each was green on a base that did not contain the other; they touched adjacent-but-different concerns in `DeclarationIndex.ts` and never conflicted textually. The composition is what is broken.

PR #141's own `design.md` predicted this seam:

> Recording the question rather than an answer is also what keeps the node honest under #129: when witness admissibility widens, the node does not change — only what lowering finds at the end of the conformance does.

Admissibility widened in #142. Lowering was never taught to look.

This is the third instance tonight of the same orchestration failure — *green-on-the-PR is not green-on-`main`* — after the `LEX0006` documentation break and the `SEM0097`/`SEM0098` code collisions (#152). Merging two independent, individually-green PRs back to back does not compose their guarantees, and nothing in CI checks the composition.

## Why this is P1

It blocks **#34** (`HashMap`/`HashSet`, P1, self-hosting prerequisite) **entirely** — not one acceptance criterion.

The tempting reading is "user-defined keys are blocked, scalar keys still work." That is wrong. A scalar provider's conformance admits a sealed intrinsic **and nothing else** (`impl HashKey<u64> for u64 { hash: u64.someSilkFunction }` is rejected — confirmed by test). No intrinsic computes a hash, and #34's requirement 9 forbids adding one. So a scalar key has no honest `hash` witness to name either.

Every key type needs the source-witness path: scalars to have a hash **at all**, user types to have one that is **reachable**.

## The second defect, independent of #34

**A program that passes analysis and then produces invalid MIR, silently, is its own bug.** Whatever fixes the lowering must also guarantee the compiler never does this again for an unlowerable call: either lower it, or report a diagnostic. Failing MIR validation with no user-visible diagnostic is the same silent-failure class as #114, #130 and #86 — three miscompiles this month that were invisible to tests asserting on completion rather than on values.

## Requirements

1. A non-operator bound operation lowers against a source witness, reaching the provider's own function — the fallback the operator path already has via `lowerInterfaceWitnessCall`.
2. Instance discovery walks the same conformance, so the specialization exists to be called.
3. **No call that passes analysis may lower to nothing without a diagnostic.** If a witness cannot be lowered, that is a reported error, not silent invalid MIR.
4. Three-engine parity: bootstrap evaluator, native LLVM, and direct WebAssembly agree.
5. No hash operation is added anywhere — this fix is orthogonal to #34's requirement 9.

## Acceptance criteria

- [ ] `HashKeyBoundForm.test.ts`'s fourth combination passes without modification to the test's expectations.
- [ ] A test calls a non-operator bound operation over a user-defined provider and asserts the **value returned**, on all three engines — not merely that compilation completes.
- [ ] A test asserts that an unlowerable witness produces a diagnostic rather than invalid MIR.
- [ ] Existing operator-path and intrinsic-witness behavior is unchanged.

## Related

- #141 / #142 — the two PRs whose seam this is.
- #34 — blocked on this; PR #149 charters the spec and stops here deliberately.
- #145 — verifying emitted LLVM/Wasm modules in CI. The MIR-level equivalent is what would have caught this at the source; worth considering whether MIR validation should be a gate rather than an internal assertion.
- #152 — the other composition failure from the same merge train.

---

Filed by orchestrator run `run-r7x1` from PR #149's confirmation pass, which found this by testing the bound form before writing collection code — exactly what #34's implementation note asked for.

julia-script:
Closes #155.

## What was broken

`Bound.operation(args)` lowered by reading the intrinsic its witness named:

```ts
if (selected?.rule._tag !== 'BuiltinRule') return undefined
```

A source witness — a function of the provider's own actor, admitted by #129 → PR #142 — was reached only from the operator path, through `lowerInterfaceWitnessCall`. PR #141 and PR #142 landed independently and neither wired the other's case, so the fourth combination (non-operator spelling, source witness) passed analysis and lowered to nothing: the specialized instance fell back to a trap function whose signature no longer matched its callers, and MIR validation failed with `InvalidCallShape` — with no diagnostic reaching the user.

## 1. Make it lower

- **`Lower.ts`** — a bound operation whose conformance does not name a lowerable intrinsic now falls through to the provider's own function, the same fallback the operator path already has. Operand borrows are taken at each operand's *own* type rather than at the provider's, because a contract may declare an operand of a fixed type the interface never parameterizes (the seeded-hash shape #34 wants); the operator path never has to represent that. The emission core is shared with `lowerInterfaceWitnessCall`, so the operator path emits exactly what it emitted before.
- **`Instances.ts`** — `interfaceWitnessTargets` walks `BoundOperationCall` alongside operator-spelled `BuiltinCall`, so the witness specialization exists to be called (requirement 2).

## 2. Make silent invalid MIR impossible here

Lowering has no diagnostic channel — an unlowerable body becomes `trapFunction`, whose all-`i32` signature no longer matches its callers, which is why the failure surfaced as `InvalidCallShape` with nothing naming a cause.

`Instances.unlowerableWitnessViolations` walks every reachable bound-operation call and resolves what lowering will resolve: a builtin-rule intrinsic, or a function of the provider's own actor. A conformance selecting neither is now a reported error instead of a dropped call. It runs in `Pipeline.realize` and `Pipeline.prepare` next to the other instance-level violation checks, where the substitution is known and diagnostics are already collected.

Scoped to the *non-operator* spelling on purpose: for an operator, "no witness" is the normal scalar answer — the width-neutral builtin lowering is correct there — so reporting it would break every intrinsic conformance.

**Diagnostic code taken: `SEM0101`** (`unlowerableBoundWitnessCode`). The highest on `main` is `SEM0097`; `SEM0098` is held by PR #119 and `SEM0099`/`SEM0100` by PR #153, so I took the next code after all three claims rather than the next after `main` — per #152, a gap is harmless and a collision is not. `diagnostics.md` was regenerated with `documentation:generate` (not hand-edited) and `documentation:check` passes.

No hash operation was added anywhere — orthogonal to #34's requirement 9.

## Tests

New `packages/compiler/test/BoundOperationWitness.test.ts` — seven cases, asserting **values returned**, not that compilation completed:

- a non-operator bound call over a user-defined provider returns `42` on the **bootstrap evaluator, the direct WebAssembly backend, and native LLVM**. The digest is computed by the witness (`20 + 22`), so a wrong witness or a placeholder result cannot produce it by accident;
- two providers with two unrelated source witnesses answer one call site differently (`10` and `32`);
- one bound operation selects an intrinsic witness for `i32` and a source witness for a struct in the same body — the intrinsic half still wraps at `i32::MIN`;
- an operand of a fixed non-parameter type (a seed) is borrowed at its own type;
- both spellings over one conformance in one body: `==` still reaches `equals`, the bound's name reaches `digest`;
- an unlowerable witness produces `SEM0101` at the call span rather than invalid MIR;
- neither witness kind a conformance may legitimately name produces that diagnostic.

Full `packages/compiler` suite: **171 files / 1376 tests, all passing.**

### A note on how the unlowerable case is driven

The conformance checker admits exactly the two witness kinds lowering knows, so no source the frontend accepts reaches the unlowerable state today — I probed for one (parametric impls, callable values carrying a bounded generic, nested generic forwarding, effect and match contexts) and every route is already rejected by an earlier diagnostic. That agreement between checker and lowering *is* the invariant, and it is precisely what broke when witness admissibility widened in #142. So the check is driven against an index in which the named witness is absent — the shape any future disagreement takes — using only the public `Analysis.instancesOf` / `Analysis.declarationIndex` accessors.

## On `HashKeyBoundForm.test.ts` (PR #149)

Its seventh case (`does not yet reach a source witness through the bound-operation call`) is a **tripwire that asserts the broken behavior** — `Blocked` / `InvalidMir` / `['InvalidCallShape']`. Its own docstring says so: *"Asserted rather than skipped. When the gap closes this test fails, which is the signal that `HashMap` over a user-defined key can be written."*

Run against this branch, the other six cases pass and that one **fails** — it now evaluates to `Completed` with value `42`, the correct outcome. It flips to red, not green, by construction. The file lives on `agent/34-hash-collections`, so PR #149 owns inverting it; the replacement assertion is in my comment on #155. I did not touch that file.


julia-script:
## Review — this is the standard the rest of the run should be held to

Three things in particular:

**The tests assert values, not completion.** A digest of `20 + 22` computed *by the witness*, checked on the bootstrap evaluator, the direct WebAssembly backend, and native LLVM. A placeholder result or a wrong witness cannot produce `42` by accident. That is the discipline whose absence let #86, #114 and #130 ship — three miscompiles this month invisible to tests that only asserted compilation finished.

**You took `SEM0101` past all three outstanding claims rather than the next code after `main`.** Exactly right, and the reasoning is the one to keep: a gap in the sequence costs nothing, a collision costs a silent overwrite. Worth knowing that I reproduced that overwrite end-to-end an hour ago ([#152](https://github.com/julia-script/silk-effect/issues/152#issuecomment-5285417146)) — the generator does not emit two rows as I had originally claimed, it emits **one**, and the loser's meaning vanishes from `diagnostics.md` while the compiler still emits the code. Your choice avoids exactly that.

**You did not touch `HashKeyBoundForm.test.ts`.** Correctly diagnosed as a tripwire asserting the broken behavior, correctly predicted it flips to red rather than green, and correctly left it to PR #149 to invert. Reaching into another lane's branch to "fix" a failing test would have destroyed the signal that lane deliberately built.

### On the unlowerable case — your qualification is the right call

You state plainly that no source the frontend accepts reaches the unlowerable state today, so `SEM0101` is driven against a synthesized index rather than a real program. That is worth being precise about, and you were: it means the diagnostic is **a guard against future checker/lowering divergence, not a user-facing error today**.

That is the correct thing to build, and your justification is the load-bearing one — *the agreement between checker and lowering is the invariant, and it is precisely what broke when admissibility widened in #142.* A test that cannot be driven from source still guards the thing that actually failed. Probing four routes (parametric impls, callable values carrying a bounded generic, nested generic forwarding, effect and match contexts) before concluding that is the difference between a claim and a finding.

Scoping the check to the non-operator spelling — because "no witness" is the normal, correct answer for an operator over a scalar — is the detail that would have made this fix break every intrinsic conformance if missed.

### Merge mechanics — I will handle this

`mergeable_state` is `dirty`. It is **not** a code collision: your branch predates #119 landing `SEM0098`, so both insert immediately after `SEM0097` and git flags the adjacency. Resolution is "keep both" plus a regeneration of `diagnostics.md`, which is the same shape I resolved for #119. I will do it at merge time along with the #153 renumber, so the three diagnostic-adding PRs land in a consistent order — you do not need to rebase.

Holding only on CI. Once it is green this merges and unblocks #34.

One small thing for next time: the PR body's footer should be the repo's standard form —

```
---
```

---

julia-script:
Verification pass complete.

**`@silk-effect/compiler` — green end to end.** `pnpm test` (its real script: `stdlib:check` → `documentation:check` → 170 parallel files → native acceptance): **1375 + 1 tests, 0 failures.**

**Everything else — green.** `turbo run typecheck test` across `language`, `llvm`, `lsp`, `docs`, `documentation`, `platform-webcontainer`: 24/24 tasks, 0 failures.

**The two pre-existing failures, verified against `origin/main` in a clean worktree rather than assumed:**

| Failure | On `origin/main` (90e292c) |
|---|---|
| `@silk-effect/wasm` `Extended.test.ts > threads address types through memory instructions` | fails identically |
| `compiler-cli` — `reports storage failures…`, `reports an atomic commit failure…`, `maps read and write failures…` | all three fail identically |

Neither is touched by this branch: the changes are bound-operation lowering, instance discovery, and one new semantic diagnostic.

Worth noting for whoever reads a red `pnpm check` here — turbo cancels sibling tasks the moment `@silk-effect/wasm#test` fails, so `compiler`, `llvm`, `lsp`, `docs` and `documentation` print `ELIFECYCLE Test failed` in that run without having failed anything. `llvm` had printed all-green test files immediately before its "failure" line. The suites above were re-run individually for that reason.

---

Issue #147, opened 2026-08-13
tooling: BuildExeCommand test runs a real native build under vitest's default 5s timeout, and times out on CI


julia-script:
## Evidence

`main` went red at `a151d94` (the merge of PR #142) — [run 31728817591, job `validate`](https://github.com/julia-script/silk-effect/actions/runs/31728817591/job/94543816010):

```
FAIL test/BuildExeCommand.test.ts > compiles a single file to a runnable executable and exits zero
Error: Test timed out in 5000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
 ❯ test/BuildExeCommand.test.ts:37:4

 Test Files  1 failed | 15 passed (16)
      Tests  1 failed | 91 passed (92)
```

It was the only failure. **The very next commit, `b20c3a2`, passed with no change to that file** — so it is intermittent, not a real break. `main` is green now.

## The cause is visible in the same log

That test compiles a single Silk file to a runnable native executable — it shells out to the toolchain and links. It is running under vitest's **default 5000ms** timeout, and nothing in the file raises it.

Meanwhile, in the same package and the same run:

| test | elapsed |
|---|---|
| `Workflow.test.ts > builds to the deterministic target and profile path` | **13,572 ms** |
| `Workflow.test.ts` (whole file, 15 tests) | 16,542 ms |
| `BuildExeCommand.test.ts` | timed out at 5,000 ms |

So native build work in this package routinely runs **2-3x the timeout that `BuildExeCommand` is held to**. The 5s budget is not tight — it is below the normal cost of the operation being tested, and it passes locally only because a developer machine is faster and unloaded.

## Why this is a *different* fix from #143

Worth stating explicitly, because a worker reading both issues could apply the wrong rule.

**#143 must not be fixed by raising its budget.** Its wall-clock number is standing in for a real property — that occurrence lookup is sub-linear — so a raised budget is a gate that no longer catches a regression.

**This one has no such property.** Nobody chose 5000ms to assert anything about build performance; it is vitest's default, inherited by a test whose subject is *"does this produce a runnable executable and exit zero"*. The timeout is not guarding build speed, it is truncating a correctness test. Raising it — or setting an explicit generous timeout for every test in this package that invokes the native toolchain — is the correct fix here, not a cop-out.

If build *performance* is worth guarding, that is a separate, deliberate test, and it should say so.

## Also: this is not the failure reported on #24

[...]

PR #163, opened 2026-08-13
test: make `pnpm check` pass on a clean checkout as root and on Node 22

### LINKED ISSUE #159: tooling: FormatWorkflow permission tests fail as root, so the suite is red in any root container
## Current behavior

`packages/compiler-cli/test/FormatWorkflow.test.ts` fails on a clean checkout when the suite runs as **root**:

```
FAIL  test/FormatWorkflow.test.ts > reports storage failures, continues in order, and gives exit class two precedence
FAIL  test/FormatWorkflow.test.ts > reports an atomic commit failure without replacing the original file

AssertionError: expected [ 'Changed' ] to deeply equal [ 'Failed' ]
- Expected
+ Received
  [
-   "Failed",
+   "Changed",
  ]
```

Reproduced on a clean `main` checkout with no local changes, `id -u` = 0.

## Cause

The tests induce failure with file permissions:

```
FormatWorkflow.test.ts:181   fileSystem.chmod(`${root}/src/A-Unreadable.silk`, 0o000)
FormatWorkflow.test.ts:202   fileSystem.chmod(`${root}/src`, 0o555)
```

**Root bypasses file permission checks.** `0o000` still reads and `0o555` still accepts a new file, so the operation the test expects to fail succeeds, and the outcome is `Changed` where the assertion wants `Failed`. There is nothing wrong with the product code — the tests simply cannot express their precondition as root.

CI is unaffected because GitHub runners execute as the non-root `runner` user, which is why this has never turned `main` red.

## Why it is worth fixing

It is a standing tax on anyone whose environment runs as root — containers, cloud dev environments, most Docker images. Every such run shows 2–3 red tests on a clean tree, and each person has to independently determine they are pre-existing and unrelated to their change.

That has already happened repeatedly in this backlog. Recent examples:

- PR #149's author reported "`pnpm check` fails in `@silk-effect/compiler-cli` (3 format-failure tests) — **both predate this branch**: verified by checking out `origin/main` at `b20c3a2` and reproducing each failure identically there." That verification was correct and also entirely wasted work.
- The #135 design pass reported `pnpm check` blocked by a *different* `compiler-cli` failure, and the two got conflated in later discussion.

A red baseline is corrosive in exactly the way #102 and #143 describe: it trains people to skim failures rather than read them, and it hides the real one when it arrives. This is the same family as #147 — an environmental assumption a test depends on but never states.

## Requirements

1. The suite passes on a clean checkout as root **and** as an unprivileged user.
2. The permission-failure paths remain covered when the environment can express them. Deleting the assertions is not acceptable — they cover a real failure mode.
3. When the environment cannot express the precondition, the test **says so** rather than failing or silently passing. A skip with a printed reason is honest; a green that asserted nothing is not.

## Possible approaches

- **Skip with a reason under root.** Detect `process.getuid?.() === 0` and skip those specific cases with a message naming why. Simplest, and consistent with requirement 3.
- **Induce the failure without permissions.** Point the write at a path that fails structurally for every user — a directory where a file is expected, a nonexistent parent, or an injected filesystem error at the `FileSystem` layer. Strictly better where it works, because coverage is preserved everywhere.

The second is preferable if the `FileSystem` seam allows injecting an error; the first is an acceptable fallback for cases where only a real permission denial will do.

## Acceptance criteria

- [ ] `vitest run test/FormatWorkflow.test.ts` passes as root on a clean checkout.
- [ ] It still passes as a non-root user, with the permission cases actually exercised there — verified by running both and pasting the results.
- [ ] Any case that skips prints why it skipped.
- [ ] No assertion is deleted or weakened.

## Related

- #147 — `compiler-cli` timeouts; same package, same class of unstated environmental assumption.
- #143, #102 — the standing argument that a red or flaky baseline on `validate` trains people to stop reading failures.

---

Filed by orchestrator run `run-r7x1` while verifying PR #154's stability claim; the failures turned out to be unrelated to that PR.

### LINKED ISSUE #161: tooling: `pnpm check` cannot pass on Node 22, but `engines` claims `>=22.13.0`
Found while verifying #147 on a clean `origin/main` worktree. Not a regression from anything recent — it reproduces on `main` with no local changes.

## The mismatch

- Root `package.json` declares `"engines": { "node": ">=22.13.0" }`
- `.github/workflows/ci.yml` pins `node-version: 24`

**On Node 22, `pnpm check` cannot pass.** CI never sees it because CI is on 24, so the declared support range is wider than the range the test suite actually runs on. Anyone who installs a supported Node and runs the documented gate gets a red build they did not cause.

## Verified bisect

`packages/wasm/test/Extended.test.ts › threads address types through memory instructions`

| runtime | V8 | result |
|---|---|---|
| Node **22.22.2** | 12.4.254.21 | `Test Files 1 failed` · `Tests 1 failed \| 7 passed (8)` |
| Node **24.19.0** | 13.6.233.17 | `Test Files 1 passed` · `Tests 8 passed (8)` |

Same commit, same worktree, only the runtime changed. The assertion that fails is `Extended.test.ts:184`:

```ts
const bytes = yield* Binary.encode(builder)
assert.isTrue(WebAssembly.validate(bytes))   // false on V8 12.4
```

## Narrowed to the exact construct

My first guess — "V8 12.4 lacks memory64" — was **wrong**, so I checked instead of filing it. A bare i64-address memory validates fine on 12.4:

```
memory64 memory declaration        validate=true
i64.load offset 0                  validate=true
i64.load offset 2^32               validate=true
memory.size on i64 memory          validate=true
mixed 64/32 memory.copy            validate=FALSE   ← the only rejection
```

Isolating `memory.copy` across all four address-type combinations, printing the encoded immediates alongside the verdict:

```
i32 <- i32           validate=true   copyimm=[fc 0a 00 01 0b]
i64 <- i64           validate=true   copyimm=[fc 0a 00 01 0b]
i64 <- i32 (mixed)   validate=false  copyimm=[fc 0a 00 01 0b]
i32 <- i64 (mixed)   validate=false  copyimm=[fc 0a 00 01 0b]
```

**The encoder is not at fault.** The emitted bytes are byte-identical in all four cases — `fc 0a` then dst memidx `00`, then src memidx `01`, which is the spec order. Only the memory *types* in the memory section differ, and only the mixed pairs are rejected, symmetrically in both directions. That is consistent with V8 12.4 requiring both memories to share an address type for `memory.copy`, and 13.6 implementing the adopted memory64 rule (dst index takes dst's type, src index takes src's type, count takes the narrower of the two — which is what the test builds).

So this is a V8 capability gap, not a bug in `packages/wasm`. **No production code needs changing.**

## Options

Ordered by how I'd weigh them, but this is a maintainer's call:

1. **Narrow `engines` to `>=24`** — smallest change, and makes the declared range honest. Costs Node 22 support, which may or may not be wanted.
2. **Gate the mixed-copy assertion on a runtime capability probe** — validate a tiny mixed-copy module once and skip that one assertion when the host V8 rejects it. Keeps Node 22 usable for contributors without weakening the assertion where it can run. A bare `WebAssembly.validate` probe on a 13-byte module is enough.
3. **Leave as-is and document it** — cheapest, but leaves `pnpm check` red-by-default for anyone on a version the manifest says is supported.

I'd avoid simply deleting the mixed-copy case; it is testing real, correct behaviour that a current V8 accepts.

## Reproduction

```bash
nvm install 22 && nvm use 22
pnpm install --frozen-lockfile && pnpm build
cd packages/wasm && npx vitest run test/Extended.test.ts   # 1 failed | 7 passed
nvm use 24
npx vitest run test/Extended.test.ts                       # 8 passed
```

Note for anyone reproducing the *full* gate on a container that runs as root: three permission-injection tests in `compiler-cli` (`FormatCommand` › maps read and write failures…, `FormatWorkflow` › reports storage failures…, `FormatWorkflow` › reports an atomic commit failure…) also false-fail there, because root ignores `chmod 0o000`. Drop `cap_dac_override`/`cap_dac_read_search` (e.g. via `capsh`) or run as a normal user. That is unrelated to this issue and unrelated to the repo.

## Related

- #147 — where this surfaced; its PR (#154) is unaffected and passes 92/92 on both Node 22 and Node 24.

julia-script:
Closes #159
Closes #161

Two causes of the same symptom — a red baseline on a clean checkout — kept as two commits. Fixing either alone leaves the tax in place, so they travel together.

No production code changes. No assertion deleted or weakened.

## #159 — permission tests fail as root

`FormatWorkflow.test.ts` and one case in `FormatCommand.test.ts` established their failure preconditions with `chmod(path, 0o000)` and `chmod(directory, 0o555)`. Root bypasses the file permission check, so the read and the rename both succeed, the outcome is `Changed` where the assertion wants `Failed`, and three tests fail. CI never saw it because GitHub runners are non-root.

**Approach 1 (induce the failure without permissions) worked for both tests, so no skip was needed anywhere.** The failure is injected at the `FileSystem` seam by `test/fileSystemFailures.ts`, which wraps the platform filesystem so a named operation on a named path fails with the same `PermissionDenied` `PlatformError` the platform layer produces when the OS denies it for real:

| test | old precondition | new precondition |
|---|---|---|
| `FormatWorkflow` › reports storage failures… | `chmod 0o000` on the source | `readFile` denied on that path |
| `FormatWorkflow` › reports an atomic commit failure… | `chmod 0o555` on `src/` | `rename` denied on the destination — the atomic swap itself |
| `FormatCommand` › maps read and write failures… | both of the above | both of the above |

Only those two operations are overridden, and only for paths the caller names — the reads each test makes to assert a file was left untouched still go to the real filesystem. Coverage is now identical under every UID instead of silently better on CI. If a refactor moved the workflow off `readFile`/`rename`, the injection would stop firing and the tests would fail rather than pass quietly.

## #161 — `engines` promises Node 22, the suite could not pass on it

I did not re-derive the diagnosis; #161's is complete and correct. I did re-confirm it with a hand-written module independent of the encoder, which reproduces the reported table exactly — including that the first hypothesis (missing memory64) is wrong:

```
node v22.22.2  v8 12.4.254.21
i32<-i32         true
i64<-i64         true
i64<-i32 mixed   false
i32<-i64 mixed   false
```

Taking **option 2** — a runtime capability probe — rather than narrowing `engines`, which is a support-policy call for the maintainer to make deliberately rather than as a side effect of one V8 gap. `engines` is unchanged.

`test/support/hostCapability.ts` validates a 41-byte hand-written mixed-copy module once. It is hand-written rather than built through the builder on purpose: a probe routed through the encoder under test would make an encoder regression look exactly like a missing host feature and would disable the assertion in the one situation it exists to catch. Verified the probe module is genuinely valid — `validate: false` on Node 22 / V8 12.4, `true` on Node 24 / V8 13.6 — so the guard cannot be permanently stuck off.

The host validation moved into its own test so the skip is a real skip. Everything else in the original test — the address-type rejection, the encode — still runs unconditionally on every host.

### On "prints why it skipped"

`console.log` is not enough, and this is measurable rather than a matter of taste. The default reporter this repo runs prints neither log output nor test annotations:

```
$ vitest run test/diag.test.ts               # console.log in the test body
 Test Files  1 passed (1)
      Tests  1 passed (1)                    # ← the log appears nowhere
```

That is exactly how #151's `opt` cross-check went its whole life without running. The same pattern is still live at `packages/compiler/test/ModuleVerification.test.ts:55` — worth a follow-up, out of scope here.

So the reason goes into the skipped test's **name**, which every reporter that lists a test at all will show, and the skip itself raises a non-zero skipped count in the summary — the run reports that something did not happen:

```
 ✓ threads address types through memory instructions
 ↓ the host validates a module threading address types through memory instructions — skipped:
   this host's WebAssembly rejects a mixed 64/32-address memory.copy, so it cannot serve as the
   validity oracle for a module containing one; the encoding is unchecked here. Node 24 accepts
   it (see issue #161)

      Tests  8 passed | 1 skipped (9)
```

Under CI a missing capability **fails instead of skipping** — CI pins a runtime that has it, so its absence means the pinned runtime changed and the assertion has quietly stopped running:

```
$ CI=1 vitest run test/Extended.test.ts      # on Node 22
AssertionError: the host validates a module threading address types through memory instructions:
this host's WebAssembly rejects a mixed 64/32-address memory.copy … CI pins a runtime that
supports it, so this is a change in the pinned runtime rather than an environment to skip.
      Tests  1 failed | 8 passed (9)
```

The condition is a probed capability, not a version string or an environment guess.

## Verification

Full `pnpm check`, both UIDs × both Node versions, `TURBO_FORCE=true` throughout so nothing replays from cache. **Both issues in this lane are fixed in all four cells** — `compiler-cli` 92/92 and `wasm` 9/9 files in every run:

```
every run:  @silk-effect/compiler-cli:test:  Tests  92 passed (92)          ← #159, was 3 failed as root
Node 24:    @silk-effect/wasm:test:          Tests  83 passed (83)          ← assertion runs
Node 22:    @silk-effect/wasm:test:          Tests  82 passed | 1 skipped   ← #161, was 1 failed
```

`pnpm check` overall, however, is still not green on Node 22 — for a **third, unrelated** reason:

| | root (uid 0) | non-root (uid 1000) |
|---|---|---|
| **Node 24.19.0** | `EXIT=0` — 30/30 tasks | `EXIT=0` — 30/30 tasks |
| **Node 22.22.2** | `EXIT=1` — `@silk-effect/compiler#test` | `EXIT=1` — `@silk-effect/compiler#test` |

`packages/compiler/test/LexerPressure.test.ts › rolls back every token and diagnostic allocation failure on all three engines` exceeds its `180_000` ms budget on Node 22 (180 028 ms / 180 031 ms) while taking 139–150 s on Node 24. `git diff origin/main --stat -- packages/compiler packages/lsp` is empty — this branch does not touch that package.

**[The verification comment](https://github.com/julia-script/silk-effect/pull/163#issuecomment-5287350304) has the full matrix and three further pre-existing baseline problems I found while running it**, with measurements: the `LexerPressure` budget above, an `lsp` test running against vitest's default 5 s timeout, and a `NativeToolchain` test that promotes to the fixed shared path `/tmp/promoted.o` and collides across users. None are fixed here — reported rather than silently absorbed, and happy to take any of them in a follow-up.

### DIFF
--- a/packages/compiler-cli/test/FormatCommand.test.ts
+++ b/packages/compiler-cli/test/FormatCommand.test.ts
@@ -5,6 +5,7 @@ import * as FileSystem from 'effect/FileSystem'
 import * as Result from 'effect/Result'
 import { Command } from 'effect/unstable/cli'
 import * as Cli from '../src/Cli.js'
+import { denying } from './fileSystemFailures.js'
 
 const compact = 'pub fn main() -> i32 { return 42 }'
 const canonical = 'pub fn main() -> i32 {\n  return 42\n}\n'
@@ -100,18 +101,26 @@ it.effect('maps read and write failures to exit class two without corrupting sou
     const readRoot = yield* fileSystem.makeTempDirectoryScoped()
     yield* makeProject(readRoot, canonical)
     yield* fileSystem.writeFileString(`${readRoot}/src/A-Unreadable.silk`, compact)
-    const readFailure = yield* Effect.acquireUseRelease(
-      fileSystem.chmod(`${readRoot}/src/A-Unreadable.silk`, 0o000),
-      () => execute(['format', '--manifest-path', `${readRoot}/silk.toml`]),
-      () => fileSystem.chmod(`${readRoot}/src/A-Unreadable.silk`, 0o644),
+    const canonicalReadRoot = yield* fileSystem.realPath(readRoot)
+    const readFailure = yield* execute(['format', '--manifest-path', `${readRoot}/silk.toml`]).pipe(
+      Effect.provideService(
+        FileSystem.FileSystem,
+        denying(fileSystem, { readFile: [`${canonicalReadRoot}/src/A-Unreadable.silk`] }),
+      ),
     )
 
     const writeRoot = yield* fileSystem.makeTempDirectoryScoped()
     yield* makeProject(writeRoot)
-    const writeFailure = yield* Effect.acquireUseRelease(
-      fileSystem.chmod(`${writeRoot}/src`, 0o555),
-      () => execute(['format', '--manifest-path', `${writeRoot}/silk.toml`]),
-      () => fileSystem.chmod(`${writeRoot}/src`, 0o755),
+    const canonicalWriteRoot = yield* fileSystem.realPath(writeRoot)
+    const writeFailure = yield* execute([
+      'format',
+      '--manifest-path',
+      `${writeRoot}/silk.toml`,
+    ]).pipe(
+      Effect.provideService(
+        FileSystem.FileSystem,
+        denying(fileSystem, { rename: [`${canonicalWriteRoot}/src/Main.silk`] }),
+      ),
     )
 
     assert.strictEqual(status(readFailure), 2)
--- a/packages/compiler-cli/test/FormatWorkflow.test.ts
+++ b/packages/compiler-cli/test/FormatWorkflow.test.ts
@@ -4,6 +4,7 @@ import * as Effect from 'effect/Effect'
 import * as FileSystem from 'effect/FileSystem'
 import * as Result from 'effect/Result'
 import * as FormatWorkflow from '../src/FormatWorkflow.js'
+import { denying } from './fileSystemFailures.js'
 
 const canonical = 'pub fn main() -> i32 {\n  return 42\n}\n'
 const compact = 'pub fn main() -> i32 { return 42 }'
@@ -176,11 +177,13 @@ it.effect('reports storage failures, continues in order, and gives exit class tw
     yield* makeProject(root, canonical)
     yield* writeFile(`${root}/src/A-Unreadable.silk`, compact)
     yield* writeFile(`${root}/src/Zeta.silk`, compact)
+    const canonicalRoot = yield* fileSystem.realPath(root)
 
-    const summary = yield* Effect.acquireUseRelease(
-      fileSystem.chmod(`${root}/src/A-Unreadable.silk`, 0o000),
-      () => FormatWorkflow.run({ workingDirectory: root }),
-      () => fileSystem.chmod(`${root}/src/A-Unreadable.silk`, 0o644),
+    const summary = yield* FormatWorkflow.run({ workingDirectory: root }).pipe(
+      Effect.provideService(
+        FileSystem.FileSystem,
+        denying(fileSystem, { readFile: [`${canonicalRoot}/src/A-Unreadable.silk`] }),
+      ),
     )
 
     assert.deepEqual(
@@ -197,11 +200,13 @@ it.effect('reports an atomic commit failure without replacing the original file'
     const fileSystem = yield* FileSystem.FileSystem
     const root = yield* fileSystem.makeTempDirectoryScoped()
     yield* makeProject(root, compact)
+    const canonicalRoot = yield* fileSystem.realPath(root)
 
-    const summary = yield* Effect.acquireUseRelease(
-      fileSystem.chmod(`${root}/src`, 0o555),
-      () => FormatWorkflow.run({ workingDirectory: root }),
-      () => fileSystem.chmod(`${root}/src`, 0o755),
+    const summary = yield* FormatWorkflow.run({ workingDirectory: root }).pipe(
+      Effect.provideService(
+        FileSystem.FileSystem,
+        denying(fileSystem, { rename: [`${canonicalRoot}/src/Main.silk`] }),
+      ),
     )
 
     assert.deepEqual(
--- a/packages/compiler-cli/test/fileSystemFailures.ts
+++ b/packages/compiler-cli/test/fileSystemFailures.ts
@@ -1,0 +1,62 @@
+/**
+ * Deterministic storage failures for the format tests.
+ *
+ * The failure paths these tests cover — a source file that cannot be read, an atomic commit that
+ * cannot be completed — used to be induced with `chmod(path, 0o000)` and `chmod(directory, 0o555)`.
+ * That works for an unprivileged user and silently does nothing for root, which bypasses the file
+ * permission check entirely: the read succeeds, the rename succeeds, and the assertion that wanted
+ * `Failed` sees `Changed`. Issue #159 records the cost — a clean checkout is red in any root
+ * container, and every contributor working in one pays to rediscover that the red is not theirs.
+ *
+ * Injecting the failure at the `FileSystem` seam removes the dependency on who is running the
+ * suite. The workflow still takes the same branch it takes when the operating system denies the
+ * operation for real, because it reaches that branch through the same `PlatformError` the platform
+ * layer would have produced. The precondition is now expressible under every UID rather than only
+ * under some, so the coverage is the same everywhere instead of silently better on CI.
+ *
+ * Only the two operations the failure paths depend on are overridden, and only for paths named by
+ * the caller. Every other call — including the reads a test makes to assert that a file was left
+ * untouched — goes to the real filesystem.
+ */
+import * as Effect from 'effect/Effect'
+import type * as FileSystem from 'effect/FileSystem'
+import * as PlatformError from 'effect/PlatformError'
+
+export interface Denials {
+  /** Canonical paths whose `readFile` fails, standing in for a source file the process cannot read. */
+  readonly readFile?: ReadonlyArray<string>
+  /** Canonical destination paths whose `rename` fails, standing in for an atomic commit that cannot land. */
+  readonly rename?: ReadonlyArray<string>
+}
+
+const denied = (method: string, syscall: string, path: string): PlatformError.PlatformError =>
+  PlatformError.systemError({
+    _tag: 'PermissionDenied',
+    module: 'FileSystem',
+    method,
+    syscall,
+    pathOrDescriptor: path,
+    description: `Injected permission denial for ${path}`,
+  })
+
+/**
+ * Wraps a filesystem so the named operations fail with `PermissionDenied` on the named paths.
+ *
+ * Paths are compared exactly, so callers pass canonical paths: the workflow resolves its selection
+ * through `realPath` before touching anything, and a temporary directory is a symlink on some
+ * platforms.
+ */
+export const denying = (
+  fileSystem: FileSystem.FileSystem,
+  denials: Denials,
+): FileSystem.FileSystem => ({
+  ...fileSystem,
+  readFile: (path) =>
+    denials.readFile?.includes(path)
+      ? Effect.fail(denied('readFile', 'open', path))
+      : fileSystem.readFile(path),
+  rename: (oldPath, newPath) =>
+    denials.rename?.includes(newPath)
+      ? Effect.fail(denied('rename', 'rename', newPath))
+      : fileSystem.rename(oldPath, newPath),
+})
--- a/packages/wasm/test/Extended.test.ts
+++ b/packages/wasm/test/Extended.test.ts
@@ -12,6 +12,7 @@ import * as Table from '../src/Table.js'
 import * as Type from '../src/Type.js'
 import * as ValType from '../src/ValType.js'
 import type { WasmError } from '../src/WasmError.js'
+import { mixedMemoryCopy, requiring } from './support/hostCapability.js'
 
 const failure = <A>(effect: Effect.Effect<A, WasmError>) =>
   effect.pipe(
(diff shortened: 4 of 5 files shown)


julia-script:
## Review — and you corrected my instruction, which is the most valuable part

I asked for skips that "print why they skipped." **You measured that and showed it is not enough**, which I should have checked before asking:

```
$ vitest run test/diag.test.ts               # console.log in the test body
 Test Files  1 passed (1)
      Tests  1 passed (1)                    # ← the log appears nowhere
```

The default reporter this repo runs prints neither log output nor annotations. So "skip with a printed reason" produces a green run that says nothing — which is **exactly how #151's `opt` cross-check went its entire life without running**, and I had just spent an hour insisting on hard-failing that same check without noticing the local half of the fix was equally invisible.

Putting the reason in the **test name** is the right answer: every reporter that lists a test at all shows it, and the skipped count in the summary is non-zero, so the run reports that something did not happen. That is a strictly better mechanism than the one I specified.

**And you found the live instance**: `packages/compiler/test/ModuleVerification.test.ts:55` still has the `console.log` pattern — code I merged an hour ago in #151. I have passed this to PR #150, which is already refactoring that file into a shared `llvmToolchain` helper, so the fix lands where the code is moving rather than colliding with it. Correctly scoped out of your PR.

### The rest

**#159 — you took approach 1 for both tests, so nothing skips.** Injecting `PermissionDenied` at the `FileSystem` seam rather than skipping under root means coverage is now *identical under every UID* instead of silently better on CI. And the property you called out is the one that makes it safe:

> If a refactor moved the workflow off `readFile`/`rename`, the injection would stop firing and the tests would fail rather than pass quietly.

Overriding only the two named operations on caller-named paths — and letting the assertion reads still hit the real filesystem — is what keeps that true.

**#161 — the probe is hand-written on purpose**, and the reasoning is the load-bearing bit:

> a probe routed through the encoder under test would make an encoder regression look exactly like a missing host feature and would disable the assertion in the one situation it exists to catch

That is the trap in every capability-probe design and you avoided it. Verifying the 41-byte module is genuinely valid on both hosts — `false` on V8 12.4, `true` on 13.6 — closes the other half: the guard cannot be permanently stuck off.

Leaving `engines` alone was right. Narrowing to `>=24` is a support-policy decision for Julia to make deliberately, not a side effect of one V8 gap.

**Failing rather than skipping under CI** is the correct inversion: CI pins a runtime that has the capability, so its absence means the pinned runtime changed and the assertion has silently stopped running.

### Outstanding

Only the verification matrix — root × non-root × Node 22 × Node 24. Post it and I will merge once `validate` is green.

---

julia-script:
## Your approach is confirmed — no rework, keep going

A maintainer definition decision landed on #161 at 21:31, **after** you were dispatched, and its first line pinned runtime support at Node `>=24.0.0` — which would have invalidated the capability probe and required narrowing `engines` instead.

Julia has since overridden that line directly: **Node 22 stays supported.** So option 2 — the runtime capability probe with `engines` untouched — is the right answer, and **PR #163 as built already implements it.** Nothing to change.

The rest of the 21:31 decision stands and you are consistent with it: the pinned `wasm-tools` validator remains the correctness oracle, and you kept the encode and address-type-rejection assertions unconditional, gating only the host-validation-as-oracle step.

Full detail on [#161](https://github.com/julia-script/silk-effect/issues/161#issuecomment-5286664964), including one thread deliberately **not** folded into this PR: `Extended.test.ts` asserts on the *host's* `WebAssembly.validate`, which is why a V8 gap can look like an encoder defect at all. Making `wasm-tools` the oracle for that test would remove the host-version sensitivity entirely — a more durable fix, its own change, and not yours tonight.

Still waiting only on your root × non-root × Node 22 × Node 24 matrix, and on `validate`.

---

julia-script:
## Verified independently on the axis that was broken — merging

Your matrix had not posted yet, and this container happens to be the exact failing configuration — **root (uid 0) on Node v22.22.2** — so I ran that half myself rather than merge on the strength of the design. All four properties hold.

**1. #159 — root, Node 22**

```
$ npx vitest run test/FormatWorkflow.test.ts test/FormatCommand.test.ts
 Test Files  2 passed (2)
      Tests  12 passed (12)
```

Three of these failed on `main` before this branch. Nothing skips — the `FileSystem` injection works under root exactly as it does under any other UID, which was the point of taking approach 1.

**2. #161 — the skip happens and is counted**

```
$ npx vitest run test/Extended.test.ts
      Tests  8 passed | 1 skipped (9)
```

**3. The reason is visible, and it is genuinely informative**

```
 ↓ the host validates a module threading address types through memory instructions — skipped:
   this host's WebAssembly rejects a mixed 64/32-address memory.copy, so it cannot serve as the
   validity oracle for a module containing one; the encoding is unchecked here. Node 24 accepts
   it (see issue #161)
```

It names the construct, the consequence (*the encoding is unchecked here* — so the reader knows what stopped being verified, not merely that something was skipped), and where to read more. That is the standard the rest of the repo should follow, and #165 now tracks bringing `llvmToolchain` onto it.

**4. `CI=1` fails rather than skips**

```
$ CI=1 npx vitest run test/Extended.test.ts
AssertionError: … CI pins a runtime that supports it, so this is a change in the pinned runtime
rather than an environment to skip.
      Tests  1 failed | 8 passed (9)
```

The message tells whoever hits it what actually happened — the pinned runtime changed — rather than leaving them to guess. That inversion is what makes the local skip safe to have at all.

### Coverage

Between this run and CI, the matrix's load-bearing cells are covered: **root × Node 22 by me**, **non-root × Node 24 by `validate`** (green). Your full four-cell matrix is still worth posting when it finishes — the two remaining cells should be uneventful, and if either is not, that is a finding worth its own comment even though this has merged.

Merging.

---

julia-script:
## Verification — full `pnpm check`, both UIDs × both Node versions

Every run with `TURBO_FORCE=true` so nothing is replayed from cache (`Cached: 0 cached, N total` in each). Non-root runs as `ubuntu` (uid 1000) with the tree chowned to it.

**The two issues in this PR's lane are fixed in all four cells.** In *every* run, on both Node versions and both UIDs:

```
@silk-effect/compiler-cli:test:  Test Files  16 passed (16)
@silk-effect/compiler-cli:test:       Tests  92 passed (92)      ← #159, was 3 failed as root
```

```
Node 24:  @silk-effect/wasm:test:  Tests  83 passed (83)                  ← assertion runs
Node 22:  @silk-effect/wasm:test:  Tests  82 passed | 1 skipped (83)      ← #161, was 1 failed
```

The Node 22 line is the mechanism working as designed: a visible skipped count, and the reason on the test itself.

### But `pnpm check` overall is still not green, and not because of #159 or #161

| | root (uid 0) | non-root (uid 1000) |
|---|---|---|
| **Node 24.19.0** | `EXIT=0` — 30/30 tasks | `EXIT=0` — 30/30 tasks |
| **Node 22.22.2** | `EXIT=1` — `@silk-effect/compiler#test` ¹ | `EXIT=1` — `@silk-effect/compiler#test` ¹ |

¹ Not this PR's packages. `git diff origin/main --stat -- packages/compiler packages/lsp` is **empty** — this branch touches only `packages/compiler-cli/test` and `packages/wasm/test`.

I found **three further pre-existing baseline problems** while running the matrix. All are out of this lane, none are touched by this PR, and I have not fixed any of them. Reporting rather than silently absorbing, since that is the whole point of the lane.

---

### Finding A — `LexerPressure` exceeds its 180 s budget on Node 22 (blocks Node 22 today)

`packages/compiler/test/LexerPressure.test.ts › rolls back every token and diagnostic allocation failure on all three engines`, budget `180_000` (`LexerPressure.test.ts:471`).

| runtime | uid | duration | result |
|---|---|---|---|
| Node 24.19.0 | 0 | 139 679 ms | pass (78 % of budget) |
| Node 24.19.0 | 1000 | 140 774 ms | pass |
| Node 24.19.0 | 1000 | 150 145 ms | pass (83 % of budget) |
| Node 22.22.2 | 0 | **180 028 ms** | `Error: Test timed out in 180000ms.` |
| Node 22.22.2 | 1000 | **180 031 ms** | `Error: Test timed out in 180000ms.` |

The very first root/Node 22 run of the matrix did pass end to end (`EXIT=0`), so it is marginal rather than certain — but it landed on the wrong side in 2 of 3 Node 22 runs, and never on Node 24.

**This means Node 22 still cannot reliably pass `pnpm check` after this PR.** #161 was a real cause and is fixed; it was not the only one. This is the #147 family — a budget sized on one runtime and never checked on the other end of the declared `engines` range. It deserves its own issue, and it is also evidence for whichever way the `engines` policy question is eventually decided.

### Finding B — `lsp` `Document.test.ts` runs against the default 5 s timeout

`packages/lsp/test/Document.test.ts › completes standard-library namespace and source service operations`. `packages/lsp` has no `vitest.config.ts`, so this inherits vitest's default `testTimeout` of 5 000 ms — unlike `packages/compiler`, which sets `testTimeout: 30_000`.

| context | duration | result |
|---|---|---|
| root / Node 24, full gate | 2 900 ms | pass |
| non-root / Node 22, full gate | **5 116 ms** | `Error: Test timed out in 5000ms.` |
| non-root / Node 22, file alone | — | pass, 55/55 |

It passes in isolation and fails under the parallel load of the full gate, at 58 % → 102 % of budget. Same family as Finding A.

### Finding C — `NativeToolchain` promotes to a fixed shared path in the system temp dir

`packages/compiler/test/NativeToolchain.test.ts:260` promotes to `join(scope.root, '..', 'promoted.o')`, which resolves to **`/tmp/promoted.o`** — a fixed path, not a scoped temp file, and never cleaned up (the test asserts it exists and leaves it).

I hit this for real: my root runs left `/tmp/promoted.o` owned by root, and the next non-root run failed with

```
Error: EACCES: permission denied, copyfile '/tmp/silk-cleanup-yW5CvO/program.o' -> '/tmp/promoted.o'
 ❯ Module.promote src/NativeToolchain.ts:275:3
```

Two users on one machine, two checkouts, or a reused CI runner with a different uid will collide the same way. Deleting the stale file made it pass. This one is a genuine test bug rather than a timing budget — the scope-owned intermediate is correctly scoped, but the *promotion target* deliberately escapes the scope to a name nobody owns.

---

Happy to open issues for A, B and C, or to take any of them in a follow-up branch — say the word. I have left them alone here so this PR stays exactly the two issues it closes.

---

PR #164, opened 2026-08-13
fix(compiler-cli): compile a settled source tree in watch mode

### LINKED ISSUE #158: Workflow.watch recompiles on the raw fs event, so a non-atomic write is compiled while truncated
Filed from #147, which asked whether the `packages/compiler-cli/test/Workflow.test.ts:383` watcher failure reported during the #135 design work (`passes[1]` is 1, expected 0) is genuinely stable. **It is not stable, and it is not a test-only problem.** Here is the reproduction.

## It does not reproduce on this machine at all

25 consecutive runs of `Workflow.test.ts` under 6× CPU oversubscription: **0 failures**. So the "stable, unrelated watcher failure" claim should not be repeated as-is — on this hardware the test is solidly green, exactly as it was in the CI run that produced #147 (15/15).

## But the race is real, and here is the mechanism

`Workflow.watch` recompiles on **every** raw filesystem event, with no debounce or settle. The source already flags this:

```ts
// ponytail: every event recompiles the whole graph; debounce if editors emit noisy bursts.
```

`fs.writeFile` is not atomic — it is `open(O_TRUNC)` → `write` → `close`. `fs.watch` fires on the truncate, so a reader woken by that event can read the file at zero length. An empty `Main.silk` has no `main` entry, so `Workflow.check` returns **1** — precisely the reported `passes[1]` is 1.

**Instrumenting the test** (recording every pass and the bytes it actually compiled) shows a single write produces **two** passes in 11 of 12 runs:

```
11×  [{status:1, phase:"before-edit", bytes:38}, {status:0, after-edit, bytes:34}, {status:0, after-edit, bytes:34}]
 1×  [{status:1, phase:"before-edit", bytes:38}, {status:0, after-edit, bytes:34}]
```

Two recompiles per write means two independent reads, each racing the writer.

**Measuring the window directly** — watch a directory and, on each event, read the file the way `check` does, while rewriting it non-atomically:

```js
const watcher = watch(dir, () => { readFile(file, 'utf8').then(t => tally(t.length)) })
for (let i = 0; i < 3000; i += 1) await writeFile(file, good)   // 34 bytes
```

```
{"writes":3000,"events":5149,"empty":3683,"short":0,"lengths":[[34,1466],[0,3683]]}
{"writes":3000,"events":5270,"empty":3607,"short":0,"lengths":[[34,1663],[0,3607]]}
{"writes":3000,"events":5202,"empty":3774,"short":0,"lengths":[[34,1428],[0,3774]]}
```

**~71% of watch-woken reads observe a zero-length file.** The window is not narrow; it is the common case when writes are close together. Full script: <https://github.com/julia-script/silk-effect/pull/154> discussion, and reproduced verbatim above.

## Why it looks "stable" to one person and invisible to another

Whether the woken reader wins the race against the truncate is decided by filesystem and `fs.watch` backend scheduling. In the test the writes are spaced by 50 ms polls so the reader usually wins here; on a different filesystem or watch backend the first event can reliably land inside the truncate window, making the failure deterministic on that machine and absent everywhere else. That is the signature of the report in #135, and it is why a "stable" failure green-builds on `main`.

## Why this is not just a test bug

It is tempting to make the test's helper write atomically (temp file + rename) — every assertion would then pass untouched. **I deliberately did not do that in #147's PR**, because it would hide a real product behaviour: anything that writes non-atomically into a watched tree — shell `>` redirection, `sed -i`, editors without atomic save — will make `silk` watch mode emit a spurious "no main entry" diagnostic during an ordinary save. The test is surfacing that, not inventing it.

## Suggested fix (needs a maintainer's call)

Both are product changes, which is why this is filed rather than folded into #147:

1. **Debounce/settle** the event stream in `Workflow.watch` before recompiling — the fix the existing `ponytail` comment already anticipates. Coalesces the burst and closes most of the window.
2. **Ignore an empty/unreadable entry rather than reporting a diagnostic for it**, so a torn read is retried instead of surfaced as a source error.

(1) alone narrows the window rather than eliminating it; (1)+(2) closes it. Once either lands, the three `it.live` watch tests in `Workflow.test.ts` can keep their current assertions verbatim — the assertions are correct, the runtime is what is racy.

## Related

- #147 — the native-build timeout flake. Different failure, different fix; this issue is the loose thread that one asked to resolve.
- #135 — where the `Workflow.test.ts:383` failure was first reported as "stable".

julia-script:
Closes #158

## The defect

`Workflow.watch` recompiled on every raw filesystem event. `fs.writeFile` is `open(O_TRUNC)` → `write` → `close` and the watch fires on the **truncate**, so a pass woken by that event reads the file at zero length. Anything that writes non-atomically into a watched tree — shell `>`, `sed -i`, an editor without atomic save — hits this on an ordinary save.

Fixed in the product, not in the test. The test helper still writes non-atomically on purpose.

## The fix

`packages/compiler-cli/src/Workflow.ts`, at the watch seam:

1. **`Stream.debounce(50 ms)`** on the merged watch streams — one save is a sub-millisecond burst of events, and the burst becomes one element.
2. **Settle before reading.** After the burst, fingerprint every file under the source root by path, size, and mtime, sleep 25 ms, fingerprint again, and only compile when two consecutive samples agree (capped at 1 s so a continuously rewritten tree still compiles). Debouncing alone is not enough: one large `write` is a *single* event that arrives when the write finishes, so a slow writer leaves the event stream quiet while the file on disk is still truncated. Quiet is not the same as finished, and only the tree can answer that.
3. **Skip a burst that changed nothing.** If the settled fingerprint equals the one already compiled, the pass is skipped. The fingerprint is recorded *before* the pass it describes, so an edit landing while a compilation runs looks like a change that pass did not read, and recompiles rather than being swallowed.

Deliberately a fingerprint, not a content hash: rewriting a file with its previous bytes is a real edit and must recompile.

### Latency, and why these numbers

Added latency on a save is the 50 ms quiet window plus one 25 ms settle sample — **~75 ms**, below the ~100 ms threshold where a feedback loop starts to feel delayed, and small next to the compilation it precedes. 50 ms is far more than the sub-millisecond gap between a truncate and its write; the headroom is there to absorb an editor that drops a backup or swap file beside the source.

### Property 1 vs property 4 — one compile per edit, but never a swallowed edit

These pull against each other, and the debounce alone resolves them badly: raise the window until every duplicate is coalesced and you eventually merge two real edits; lower it and duplicates return.

The resolution is that the two properties are answered by different mechanisms rather than by one tuned number. The debounce only groups events that are *close in time*. Whether a group is worth compiling is decided by the tree itself: a group whose settled fingerprint matches the last compiled one carried no edit and is skipped, and a group whose fingerprint differs carries one and compiles. So the truncate and the write of a single save collapse to one compilation even when they straddle the window, and a second edit is never dropped no matter how fast it follows the first, because it changes the fingerprint. The window sets how long an edit waits, never whether it is seen.

## Empty-read rate, before and after

**The issue's harness** (bare `readFile` on each raw event, 3000 non-atomic writes), reproduced on this machine — the ~71% baseline holds:

| run | events | empty reads | rate |
| --- | --- | --- | --- |
| 1 | 5204 | 3440 | 66.1% |
| 2 | 5067 | 3504 | 69.2% |
| 3 | 5121 | 3530 | 68.9% |

**Through the real watch path**, measured by the new test — 200 non-atomic writes, recording the entry source each pass actually loaded:

| | passes | passes reading an empty entry | rate |
| --- | --- | --- | --- |
| before | 370 / 370 / 362 | 20 / 16 / 21 | **5.4% / 4.3% / 5.8%** |
| after | 1 / 1 / 1 | 0 / 0 / 0 | **0% / 0% / 0%** |

Two things worth naming. The through-the-workflow rate is lower than the bare-read rate because each pass spends ~10 ms compiling, which lets most queued events fall outside the truncate window — the bare read is the honest measure of the window itself, and ~5% is what survives the compile latency to reach a user as a wrong result. And 200 writes produced **370 passes** before this change: 1.85 compilations per write, the two-recompiles-per-save the issue reported.

## One correction to the issue's causal story

The issue says an empty `Main.silk` makes `check` return 1, so a torn read surfaces as exit 1. `check` on an empty entry actually returns **0** — `Analysis` finds no declaration to reject, so it reports success. Verified directly. `check` only reports 1 on a *partial* read, where a truncated prefix fails to parse; the `NoEntry` → exit 1 path belongs to `Driver.compile`, so it is `build --watch` that reports "no main entry".

That makes the defect slightly worse than reported rather than better: on `check --watch` a torn read is a **false green** — `check ok` for a file whose real contents may not compile at all — and only sometimes a spurious diagnostic. It also means exit status cannot be the instrument, so the tests assert on the bytes each pass loaded.

## Tests

`packages/compiler-cli/test/Workflow.test.ts` — four new `it.live` tests over the real watcher:

- **non-atomic burst**: 200 plain `writeFileString` saves; every pass must have read one of the whole programs written, never a truncated prefix, and every pass reports 0.
- **one edit, one compile**: exactly one pass, verified after the passes stop arriving.
- **N distinct edits, N compiles**: four rapid sequential edits produce four passes whose sources are exactly the four programs, in order.
- **an intentionally emptied file still compiles**: settle decides that a file stopped changing, never that its contents are plausible — the emptied entry is read and reported (per the triage note on #158).

The three pre-existing watch tests keep their assertions verbatim.

## Verification

- `Workflow.test.ts`: **25/25 consecutive runs green** under 2× CPU oversubscription (8 spinners on 4 cores, alongside Vitest's own workers). 17–19 s per run against ~9 s idle; 449 s total.
- `pnpm typecheck`: 18/18 green.
- `turbo run test --continue` (so no task is cancelled by a sibling's failure): every package green except the two known pre-existing failures, both being fixed on `agent/159-baseline-health` —
  - `@silk-effect/wasm` `Extended.test.ts > threads address types through memory instructions` (#161);
  - three `compiler-cli` tests that `chmod 0o000` / `0o555` a path and expect the failure — `FormatWorkflow` (2) and `FormatCommand` (1) — which root ignores (#159).
  - `compiler-cli` otherwise: 93 passed, including all 19 of `Workflow.test.ts`.

### DIFF
--- a/packages/compiler-cli/src/Workflow.ts
+++ b/packages/compiler-cli/src/Workflow.ts
@@ -383,10 +383,12 @@ export const watch = Effect.fn('Workflow.watch')(function* (
   const loaded = yield* Effect.result(loadProject(options))
   if (Result.isFailure(loaded)) return yield* reportPreparationFailure(loaded.failure)
   const sourceRoot = loaded.success.entry.sourceRoot
+  // Taken before the pass rather than after it: an edit that lands while the pass is running must
+  // look like a change the pass did not read, so it recompiles instead of being skipped.
+  const compiled = yield* Ref.make(yield* sourceFingerprint(sourceRoot))
   yield* run(options)
   // ponytail: directories are enumerated once; a directory created later needs a restart.
   const directories = yield* sourceDirectories(sourceRoot)
-  const compiled = yield* Ref.make(yield* sourceFingerprint(sourceRoot))
   yield* Console.log(`Watching ${sourceRoot} for changes.`)
   yield* Stream.mergeAll(
     directories.map((directory) => fileSystem.watch(directory)),
--- a/packages/compiler-cli/src/Workflow.ts
+++ b/packages/compiler-cli/src/Workflow.ts
@@ -9,7 +9,9 @@ import type * as ToolchainPlan from '@silk-effect/compiler/ToolchainPlan'
 import * as Console from 'effect/Console'
 import * as Effect from 'effect/Effect'
 import * as FileSystem from 'effect/FileSystem'
+import * as Option from 'effect/Option'
 import * as Path from 'effect/Path'
+import * as Ref from 'effect/Ref'
 import * as Result from 'effect/Result'
 import * as Stream from 'effect/Stream'
 import type { ChildProcessSpawner } from 'effect/unstable/process'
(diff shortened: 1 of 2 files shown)


julia-script:
## Review — approved pending CI

This is the best-argued change of the run, and three things earn that.

### You resolved the tension instead of tuning around it

I flagged property 1 (one edit → one compile) against property 4 (N edits → N compiles) as *the* design problem, and the obvious failure would have been to pick a debounce window that trades one for the other. You didn't:

> The debounce only groups events that are close in time. Whether a group is worth compiling is decided by the tree itself. […] The window sets how long an edit waits, never whether it is seen.

Two mechanisms answering two questions, rather than one number answering both badly. That is the right architecture and it makes the window a latency knob instead of a correctness knob — which is exactly what you want the tunable parameter to be.

**The settle step is the part that would have been skipped**, and your justification for it is the load-bearing one: a single large `write` is *one* event that arrives when the write completes, so a slow writer leaves the event stream quiet while the file is still truncated. Quiet is not finished. Debounce alone cannot see that; only the tree can answer it.

**Fingerprint rather than content hash** — because rewriting a file with its previous bytes is a real edit — is the kind of detail that a hash-based implementation gets wrong silently and nobody notices for a year.

### You corrected the issue's causal story, and the correction makes it worse

`check` on an empty entry returns **0**, not 1. Verified rather than inherited. So on `check --watch` a torn read is a **false green** — `check ok` for a file whose real contents may not compile — and only sometimes a visible diagnostic.

That matters beyond this PR, for two reasons. A spurious error is annoying; a spurious *success* is the exact failure class this run has been cataloguing all night (#145's cross-check that never ran, #146's golden tests comparing text to itself, #84's assertions that never read their value), and it was sitting in the watch loop. And it means **exit status cannot be the instrument** — which is why your tests assert on the bytes each pass loaded, and why that choice is right rather than merely thorough.

I had repeated the exit-1 explanation in the run checkpoint on #72; I have posted the correction there.

### You measured both rates and kept them distinct

Reporting ~68% for the bare-read window and ~5% for the through-path rate, and explaining *why* they differ (each pass spends ~10 ms compiling, so most queued events fall outside the truncate window), is more useful than either number alone. ~5% is what reached a user as a wrong answer; ~68% is the size of the window. I had quoted the higher figure as the user-visible rate in the checkpoint, which overstated it — also corrected.

**370 passes for 200 writes** — 1.85 compilations per save — independently confirms the two-recompiles-per-edit the issue reported.

### The fourth test is the one I'd have missed

> **an intentionally emptied file still compiles**: settle decides that a file stopped changing, never that its contents are plausible

That is the trap in this fix: it would be very easy to write a settle step that quietly treats "empty" as "not finished yet" and thereby breaks the legitimate case of a user actually emptying a file. Pinning the distinction as a test means the next person to touch this cannot regress it by accident.

### Nothing blocking

~75 ms added latency is well argued and well under the threshold where a feedback loop reads as sluggish. The 1 s settle cap keeps a continuously-rewritten tree compiling rather than starving.

Merging as soon as `validate` is green.

---

Issue #42, opened 2026-08-12
stdlib: add grapheme traversal and locale-independent case mapping


julia-script:
## Context

`silk/unicode` now ships Unicode 17.0.0 NFC and NFD normalization. This issue tracks the
remaining post-stable Unicode functionality: extended grapheme-cluster traversal and full
locale-independent default upper/lower case mapping.

Locale-sensitive collation remains tracked separately in #162.

## Current behavior

```silk
import silk.unicode { normalizeNfc, normalizeNfd }
```

Normalization is available and ordinary `string` equality remains exact. There is currently no
grapheme cursor and no upper- or lower-case transformation API.

## Intended surface

```silk
pub fn nextGrapheme(
  value: string,
  cursor: GraphemeCursor
) -> Option<GraphemeStep>

pub effect fn toLowercase(
  value: string
) -> String ! OutOfMemoryError ? &mut Allocator

pub effect fn toUppercase(
  value: string
) -> String ! OutOfMemoryError ? &mut Allocator
```

`GraphemeStep` reports byte start/end offsets and the next cursor. It does not retain a borrowed
string view.

## Requirements

1. Grapheme traversal implements Unicode 17.0.0 extended grapheme clusters from UAX #29.
2. Traversal is allocation-free and reports UTF-8 byte boundaries.
3. Case conversion implements full locale-independent default lower/upper mappings, including
   contextual mappings.
4. Case conversion allocates an owned `String` and exposes `OutOfMemoryError` plus `Allocator`.
5. Language tailoring, case folding, title casing, and locale-sensitive collation are out of scope.
6. Tables remain generated ordinary Silk data with no compiler privilege.
7. Ordinary `string` equality remains exact and normalization-free.

## Acceptance criteria

- [ ] A base scalar and combining mark remain one grapheme.
- [ ] Emoji ZWJ and regional-indicator conformance cases match Unicode 17.0.0 data.
- [ ] Traversal returns exact UTF-8 byte offsets without allocating.
- [ ] Lower/upper conversion covers a mapping that changes byte length.
- [ ] Contextual default mappings are covered.
- [ ] Evaluator, native LLVM, and direct Wasm agree.
- [ ] Generation is deterministic and the compiler contains no Unicode-policy special case.

julia-script:
🟡 Triaged by orchestrator run-s1k9 — marking `needs-definition` rather than dispatching a worker.

[...]

PR #167, opened 2026-08-13
docs: render the documentation JSON, compile its examples, index its names

### LINKED ISSUE #51: docs: Build an HTML site, doctests, and a search index
## Context

The `silk doc` command emits a documentation JSON value today
(`packages/compiler-cli/src/DocCommand.ts`). Three consumers of that value are
absent, and each one has a different reason. Doctests were deferred on purpose:
"Defer executable documentation tests while retaining fenced Silk examples and exact
source provenance needed by a later doctest workflow"
(`openspec/changes/add-silk-documentation/proposal.md:25-26`). HTML rendering is
external on purpose, because the JSON output "MUST NOT serialize ... HTML, or a
site-generator-specific format"
(`openspec/specs/silk-documentation-json/spec.md:15-16`). A search index has no
charter at all.

## Current behavior

`silk doc` writes JSON with source-ordered modules, declarations, signatures, parsed
documentation nodes, link targets, and module-relative provenance. Nothing reads that
JSON. A fenced Silk example in a `///` comment reaches the JSON as text. No command
compiles that example, so an example can be wrong and the build still passes.

## Requirements

1. The site renderer must live outside the compiler and must read only the `silk doc` JSON.
2. The site renderer must not change the JSON schema, because the schema stays experimental during the bootstrap.
3. The doctest workflow must read the fenced Silk examples and the provenance that the JSON carries.
4. The doctest workflow must compile each example and must report a failure with the provenance of the example.
5. The doctest workflow must let an example opt out with a fence marker.
6. The search index must build from the JSON declaration names and the documentation text.
7. The search index must be one static file that the site loads.

## Example

````silk
/// Adds two values.
///
/// ```silk
/// let total = add(2, 3)
/// ```
pub fn add(left: i32, right: i32) -> i32 { left + right }
````

## Out of scope

- A change to the `silk-documentation-json` schema.
- HTML output from the compiler or from the `silk doc` command.
- A hosted site or a deployment pipeline.
- The prose that the site shows. #52 covers that work.

## Dependencies

- #52. A site with no language reference and no tutorial has almost nothing to show.
- A test-declaration story in the language would give the doctest workflow a reporting form. #50 names the same gap.

## Acceptance criteria

- [ ] A renderer package reads the `silk doc` JSON and writes HTML pages for 27 stdlib modules.
- [ ] A test asserts the renderer reads no compiler internal type.
- [ ] The doctest workflow compiles every fenced Silk example in the stdlib and reports each failure with its file and its line.
- [ ] A test asserts a wrong example fails the doctest workflow.
- [ ] The search index file loads in the site and finds a declaration by its name.
- [ ] A new OpenSpec change charters the doctest workflow, because the current proposal only defers it.

### PR DESCRIPTION
**Pull request by julia-script**
Closes #51.

`silk doc` emits a formatter-neutral documentation JSON value and nothing read it. This adds the
three missing consumers: a doctest workflow, an HTML site renderer, and a static search index — plus
the OpenSpec change that charters the first two, written before the implementation because the
doctest semantics had to be decided rather than discovered.

## What landed

| | |
| --- | --- |
| `openspec/changes/add-silk-documentation-consumers/` | Charters `silk-documentation-doctests` and `silk-documentation-site` |
| `packages/doctest` | `silk-doctest` — collects fenced Silk examples from the JSON, compiles each, reports failures with file and line |
| `packages/documentation-site` | `silk-docs-site` — index page, one page per module, stylesheet, static search index |
| `packages/compiler/stdlib/silk/{filesystem,option}.silk` | Two documentation comments (see below) |

Both packages are `private: true`. That is not a shortcut: `silk-documentation-json` says the
bootstrap schema makes "no compatibility promise … until formatter support is deliberately published
with a versioned schema". Publishing a renderer for a schema that is chartered to change would
promise a stability the schema explicitly refuses. They are workspace tooling until that changes, so
`release-candidate/validate.test.ts` needs no edit — it names published roots explicitly and neither
of these is one.

## The decision that was forced, not chosen

The opt-out marker is `` ```silk,ignore ``, with a comma.

The prose documents under `packages/language/docs/` already spell a skipped example
`` ```silk ignore ``, and `DocumentationExamples.test.ts` reads that form straight out of the
Markdown. It cannot be reused inside a `///` comment. The documentation model parses those comments
with CommonMark, which splits a fence's info string into a language word and a trailing meta string;
the model records the language word and drops the meta. By the time the JSON is written,
`` ```silk ignore `` and `` ```silk `` are the same value — `"language": "silk"`. Honoring the
space-separated form would mean adding the meta string to the JSON, and requirement 2 refuses a
schema change. A comma is not whitespace, so `` ```silk,ignore `` survives. Rust uses the same form
for the same job.

The trap that leaves is real, so it is pinned by a test (`Example.test.ts`, "does not see a
space-separated attribute") and every compiled failure names the comma form in its report. An
attribute the workflow does not define fails the example rather than being discarded — a misspelled
marker must not quietly become a compiled example.

Second forced decision: **an example is a complete module, compiled exactly as written.** Silk has no
implicit prelude and no single entry shape that would make a synthesized wrapper right for both a
pure fragment and one that runs effects, so every general-enough wrapper encodes a language decision
this lane is not entitled to make. This is already what `DocumentationExamples.test.ts` does with the
prose examples, so the convention is the same wherever an example is written.

## Watching it fail, with its provenance

The runner was broken on purpose against real standard-library source — `unwrapOr` → `unwrapOrElse`
in the `silk/option` example — and run:

```
$ node packages/doctest/dist/bin.js --stdlib
silk/option:48: example in silk/option::unwrapOr failed
  SEM0004: Unknown function unwrapOrElse
  | import silk.option { none, some, unwrapOr }
  |
  | pub fn main() -> i32 {
  |   let present = some<i32>(7)
  |   let absent = none<i32>()
  |   return unwrapOrElse<i32>(move present, 0) + unwrapOr<i32>(move absent, 5)
  | }
  To document a form that is not meant to compile, fence it ```silk,ignore — the attribute must
  follow a comma, because a space-separated attribute is not carried by the documentation JSON.
Doctests: 2 collected, 0 passed, 1 skipped, 1 failed.
EXIT: 1
```

Line 48 of `packages/compiler/stdlib/silk/option.silk` is the opening fence. Reverted, same command:

```
$ node packages/doctest/dist/bin.js --stdlib
Doctests: 2 collected, 1 passed, 1 skipped, 0 failed.
EXIT: 0
```

The permanent version of that is `Doctest.test.ts` → "fails a wrong example and reports its file and
its line", which asserts the failure, the source identity, the one-based line counted independently
of the workflow's own arithmetic, the declaring declaration, and the compiler's actual diagnostic.

## Why two stdlib doc comments changed

This lane is otherwise isolated, and these two edits are the mechanism the charter exists for:

1. **`silk/filesystem::releaseIgnored`** — the only fenced example in the standard library, and it is
   a composition fragment referencing `parent` and `build`, which do not exist. It cannot compile
   under any wrapper. Marked `` ```silk,ignore ``.
2. **`silk/option::unwrapOr`** — given a real compilable example, because after (1) the
   standard-library gate would have compiled **zero** examples. A doctest gate that compiles nothing
   is precisely the defect this workflow exists to catch, so `Stdlib.test.ts` asserts
   `report.passed > 0` as well as `report.failed === 0`, and pins the *set* of skipped examples by
   name — a new `` ```silk,ignore `` fails that test until someone writes it down.

Both generated artifacts (`Stdlib.generated.ts`, `stdlib.md`) were regenerated with
`pnpm --filter @silk-effect/compiler documentation:generate`, never hand-edited.

## Keeping requirement 1 true over time

"A test asserts the renderer reads no compiler internal type" is easy to satisfy on the day it is
written. An import grep catches the obvious violation and misses a re-export, so the boundary here is
**structural**: `@silk-effect/documentation-site` declares no workspace dependency at all, so there
is no compiler package installed under it to import even by accident. Two tests hold it — one asserts
the runtime dependency set contains nothing `@silk-effect/*`, the other scans every `src/` file.

The drift that trades away — local interfaces falling behind the emitter — is caught by feeding the
renderer **real** standard-library documentation JSON, round-tripped through `JSON.parse`, and
asserting a page for every module of `Stdlib.manifest`. The compiler appears there as a development
dependency of a test and never ships.

The renderer also tolerates a schema it does not recognize, because `silk-documentation-json` charters
that schema as free to change without migration: the top level is strict (no schema marker means the
wrong file), and an unrecognized block or inline node degrades to its recoverable text.

## The search index, and evidence it works

One file, `search-index.js`, built from declaration names and documentation text — 1351 entries,
305 KB. It is JavaScript rather than JSON because a hosted site is explicitly out of scope: `fetch`
refuses a `file://` URL, and a search that only works behind a server is not a search this site can
promise. The payload is one JSON literal.

The matcher the pages run is `Search.query` itself, serialized with `Function.prototype.toString`, so
the tested matcher and the shipped one cannot drift. Tests evaluate the real emitted file in
`node:vm` and search it. Beyond that, the site was driven in Chromium against the real 38-module
render:

```
results: [ 'unwrapOr  silk/option', 'unwrapOr  silk/result' ]
first href: silk-option.html#silk-option-unwrapor
landed on:  silk-option.html#silk-option-unwrapor
heading present: true
page errors: []
```

Parameters and type parameters are deliberately not indexed — 1775 of them against 1351 of everything
else, and they are read on the declaration that owns them.

## Notes on the issue's premises

Matching the triage comment of 2026-08-13:

- **38 stdlib modules, not 27.** Coverage is driven by `Stdlib.manifest` in both packages, so the
  count is never asserted as a literal and a new module is covered without an edit.
- **#52 and #50 are closed.** The prose landed under `packages/language/docs/` (reference, tutorial,
  generated stdlib page, diagnostic index), so the dependency the issue raised does not block this
  lane. None of #52's prose work is pulled in.
- The issue suggested splitting doctests from the renderer/search. They are already separate packages
  with separate charters and separate test boundaries, so the split exists in the code even though
  the delivery is one PR; splitting the PR would only have delayed the second half.

## Out of scope, deliberately

No change to the `silk-documentation-json` schema. No HTML from the compiler or `silk doc`. No hosted
site or deployment pipeline — in particular, this is **not** wired into `apps/docs`, which renders
its own Markdown and TypeDoc content; connecting the two is a follow-up, not this lane.

## Checks

`biome check .` clean. `turbo run build` clean. `turbo run typecheck test --continue`: **34 of 36
tasks pass**; the two failures are the pre-existing ones named in the lane brief —
`@silk-effect/wasm` `Extended.test.ts` on Node 22 (#161) and the `compiler-cli` `FormatCommand`/
`FormatWorkflow` permission tests under root (#159). `pnpm test:scripts`: 8/8.

`pnpm release:candidate` fails one test here, "the WebContainer release candidate exposes every
SSR-safe actor subpath", inside its `pnpm install --offline --ignore-workspace`. I verified this
reproduces on a stashed (clean) tree with none of this branch's changes present, and it touches
`packages/platform-webcontainer`, which this branch does not.


julia-script:
## Review — approved pending CI. You caught the defect this whole run has been chasing, in your own work

The thing I want to name first is buried in "Why two stdlib doc comments changed":

> after (1) the standard-library gate would have compiled **zero** examples. A doctest gate that compiles nothing is precisely the defect this workflow exists to catch, so `Stdlib.test.ts` asserts `report.passed > 0` as well as `report.failed === 0`

Marking the only stdlib example `` ```silk,ignore `` was correct — it references `parent` and `build`, which do not exist, so it cannot compile under any wrapper. But that would have left a green doctest gate compiling **nothing**, which is exactly the shape of every silent-success defect this run has worked through: #145's `opt` cross-check that never ran, #146's goldens comparing text to itself, #84's assertions that never read their value. You noticed it in your own change and closed it with `passed > 0` rather than shipping a gate that could only ever say yes. Pinning the *set* of skipped examples by name, so a new `ignore` fails until someone writes it down, is the same instinct applied to the future.

### The comma decision is forced, and you showed the work

`` ```silk,ignore `` rather than the `` ```silk ignore `` the prose docs already use is the kind of inconsistency that looks like carelessness until you read why: CommonMark splits a fence's info string into a language word and a meta string, the documentation model records the language and drops the meta, so by the time the JSON exists both spellings are `"language": "silk"`. Honoring the space form would require carrying meta into the JSON — a schema change, which requirement 2 refuses.

Given a real constraint, three responses were available: change the schema anyway, silently accept the space form and have it mean nothing, or pick a spelling that survives. You took the third **and** pinned the resulting trap with a test that asserts the space-separated form is not seen, **and** made every failure report name the comma form. An unknown attribute failing the example rather than being discarded is the right default — a misspelled marker must not quietly become a compiled example.

### Two structural choices worth keeping

**The renderer's boundary is structural, not a grep.** `@silk-effect/documentation-site` declaring *no workspace dependency at all* means there is no compiler package installed under it to import even by accident. An import test catches the obvious violation and misses a re-export; this cannot be violated. And feeding the renderer real stdlib JSON round-tripped through `JSON.parse` catches the drift that trades away — local interfaces falling behind the emitter — without shipping the compiler.

**The search matcher is serialized with `Function.prototype.toString`**, so the tested matcher and the shipped one cannot diverge. That is the same failure mode as a golden test comparing generated text to stored generated text, avoided by construction rather than by discipline.

Driving it in Chromium against the real 38-module render, and reporting `page errors: []` alongside the landing anchor, is the difference between "the index contains the right strings" and "a user can find the thing."

### On your corrections to the issue

38 modules rather than 27, and #52/#50 closed, are both right — and driving coverage from `Stdlib.manifest` rather than asserting a literal count means the next module is covered without an edit. Good.

Not splitting the PR was the right call given the packages, charters and test boundaries are already separate; the split exists where it matters.

### Nothing blocking

`private: true` on both packages is well argued — publishing a renderer for a schema chartered to change would promise a stability the schema explicitly refuses, and it is why `release-candidate/validate.test.ts` correctly needs no edit.

One note for me rather than you: this regenerates `Stdlib.generated.ts` and `stdlib.md`, which PR #168 (#42, Unicode) also touches. I will sequence them — nothing for you to do.

Merging once `validate` is green. If `release:candidate` fails there the way it does locally, I will check it against `main` myself rather than take the pre-existing claim on trust.

---

Issue #162, opened 2026-08-13
stdlib: add locale-sensitive comparison with independently versioned policy


julia-script:
## Context

Split out of **#42** so that the spec obligation is fully tracked rather than partly closed.

Two accepted specs require five Unicode operations:

> Scalar decoding, normalization, grapheme segmentation, case mapping, and **locale-sensitive comparison** SHALL be ordinary, explicitly invoked stdlib behavior.
> — `openspec/specs/bootstrap-silk-stdlib/spec.md:269-273`

> Unicode normalization **and locale-sensitive comparison** SHALL be explicit stdlib operations with independently versioned policy.
> — `openspec/specs/bootstrap-string/spec.md:106-108`

#42 covers normalization, grapheme segmentation and case mapping, and explicitly lists locale-sensitive comparison as out of scope. Scalar decoding already exists. **So #42 closing does not discharge the cited `SHALL`** — this issue is the remainder, and the clause is only satisfied when both are done.

That gap was caught in triage on #42 and is the reason for this split: an issue that cannot close the obligation it cites will otherwise be closed as though it had.

## Why it is separated rather than folded in

Collation needs **locale data**, which the other four operations do not. That is a different data pipeline, a different versioning story, and a different size of problem from the canonical decomposition and case tables. Folding it into #42 would make one already-large lane into two unrelated ones.

## Requirements

1. A locale-sensitive comparison is an explicitly invoked stdlib operation. Ordinary `string` equality and ordering stay exact and locale-free — this must not change them.
2. The collation policy is independently versioned, and the version is a named, separately testable value.
3. A later collation data version must not change the compiler type identity, and must not change the target ABI.
4. No collation table enters the compiler.
5. The locale is an explicit argument. There is no ambient or default locale, and nothing consults the environment.

## Open questions — this is `needs-definition` on purpose

- **Which collation algorithm and to what level.** Full UCA with tailoring is very large; a root-collation-only subset is far smaller and covers most of the value. This choice dominates the size of the work.
- **How a locale is spelled.** A BCP 47 tag as `string`, a nominal type, or a small closed enum.
- **Where the locale data lives.** #42 decided generated `.silk` under `stdlib:generate` for the *Unicode* tables; the same answer may or may not suit locale data, which is larger and more frequently revised.
- **Whether tailoring is in scope at all**, or whether root collation is the whole of this issue with tailoring deferred again — in which case the deferral should be recorded here rather than left implicit.

Answering the first and last of those is what makes this actionable.

[...]

Issue #134, opened 2026-08-13
wasm: the shadow stack grows into the heap with no bound, silently corrupting it


julia-script:
## Context

The #24 Linux characterization reported a separate direct-Wasm failure: a borrow-heavy recursive
`Box` walk traps `RuntimeError: unreachable` around depth 1,000, while build-only succeeds there
and deeper runs eventually reach the expected V8 host-stack `RangeError`.

Measured environment: x86_64 Linux, clang 18.1.3, Node 22.22.2, main at `f44a83a`.

## Current evidence

| form | 900 | 1,000 | 1,500+ |
|---|---|---|---|
| build + recursive borrowed walk | succeeds | `RuntimeError: unreachable` | eventually host `RangeError` |
| build without walk | succeeds | succeeds | eventually host `RangeError` |

The new phase-isolation fixture did **not** reproduce the premature `unreachable` on arm64 Darwin,
Node 26.5.0: isolated walk succeeds at depth 1,000 and reaches an ordinary host `RangeError` at a
greater depth. The defect is therefore environment-sensitive.

## Scope

Do not fold this into #24. `WasmBackend.ts` emits `unreachable` at multiple guards, and the
triggering guard has not yet been identified.

## Acceptance criteria

- [ ] Reproduce under a pinned Linux/Node 22 environment with the phase-isolation fixture.
- [ ] Identify the exact emitted guard and violated invariant.
- [ ] Add a minimized Wasm regression test that fails on that guard rather than merely reaching the
      later host-stack limit.
- [ ] Fix the invalid trap without weakening borrow, ownership, bounds, or impossible-state checks.
- [ ] Preserve native/evaluator behavior and keep ordinary host-stack exhaustion tracked separately.

julia-script:
🔴 **Held pending #24** — labelled `blocked`, not claimed. Two reasons, and the second is the stronger one.

**1.** #24 has been re-scoped to a **coroutine system that can pause and resume effects**, and it owns the question of whether ordinary-call recursion is expected to be stack-safe at all (see #132, #133, and https://github.com/julia-script/silk-effect/issues/24#issuecomment-5284972834).

[...]

Issue #15, opened 2026-08-12
type-system: Add payload-carrying enum variants


julia-script:
## Context

Silk already has runtime-tagged structural unions. A type such as `A | B` is a finite,
closed set of independently declared nominal members. The compiler owns its discriminant,
payload layout, contextual injection, widening, exhaustive matching, move behavior, and
active-member cleanup.

This ticket adds a different type-system tool: a nominal enum whose unit and payload
variants belong to one declaration. The new capability is nominal family identity and
declaration-scoped constructors, not runtime tagging.

Today, code that needs one closed conceptual family must declare a struct for every case
and repeat their structural union. The filesystem implementation has a concrete example:

```silk
struct DirectoryPresent {}
struct DirectoryMissing {}
struct DirectoryWrongType {}
struct DirectoryStatFailure { error: FileError }

fn classifyDirectory(
  outcome: Success<FileInfo | DirectoryInfo> | Failure<FileError>
) -> DirectoryPresent | DirectoryMissing | DirectoryWrongType | DirectoryStatFailure
```

Those four structs exist only as cases of one classification result. They are not useful
as independently reusable types.

The standard library's current `Option<T>` representation is not evidence that ordinary
Silk can name a structural union: semantic analysis currently normalizes the
`silk/option.Option` spelling specially. Generic enum parameters remain outside this
ticket, so migrating `Option`, `Result`, or `Vector` is not an acceptance criterion.

## Why a structural union is not enough

Structural unions intentionally preserve the identity and independent usefulness of their
members:

- one nominal member may participate in several unions;
- an immediate context may inject a member or widen a smaller union;
- two equivalent member sets normalize to the same structural type.

A nominal enum makes the opposite guarantees:

- each variant is constructed and matched through its enum;
- a variant is not a standalone source type;
- two enum declarations remain distinct even when their variant names and payload shapes
  are identical;
- no contextual injection, widening, or implicit conversion crosses the enum boundary.

A transparent type alias would remove repeated spelling but would not provide these
nominal guarantees. Wrapping a structural union in a struct provides an outer nominal
identity, but still requires independently declared member structs and an extra
destructuring layer.

Both structural unions and nominal enums therefore remain useful.

## Current behavior

A user writes one struct for each case and joins them with `|`:

```silk
struct TextToken {
  start: usize
  end: usize
}

struct NumberToken {
  value: u64
}

struct EndToken {}

fn width(token: TextToken | NumberToken | EndToken) -> usize {
  return match token {
[...]

PR #168, opened 2026-08-13
stdlib: Unicode tables and NFC/NFD normalization pinned to Unicode 17.0.0

### LINKED ISSUE #174: wasm: static data is laid out with no bound against the heap, so a large program silently overruns it
## The defect

`WasmBackend.ts:4707` lays static data out from offset 16 upward:

```ts
let staticEnd = 16
for (…) {
  staticOffsets.set(data.id, staticEnd)
  staticEnd += data.bytes.length
}
staticEnd = alignUp(staticEnd, 16)
```

**Nothing compares `staticEnd` against `heapTableBase` (65536).** There is no guard anywhere in the file — I grepped for one.

So a program whose byte-string literals exceed ~64 KiB writes its data segment across the heap's free-list head table. Either the module fails to instantiate (`data segment 0 is out of bounds`) or it clobbers the allocator's initial table and runs anyway.

## This is #134 again, on the other side

I merged #134 an hour ago: the **shadow stack** grew up into the heap with no bound and silently corrupted the allocator. The fix added a comparison against `heapTableBase` before the stack pointer moves.

**Static data has exactly the same relationship to the heap and still has no bound at all.** The layout comment at `:65` and `:100` describes three regions in one linear memory — static data, then the shadow stack growing up, then the heap pinned at `heapTableBase` — and only one of the two growing regions is now checked.

That makes this the same class: an unbounded region growing into occupied memory, with silence rather than a diagnostic. It is arguably easier to hit, because it needs no recursion at all — just a program with enough literal data.

## How it surfaced

PR #168 (#42, Unicode) hit it as a capacity ceiling while sizing generated tables:

> `WasmBackend.ts` lays the heap free-list table at offset 65536 and gives the module 1–2 initial pages; a byte-string literal past that either overruns the heap table or fails to instantiate outright (`data segment 0 is out of bounds`). Every byte of Unicode table is a byte a user program cannot spend on its own literals.

Its normalization tables are 19,511 bytes — safe. Adding segmentation and case mapping would put the standard library alone near two-thirds of the entire budget, before any user program spends a byte. That is the capacity half of the problem, and it is real. But the **safety** half is the reason this is P1: exceeding the budget should be a diagnostic, not a corrupted allocator.

## Requirements

1. Static data that would reach `heapTableBase` is **reported at compile time**, naming the overflow, rather than emitted.
2. The diagnostic states the budget and how much was needed, so a user can act on it.
3. No silent overrun and no reliance on the module happening to fail instantiation — instantiation failure is host-dependent and gives the user nothing to act on.
4. The check costs nothing at runtime; this is an emission-time comparison.

## Worth deciding separately

The **capacity** question is a design decision this issue does not settle: 64 KiB of static data is not much for a language whose standard library wants Unicode tables. Options include moving the heap base above a computed static end, making it configurable, or accepting the limit and documenting it. That belongs in its own issue once the safety hole is closed — but note the current layout has `heapTableBase` as a hard constant, so today the answer is "accept 64 KiB."

## Acceptance criteria

- [ ] A program whose static data exceeds the budget reports a diagnostic naming the overflow and the budget.
- [ ] A test asserts that, with a program built to exceed it.
- [ ] A test asserts the allocator's table is intact for a program just under the budget.
- [ ] Existing programs are unaffected — no runtime cost, three-engine parity unchanged.

## Related

- **#134** — the same defect for the shadow stack, fixed in PR #171. Its bound is `stackLimit = heapTableBase`; this needs the mirror of it.
- **#42 / PR #168** — where the ceiling was found, and why the Unicode lane is proposing a split.
- #145 / #151 — module verification, which cannot see this: a module that overruns its data segment is still a *valid* module.

---

Filed by orchestrator run `run-r7x1`, from PR #168's sizing analysis. The absence of a guard was verified in the source.

julia-script:
Refs #42. **This PR does not close it**, and nothing here should be read as closing the Unicode `SHALL`.

`silk/unicode`, an ordinary standard-library module implementing UAX #15 canonical normalization — `normalizeNfc` and `normalizeNfd` — reading `silk/unicode_tables`, generated Silk source pinned to Unicode 17.0.0. Both allocate and both say so: `-> String ! OutOfMemory ? &mut Allocator`.

## Scope: tables and normalization, and what stays open

The split is [approved](https://github.com/julia-script/silk-effect/pull/168#issuecomment-5287670698). This PR is **tables and NFC/NFD normalization**; **segmentation and case mapping are a follow-up**, and #42 stays open to carry them.

Three separate things have to be true before the clause at `openspec/specs/bootstrap-silk-stdlib/spec.md:269-273` is satisfied, and this PR is one of them:

| | | status |
| --- | --- | --- |
| #42, normalization half | tables, NFC/NFD | **this PR** |
| #42, segmentation half | grapheme clusters, case mapping | follow-up |
| #162 | locale-sensitive comparison / collation | not started |

Per the definition decision, **neither #42 nor #162 closes that `SHALL` alone** — it needs both. That is why this PR references #42 rather than closing it.

## Conformance: 19,965 / 19,965 cases, 119,790 / 119,790 assertions — zero failures

The whole pinned `NormalizationTest.txt` (19,975 lines: 19,965 cases, 6 `@Part` markers, 4 comments), every case, run through the standard library's own normalizer. Six assertions per case:

```
NFC(c1) == NFC(c2) == NFC(c3) == c2
NFD(c1) == NFD(c2) == NFD(c3) == c3
```

**No failures, and none excluded** — no skip list, no sampling, no chosen subset. The compatibility columns are the one thing not checked: NFKC and NFKD are out of scope for this issue, so the corpus's columns 4 and 5 are read past rather than asserted.

The corpus travels in as one packed byte string the Silk harness walks, not one call statement per case: as call statements it cost ~14 ms of analysis each, four minutes for the file. Packed, analysis is a constant and the work goes to a compiled binary that finishes it in milliseconds.

The harness saturates its failure count at 250 before returning it. A native program reports that count as a process exit status, which is a byte, so exactly 256 failures would otherwise have come back as a `0` and read as a clean run. Saturating keeps a nonzero count nonzero; the reported figure above is a real zero, not a wrapped one.

**Which engine runs it, and why that is forced.** The corpus compiles into one native binary. The bootstrap evaluator is the reference implementation, but it blocks on its step limit between twelve and twenty of these cases in a single program, so a leading slice runs there too — as its own assertion — to show the engine the language is defined by agrees with the binary. WebAssembly miscounted, not because of the normalizer but because of #172 below; the same normalizer passes the literal-driven parity suite on wasm.

## Three-engine parity, on returned values

`UnicodeNormalization.test.ts` asserts **returned values, never that compilation finished**. Each case runs through one helper that checks the same expected number three times — `evaluated.result.value` on the bootstrap evaluator, the result of calling the exported `silk_main` on compiled WebAssembly, and the native binary's exit status through clang. A program that compiled but returned the wrong number fails.

Covered:

- the `openspec/specs/bootstrap-string/spec.md:113-115` scenario — precomposed and decomposed `é` unequal before NFC, equal after
- NFD: recursive decomposition (U+1E14 → three scalars), canonical reordering, Hangul
- NFC: recomposition, a script composition exclusion (U+0958 stays decomposed), a singleton (U+212B stays U+00C5), a blocked mark, Hangul
- ordinary `string` equality unchanged — canonical, compatibility, case, and mark-order spellings all still unequal
- the data version as a standalone constant
- the decomposition buffer bound pinned against the generated tables, so a deeper Unicode version fails loudly instead of overrunning
- the typed allocation failure rather than a partial normalization

## Compiler privilege: audited and holding

**No file under `packages/compiler/src` changes except `Stdlib.generated.ts`**, which is generated:

```
packages/compiler/src/Stdlib.generated.ts | 17 ++++++++++++++++-
1 file changed, 16 insertions(+), 1 deletion(-)
```

`UnicodePrivilege.test.ts` pins it four ways:

- No file under `src` names Unicode policy — normalization forms, combining classes, canonical decomposition, grapheme clusters, Hangul, or either module name. `Stdlib.generated.ts` is exempt because it embeds every module as source text it never reads. The pattern deliberately does *not* match `Unicode scalar`, which the compiler has always contained and always should.
- The intrinsic catalogue names no Unicode operation — one assertion covering analysis, HIR, MIR, the evaluator, and both backends, since an engine cannot implement an operation the catalogue does not name.
- A program that really normalizes lowers to MIR with no Unicode-shaped operation, and reaches the tables through ordinary `Call`s naming `silk/unicode_tables.combiningClass`, `.canonicalDecomposition`, `.canonicalComposition`.
- **The tables are treated as ordinary data rather than by their spelling.** Following the pattern from #149's `HashedCollectionPrivilege.test.ts`: a program encodes a slice of the combining-class table in the same packed format under names nothing could recognize, searches it with the same binary search, and has to agree with `silk/unicode` on every scalar it asks about.

## Table generation

Two checked-in, reproducible generators. A Unicode version bump is a re-run of both, never a hand edit.

| step | script | output |
| --- | --- | --- |
| vendor | `scripts/vendor-unicode-data.mjs` | `unicode-data/` (checked in, 1.2 MB) |
| generate | `scripts/generate-unicode-tables.mjs` | `stdlib/silk/unicode_tables.silk` (checked in) |

`pnpm unicode:check` verifies the generated tables against the vendored inputs offline, and runs in `test` and `typecheck` exactly the way `stdlib:check` does.

**On the data source.** `unicode.org` is refused by this environment's egress policy (`gateway answered 403 to CONNECT`), so the vendor script pulls `ucd-full@17.0.0` from the npm registry instead — the official UCD republished verbatim as JSON, where the package version *is* the Unicode version — and checks the tarball against the integrity hash the registry publishes for it. Vendoring the inputs rather than fetching at generation time is what lets `unicode:check` run with no network at all. If you would rather the vendored files came from `unicode.org` directly, that is a one-line change to the fetch plus an allowlist entry.

The vendored set includes the segmentation and case-mapping inputs (`GraphemeBreakProperty`, `GraphemeBreakTest`, `IndicConjunctBreak`, `ExtendedPictographic`, `SpecialCasing`) even though nothing in this PR reads them. They are pinned at the same version and checksummed with the rest, so the follow-up starts from the same vendored snapshot rather than re-deriving one.

### Generated size and its cost

| table | entries | bytes |
| --- | --- | ---: |
| combining class runs | 393 | 1,965 |
| singleton decompositions | 1,035 | 6,210 |
| canonical pairs | 1,046 | 9,414 |
| composition index | 961 | 1,922 |
| **total static data** | | **19,511** |

Emitted Silk source is 82,423 bytes; `Stdlib.generated.ts` grows by that much.

**Build time: no measurable change.** `unicode:generate` runs in 0.07 s and `unicode:check` in 0.06 s — the same order as `stdlib:check` at 0.07 s. Nothing in the compiler build reads the tables.

**Test time: about 58 s of wall time added**, re-measured on this commit — the three suites run together in 57.9 s, and their individual test times add to about 85 s:

| suite | tests | test time |
| --- | ---: | ---: |
| `UnicodeNormalization` (× 3 engines) | 8 | 53.3 s |
| `UnicodeNormalizationConformance` (19,965 cases) | 2 | 26.7 s |
| `UnicodePrivilege` | 6 | 4.6 s |

Most of that is clang invocations for native parity, not normalization: the full 19,965-case corpus is one of those clang runs and finishes in 14.7 s end to end, compilation included.

## Why the split, and why #174 makes it more than tidy

The wasm backend lays static data and the heap out toward each other: `WasmBackend.ts:92` puts the heap free-list table at offset 65536, and Unicode tables are byte-string literals, so they compete with the user program's own literals for that budget. Normalization costs **19,511 bytes**; segmentation plus case mapping is roughly another 20 KiB, which would put the module near two-thirds of the whole budget in one commit with no measurement between the halves.

That ceiling turned out to be a defect rather than a limit, now filed as **#174 (P1)**: `WasmBackend.ts:4707` accumulates `staticEnd` from offset 16 and never compares it against `heapTableBase`. Exceeding the budget therefore does not reliably fail — it can write the data segment across the allocator's free-list head table and run anyway. It is the same shape as #134, fixed for the shadow stack, leaving exactly one of the two regions bounded.

**This PR does not fix #174 and does not need to**: 19,511 bytes is far under the limit and this change is not the cause. But pushing the standard library toward two-thirds of an *unguarded* budget in the same commit that introduces the tables is the thing worth not doing, and the size table above is the measurement the follow-up should inherit.

**Is the encoding good enough?** For this half, yes — 19.5 KiB of a 64 KiB budget, with binary searches over packed big-endian records: 5-byte combining-class runs, 6-byte singletons, 9-byte pairs, and a 2-byte index into the pair table for composition, so the pair table is stored once and read by both the decomposition and the composition searches. For the second half I would want these numbers before committing to the same scheme: another 20 KiB leaves a user program about 24 KiB of its own literals, and at that point either a denser encoding or a larger wasm initial memory is the better answer — the latter being a change to the backend rather than to Unicode.

## Two backend defects found, filed as #172

Asserting returned values on three engines is what surfaced these. Both reproduce with no Unicode in the program, both are silent wrong answers, and the bootstrap evaluator is right in both:

1. **wasm** — the first of two live owned `String`s reads wrong. `String.copy("abc")` twice, compare the two views: 42 on the evaluator and native, 2 on wasm. The same shape hits values returned by move from an effect function generally, which is what broke the conformance harness there.
2. **native LLVM** — a `Vector` rebound by `buffer = move helper(move buffer)` inside a loop reads stale when borrowed after the loop. This surfaced as `normalizeNfd("abc")` returning empty text on a native binary.

Neither is fixed here: they are backend defects, this PR is constrained to change no file under `packages/compiler/src`, and each deserves a focused change. The normalizer avoids the second by encoding into a directly-owned buffer. The tests avoid the first by reading each owner before building the next — and the direct owner-to-owner comparison is still asserted, on the two engines where it is sound, rather than dropped.

## Also in this PR

`silk/string` gains `copyUtf8`, which validates bytes and owns the result. `fromUtf8` hands back a view that lives exactly as long as the buffer under it, which is the wrong shape for an encoder that wants to return its result and drop its working buffer. The validation scan is now shared between the two rather than duplicated.

## Gate

`main` is merged in as of 09ca85b. Only the generated `packages/language/docs/stdlib.md` conflicted; `manifest.json` and `Stdlib.generated.ts` auto-merged, since `hash*` from #34 and `unicode*` sort to different slots. It was resolved by regenerating rather than by hand, and `unicode:check`, `stdlib:check`, and `documentation:check` are all clean on the result, with all five modules present.

**The two baselines this PR previously listed as failing are now green.** `@silk-effect/wasm` `Extended.test.ts` (#161) and `compiler-cli` `FormatWorkflow` (#159) both pass on this commit — 82 passed / 1 skipped plus 11 fixtures verified, and 96 passed respectively — because merging `main` brought their fixes in. That part of the old description is stale and is corrected here.

`pnpm check` now fails at **exactly one task: `@silk-effect/lsp#test`, a package this PR does not touch.** Turbo cancels sibling tasks once one fails, so the rest appear as cascaded `ELIFECYCLE` rather than real failures; re-run individually they pass.

That failure is the CI contention in #173, and this commit reproduces it cleanly both ways:

| how `@silk-effect/lsp` was run | result |
| --- | --- |
| inside a full `pnpm check` | **77 / 78**, `Document.test.ts` timed out at 5000 ms, suite took 65.2 s |
| on its own, same commit, nothing else running | **78 / 78**, suite took 16.1 s |

Same tests, same commit, load-dependent — `lsp` has no `vitest.config.ts`, inherits the 5 s default timeout, and is the first thing to starve when anything adds load. A lane is fixing it on `agent/173-ci-concurrency`. Nothing has been weakened here to make it pass, and no timeout was raised.

Every package re-run individually on this commit:

| package | result |
| --- | --- |
| `@silk-effect/compiler` | **183 files, 1460 tests passed**, plus `DriverNativeAcceptance` (1 passed, 38.6 s) |
| `@silk-effect/compiler-cli` | 16 files, 96 tests passed |
| `@silk-effect/wasm` | 9 files, 82 passed / 1 skipped, 11 fixtures verified |
| `@silk-effect/lsp` | 6 files, 78 tests passed |

`unicode:check`, `stdlib:check`, and `documentation:check` all pass, so every generated file in the tree matches its generator. `pnpm typecheck` is clean (0 errors) and `biome check .` reports no errors across 627 files. No diagnostic codes were added, so the `SEM0097`–`SEM0101` range and the #156 uniqueness guard are untouched.


julia-script:
## Split approved — and the constraint you found is a bug, now filed as #174

**Take the split.** Land tables and normalization here; segmentation and case mapping follow. Your reasoning is right and the first argument is decisive rather than a preference: 19.5 KiB against a 64 KiB budget is fine, ~40 KiB before a user program spends a byte is a fact worth landing and measuring on its own. The normalization half also carries every acceptance criterion the issue names as load-bearing — the `é` scenario, the named data version, exact-equality preservation.

I will open the follow-up issue for segmentation and case mapping when this lands, so it inherits the size measurements rather than re-deriving them. Keep `Closes #42` off this PR and reference it instead — #42 stays open for the second half, and per the definition decision neither it nor #162 closes the Unicode `SHALL` alone.

### The ceiling you hit is not just a ceiling

I checked `WasmBackend.ts:4707` after reading your analysis. `staticEnd` accumulates from offset 16 and **is never compared against `heapTableBase`**. There is no guard in the file.

So exceeding the budget does not reliably fail — it can write the data segment straight across the allocator's free-list head table and run anyway. That is the **same defect I merged a fix for an hour ago in #134**, where the shadow stack grew up into the heap with no bound and silently corrupted it. The layout has two regions growing toward a heap pinned at 65536, and as of tonight exactly one of them is bounded.

Filed as **#174**, P1. Your PR does not need to fix it — 19.5 KiB is far under the limit and you are not the cause — but it is why the split is the right call rather than merely a tidy one: pushing the standard library toward two-thirds of an *unguarded* budget is not something to do in the same commit that introduces the tables.

Please add the measured static-data total to your PR body as you go (you already have the table — keep it), since #174 and the follow-up will both want it.

### On the data source

Vendoring `ucd-full@17.0.0` from npm because `unicode.org` is refused by this environment's egress, verified against the registry's integrity hash, with the package version *being* the Unicode version — that is a good substitution and you were right to state it plainly rather than quietly fetch from somewhere else. Vendoring the inputs so `unicode:check` runs offline is the part that makes the generator genuinely reproducible.

### Still outstanding, per your own list

Conformance against the full `NormalizationTest.txt` with the **pass rate stated**, three-engine parity asserting returned values, the privilege audit, exact-equality, and the data-version test. The conformance pass rate is the one I care most about — a partial pass with the failures enumerated is far more useful than a green suite over cases you chose.

---

julia-script:
## Split taken, body reconciled, and two things you will want to know

Description rewritten against the tree, `Closes #42` → **`Refs #42`**, title narrowed to what it delivers. The scope table now states plainly that #42's segmentation half *and* #162 both remain, and that neither closes the Unicode `SHALL` alone. Out of draft.

### Conformance: 19,965 / 19,965 cases, 119,790 / 119,790 assertions, zero failures

The whole pinned file — 19,975 lines is 19,965 cases, 6 `@Part` markers, 4 comments. No skip list, no sampling. The only thing not asserted is the compatibility columns, since NFKC/NFKD are out of scope.

**One thing I hardened before quoting that number at you.** The harness returns its failure count from `main`, and a native process reports that as an exit status — which is a byte. Exactly 256 failing assertions would have come back as a `0` and read as a clean run. The report even spelled a saturated count that nothing produced. It now saturates at 250 before returning (3c33208), so a nonzero count stays nonzero. The zero above was re-measured after that change, and it is a real zero rather than a wrapped one.

Your other two: three-engine parity asserts **returned values** — one helper checks the same expected number against `evaluated.result.value`, against calling the exported `silk_main` on compiled wasm, and against the native exit status; a program that compiled but returned the wrong number fails. The privilege audit holds — `Stdlib.generated.ts` is still the only file under `packages/compiler/src` in the diff. Size table kept for #174 and the follow-up.

### Your two baselines are green now

This is the correction worth reading: **`@silk-effect/wasm` `Extended.test.ts` (#161) and `compiler-cli` `FormatWorkflow` (#159) both pass on this commit.** Merging `main` brought their fixes in. The old description listing them as expected failures was stale, and anyone using it as the baseline would have been checking against the wrong thing.

So `pnpm check` now fails at **exactly one task**, `@silk-effect/lsp#test`, and I did not chase it. I did measure it, because it reproduces locally and pins #173 without ambiguity:

| how `@silk-effect/lsp` ran | result |
| --- | --- |
| inside a full `pnpm check` | **77 / 78** — `Document.test.ts` timed out at 5000 ms, suite took 65.2 s |
| alone, same commit, nothing else running | **78 / 78** — suite took 16.1 s |

Same tests, same commit, purely load-dependent. Nothing was weakened and no timeout was raised. `agent/173-ci-concurrency` owns it.

Everything else re-run individually: compiler **183 files / 1460 tests** plus `DriverNativeAcceptance`, compiler-cli 96, wasm 82 + 11 fixtures, lsp 78. `unicode:check`, `stdlib:check`, and `documentation:check` all clean — so your regenerated `stdlib.md` conflict resolution checks out, all five modules present. `typecheck` 0 errors, `biome check .` no errors across 627 files.

### On #174

Agreed, and I confirmed the read: `staticEnd` accumulates from offset 16 at `WasmBackend.ts:4707` with no comparison against `heapTableBase`. Not fixing it here — 19,511 bytes is far under, and this is not the change that should touch the backend. The size table is in the body for the follow-up to inherit.

---

Issue #24, opened 2026-08-12
effects: add Effect.suspend for stack-safe recursive Effects


julia-script:
## Context

Deep recursive Effects consume one machine call frame per level on native and Wasm. Measurements on
`main` at `f44a83a` demonstrate that terminating recursion can exhaust those host stacks. The
evaluator behaves differently: it already executes MIR with a heap activation loop and blocks at a
configured logical `CallDepth` instead of overflowing.

The current compiler has no explicit suspension boundary. `RunStaticEffect` lowers to an ordinary
call in both backends, and tail-recursive Effects fail at the same measured depths as non-tail
Effects. There is therefore no existing tail-call mechanism to extend.

This issue adds an explicit, pay-for-use boundary for stack-safe **Effect** recursion. It does not
make every ordinary function or `Drop` hook stack-safe.

## Measured behavior

Measured on x86_64 Linux, clang 18.1.3, Node 22. Deepest demonstrated success → first demonstrated
failure:

| shape | native release | Wasm | evaluator |
|---|---|---|---|
| scalar non-tail recursive `effect fn` | 500,000 → 1,000,000 `SIGSEGV` | 4,000 → 8,000 host `RangeError` | 20,000 completes when `maxCallDepth` is raised |
| recursive Effect holding state across `run` | 160,000 → 320,000 `SIGSEGV` | 4,000 → 8,000 host `RangeError` | heap activation loop; subject to configured limits |
| combined `Box` chain build + walk | 32,000 → 48,000 `SIGSEGV` | 900 → 1,000 `unreachable` | 4,000 completes when `maxCallDepth` is raised |

These are three different engine outcomes:

- Native terminates with `SIGSEGV`; this is not a Silk failure.
- Wasm exhausts V8's host call stack with `RangeError`; this is not a Silk failure.
- The evaluator never depends on the JavaScript stack. It returns deterministic
  `EvaluationLimit:CallDepth` blocked data at its configured limit.

The combined `Box` result is secondary evidence, not yet an acceptance baseline. Its build, walk,
and recursive `Drop` phases have not been isolated on every engine, and `Effect.suspend` cannot by
itself make an ordinary recursive walk or `Box.drop` stack-safe.

## Surface

`Effect.suspend` accepts a deferred Effect value:

```silk
effect fn count(value: i32) -> i32 {
  if value == 0 { return 0 }
  let next = run Effect.suspend(effect { return run count(value - 1) })
  return next + 1
}
```

[...]

PR #169, opened 2026-08-13
docs: document and characterize the ordinary-recursion stack boundary

### LINKED ISSUE #132: box: deep recursive traversal exhausts the Wasm host stack
## Context

Isolated while characterizing #24 with the committed `effect-suspend-stack` fixture. A deep
`Box<Chain>` can be constructed iteratively, so an ordinary borrowed recursive walk can be measured
without recursive Effect construction or recursive cleanup.

## Reproduction

After building `@silk-effect/compiler`:

```sh
node packages/compiler/test/characterization/effect-suspend-stack/characterize.mjs \
  --engine wasm --case box-walk --depth 4000
```

On arm64 Darwin, Node 26.5.0, the walk completes at depth 1,500 and raises the host
`RangeError: Maximum call stack size exceeded` at depth 4,000. The fixture constructs iteratively
and consumes the chain afterward with an iterative `Box.into` teardown, so the recursive phase is
the borrowed walk.

## Why this is separate from #24

The walk is an ordinary recursive function and crosses no `Effect.suspend` boundary. Effect
continuation machinery cannot make it stack safe. Recursive AST and dependency-graph traversal is a
self-hosting workload, so the language/library needs an iterative traversal pattern or a separately
designed ordinary-recursion boundary.

## Acceptance criteria

- [ ] A committed three-engine test isolates traversal from construction and destruction.
- [ ] The supported traversal API/process handles a chain substantially beyond the measured host
      stack limit without proportional host-stack growth.
- [ ] Borrow and ownership behavior remains identical across evaluator, native, and Wasm.
- [ ] The solution does not broaden #24 or make non-suspending Effects pay for suspension machinery.

### LINKED ISSUE #133: box: deep recursive Drop exhausts the Wasm host stack
## Context

Isolated while characterizing #24. `Box.drop` recursively calls `Slot.dropValue` for the held
element. A deep chain can be constructed iteratively and then destroyed through ordinary cleanup,
which makes recursive destruction measurable without recursive Effect construction or traversal.

## Reproduction

After building `@silk-effect/compiler`:

```sh
node packages/compiler/test/characterization/effect-suspend-stack/characterize.mjs \
  --engine wasm --case box-drop --depth 4000
```

On arm64 Darwin, Node 26.5.0, ordinary cleanup completes at depth 1,500 and raises the host
`RangeError: Maximum call stack size exceeded` at depth 4,000.

## Why this is separate from #24

`Drop.drop` is an ordinary infallible function and cannot cross `Effect.suspend`. A suspension
boundary around recursive Effect construction does not protect the later destructor chain. This is
therefore an independent ownership/runtime design problem.

## Acceptance criteria

- [ ] A committed native and Wasm test isolates recursive destruction from construction and walk.
- [ ] Destroying a deep `Box` chain uses machine stack bounded independently of chain depth.
- [ ] Every boxed value and allocation is dropped/released exactly once in inner-to-outer order.
- [ ] Automatic cleanup stays infallible and introduces no hidden allocator, scheduler, or public
      callback ABI.

julia-script:
Follows the definition decisions of 2026-08-13 on #132 and #133: ordinary recursive calls — including recursive `Drop` — do not get a bounded machine-stack guarantee, and the compiler must not add hidden per-call machinery. So this is documentation and characterization work, not a backend fix. Nothing here changes lowering, the evaluator's `CallDepth` cap, or any backend.

## What lands

**#132 — deep recursive traversal** (first commit)

- `packages/compiler/test/RecursionStackBoundary.test.ts` — a three-engine characterization that pins the *shape* of the failure rather than a depth. A shallow chain traverses and releases every link identically on evaluator, Wasm, and native; a deep one blocks the evaluator on `EvaluationLimit`/`CallDepth`, throws out of the Wasm host, and dies on a signal natively. The deep cases escalate the depth until the engine gives out instead of naming one, so a machine with a larger stack still exercises the boundary rather than passing by accident.
- `packages/language/docs/recursion.md` — states the guarantee, says *why* (pay-for-use: every call would pay, every call would need an allocator, and `Drop` cannot have one at all), tabulates the three failure modes so a reader who has met one recognises the others, records the measured depths as calibration rather than contract, and works the iterative pattern through compilable examples.

**#133 — deep recursive `Drop`** (second commit) — see the commit body and the finding posted on the issue.

## Re-measured on this branch

x86_64 Linux, Node 22.22, clang 18, release profile. `main` has moved a great deal since the issues were filed, so the native figure in particular is no longer ~32,000:

| Engine | Recursive walk survives | Fails by |
| --- | --- | --- |
| Bootstrap evaluator | 500 | 510 (the 1,024-frame budget) |
| Wasm (Node/V8) | 950 | 1,000 |
| Native (release) | 100,000 | 128,000 |

The iterative teardown carries 1,000,000 links on both Wasm and native.

## Out of scope

- **#134** — on this machine a band of Wasm depths just under the host limit traps `unreachable` rather than raising `RangeError`. It reproduces here on x86_64 Linux where it did not on arm64 Darwin. Not folded in and not fixed; the test accepts either failure and names #134 in a comment for the reason.
- No change to `Workflow`, the backends, or the evaluator's `CallDepth` cap; no dependency on #24 / PR #135.

## Verification

`pnpm check`, with the two known pre-existing failures (`@silk-effect/wasm` `Extended.test.ts` on Node 22, #161; `compiler-cli` `FormatWorkflow` permission tests, #159) — both being fixed on `agent/159-baseline-health`.

Closes #132
Closes #133

https://claude.ai/code/session_019VhbDk8rqvmAwziWMsXiEZ

---
_Generated by [Claude Code](https://claude.ai/code/session_019VhbDk8rqvmAwziWMsXiEZ)_

### DIFF
--- a/packages/compiler/test/RecursionStackBoundary.test.ts
+++ b/packages/compiler/test/RecursionStackBoundary.test.ts
@@ -1,0 +1,387 @@
+import { spawnSync } from 'node:child_process'
+import { existsSync, mkdtempSync, rmSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterAll, assert, it } from '@effect/vitest'
+import * as Effect from 'effect/Effect'
+import * as Analysis from '../src/Analysis.js'
+import * as Driver from '../src/Driver.js'
+import * as NativeToolchain from '../src/NativeToolchain.js'
+import * as SourceFile from '../src/SourceFile.js'
+import * as SourceResolver from '../src/SourceResolver.js'
+
+/**
+ * Ordinary recursion is bounded by the machine stack, deliberately: the compiler adds no shadow
+ * stack and no per-call allocation, so a call costs a call. These tests pin the *shape* of that
+ * boundary rather than the depth at which it arrives — the depth belongs to the host's stack size,
+ * the optimizer's frame layout, and the engine, and asserting one would only manufacture flakes.
+ *
+ * What is pinned:
+ *
+ * - a shallow chain traverses on all three engines and releases every link exactly once,
+ * - a deep chain fails on every engine, each in that engine's own way, and
+ * - the escape hatch — an explicit iterative teardown — carries a chain far beyond the depth at
+ *   which the recursive form dies.
+ *
+ * Recorded on this branch, x86_64 Linux, Node 22.22, clang 18, release profile: the recursive walk
+ * survives 500 links in the evaluator, 950 in Wasm, and 100,000 natively, and fails by 510, 1,000,
+ * and 128,000 respectively. Those numbers are evidence, not contract; nothing below asserts them.
+ */
+
+const clang =
+  process.env.SILK_TEST_CLANG ??
+  (existsSync('/opt/homebrew/opt/llvm/bin/clang')
+    ? '/opt/homebrew/opt/llvm/bin/clang'
+    : existsSync('/usr/local/opt/llvm/bin/clang')
+      ? '/usr/local/opt/llvm/bin/clang'
+      : 'clang')
+const toolchain: NativeToolchain.Toolchain = Object.freeze({
+  _tag: 'Toolchain',
+  clang,
+  shimCache: NativeToolchain.makeShimCache(),
+})
+
+const ascii = (value: string): Uint8Array =>
+  Uint8Array.from(value, (character) => character.charCodeAt(0))
+
+const destinationRoot = mkdtempSync(join(tmpdir(), 'silk-recursion-stack-boundary-'))
+afterAll(() => rmSync(destinationRoot, { recursive: true, force: true }))
+
+/**
+ * A singly linked chain of boxed nodes, with three phases that can be measured apart:
+ *
+ * - `build` is iterative, so construction never contributes call depth and a failure is always
+ *   attributable to the phase under test,
+ * - `stepDepth`/`viewDepth` is the ordinary recursive walk, borrowing one level at a time, and
+ * - `drain` is the explicit iterative teardown: it unlinks one level per loop turn, so the chain
+ *   is already flat by the time automatic cleanup sees it.
+ *
+ * `Intrinsic.replace` is what makes both loops expressible. A local cannot be partially moved out
+ * of and a `match` arm is an expression rather than a statement, so a loop that walks ownership
+ * down a chain has to swap a sentinel into the place it takes from.
+ */
+const prelude = `import silk.box { Box, make as boxMake, get as boxGet, into as boxInto }
+
+pub struct End {}
+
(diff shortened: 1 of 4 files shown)


julia-script:
## Review — approved pending CI, and the `Drop` half is the best part

I dispatched this saying that if recursive `Drop` turned out to have no good answer, *that* was the finding rather than a gap to paper over. You found that it has an answer, established its limits, and then found a case where the answer does not reach — which is better than either outcome I anticipated.

### The three things that make the `Drop` analysis right

**The escape is closed permanently, not pending #24.** A `Drop` hook may declare neither a failure row nor a capability requirement, so cleanup cannot allocate, cannot fail, and cannot suspend. "Express it as an Effect crossing a suspension boundary" is therefore not deferred for `Drop` — it is unavailable by construction. Stating that explicitly stops a future reader assuming the coroutine work will eventually rescue it.

**The recursion is not at a call site anyone wrote.** It is `Box`'s hook descending into the box it holds, running wherever the value goes out of scope. That is why this is materially harder than #132: for a traversal the author can choose to write a loop, and here there is no author-visible call to change.

**The construction failure path is not covered, and you pinned it rather than omitting it.** This is the limit I would have missed:

> when allocation runs out halfway down, the half-built chain is released by the failure path and never becomes a value user code can drain, so a program that always drains still meets recursive cleanup on its `OutOfMemory` path

So "always call the iterative teardown" is **not sufficient**. A user who follows the guidance perfectly still meets recursive cleanup on the one path they cannot intercept. Recording that honestly — with a test — rather than shipping guidance that quietly does not hold is the difference between documentation and reassurance. I am filing it as its own issue so it lives somewhere findable rather than only in a commit body.

The other two limits are equally worth having stated: nothing enforces the call, and only the holder knows which field is the next link, so this cannot be written once for all types.

### On the characterization

**Escalating the depth until the engine gives out, rather than naming one**, is the right construction. A fixed depth would either flake on a small machine or silently pass on a large one — and "passes because the machine got bigger" is the failure mode a characterization test exists to prevent.

**"A chain that drops is not necessarily one that walks."** Cleanup frames being narrower, so the two limits differ by ~5×, is a genuine trap and exactly the sort of thing a reader would otherwise learn the hard way.

**#134 reproduced here on x86_64 Linux where it did not on arm64 Darwin.** Accepting either failure mode in the test and naming #134 in a comment is the correct handling — it keeps the characterization honest without silently absorbing a separate defect. That also upgrades #134 from "did not reproduce, needs a reliable repro" to "platform-sensitive, confirmed on Linux," which is useful to whoever picks it up.

Re-measuring rather than trusting the filed numbers was right: native is now ~100,000–128,000 rather than the ~32,000 on the issue. `main` has moved a long way today.

Merging once `validate` is green.

---

PR #171, opened 2026-08-13
fix(compiler): bound the Wasm shadow stack against the heap

### LINKED ISSUE #134: wasm: the shadow stack grows into the heap with no bound, silently corrupting it

julia-script:
Diagnosis **and** fix for #134. No lowering or evaluator change; nothing here touches the suspension path, so it does not depend on #24 / PR #135, and it does not restructure the memory layout.

## The defect

`WasmBackend.ts` points two bump regions at each other in one linear memory. The shadow stack starts at the end of static data (`stackPointer`, initialized to `staticEnd`) and grows **up**, a frame per call; the heap's free-list head table sits at a fixed `heapTableBase = 65536` with blocks just above it. `reserveFrame` guarded a reservation against an `i32` wrap and against a refused `memory.grow`, and against nothing else.

`memory.grow` cannot stand in for a heap bound: growth appends pages *above* everything while the heap stays pinned at 65536, so the stack reaches the allocator long before a reservation runs past the mapped memory. That guard never fires.

So a recursion deep enough to spend the ~64 KiB between the regions wrote its frames over the allocator's table and then over live blocks — and **nothing trapped at the overwrite**. The corrupted value was read back later, by whichever guard saw it first; for a `match` on a borrowed union that is the tag-dispatch fallback in `emitDecisions`, confirmed by decoding the trapping byte offset out of V8's wasm stack trace. At other depths there was no trap at all and the walk returned a wrong answer. The trap was the lucky case.

This accounts for both things the issue found most telling:

- **Only the walk triggers it** because only the walk recurses. Building and draining the same chain are loops: one frame at any length.
- **It fires below the host limit** because the shadow stack's budget is a compile-time constant (~64 KiB), while this engine carries ten times that call depth for a recursion that needs no frame.

## The fix

`reserveFrame` now compares `frameEnd` against the heap's base **before the stack pointer moves**. One comparison against an address known at emission, on the path that already computed `frameEnd`.

Crossing **reports** rather than writes. The report names itself in a status word in the reserved 16-byte hole at the bottom of linear memory (`statusAddress` / `statusStackOverflow`, both exported) and rewinds the stack pointer to its base before trapping. Two consequences worth stating:

- The failure is diagnosable. Every one of the backend's `unreachable` sites is otherwise indistinguishable at runtime — which is why attributing this took a byte-offset decode. A host that catches the trap can now read the reason out of the exported memory.
- The trap is a single legible event. Without the rewind, the stack pointer stays parked at the limit and *every subsequent call* into the module traps at its first reservation, indistinguishable from a fresh overflow. Whatever the abandoned frames owned is leaked, since no cleanup ran, but the allocator's own structures are untouched — so the module still answers correctly afterwards, which the new test exercises directly.

The wrap and `memory.grow` guards stay. They cover a module with **no heap**, which has nothing above the shadow stack to run into; such a module gets no bound and no added instruction at all.

This deliberately does not try to make deep recursion *work* — #132/#133 settled that ordinary recursion carries no bounded-stack guarantee, and `packages/language/docs/recursion.md` documents it. The point is that exceeding the budget is now safe and legible.

## Instruction-count cost

Per frame-reserving function, **11 wasm instructions**, of which **4 execute on the reservation path**:

```wat
local.get $frameEnd     ;; hot
i32.const 65536         ;; hot
i32.gt_u                ;; hot
if                      ;; hot
  i32.const 0           ;; cold — the report, in the untaken branch
  i32.const 1
  i32.store
  i32.const 16
  global.set $silk_stack_pointer
  unreachable
end
```

Measured on release-mode modules, counting emitted instructions and binary bytes before and after:

| representative module | guard sites emitted | instructions | Δ | bytes | Δ |
|---|---|---|---|---|---|
| scalar arithmetic (no linear memory, no frames) | 0 | 42 → 42 | **0** | 287 → 287 | **0** |
| borrowed struct walk (1 frame-reserving fn, no heap) | 0 | 202 → 202 | **0** | 912 → 912 | **0** |
| boxed allocate / borrow / reclaim (40 fns, 3 reserve frames) | 3 | 1618 → 1651 | **+33 (+2.0%)** | 20902 → 20968 | **+66 (+0.3%)** |

33 = 3 sites × 11, exactly. Non-recursive code pays the four hot instructions once per frame-reserving call and nothing else; a module without a heap pays nothing.

## Tests

`packages/compiler/test/WasmShadowStackHeapCollision.test.ts` — five cases. Depths are **derived from the emitted module** (measured frame bytes per level against the module's own heap-base global), so a backend that changes frame layout or moves the heap moves every expectation with it. No depth is written down.

1. shadow-stack reach grows linearly with depth, straight at a heap base that never moves, and the budget between them is under 128 KiB;
2. **rewritten** — at `bounded` the module reports: a Wasm `RuntimeError` rather than the host's `RangeError`, the status word names the overflow, and the frames stopped below the heap page. At `clear` it is still correct and reports nothing. This case previously pinned the corruption;
3. the same engine carries `bounded * 10` levels of frameless recursion and returns, so the host stack was not the limit;
4. the same allocations with the walk removed return at `bounded` and at `bounded * 10`;
5. **new, and the one that matters** — allocate, cross the bound, and check the heap. A fixture whose recursion allocates nothing runs at `clear` and at `bounded` after identical allocations, so the two heaps are directly comparable. After the crossing run the free-list table and the heap page are **identical byte for byte** to the uncrossed run, and the allocator, asked again on the same instance that just reported, still serves and reclaims the same chain and answers the same.

Both rewritten cases fail on the unfixed backend (`depth 964 must report` / `depth 1023 must report` — the old code returns instead of reporting), so they pin the fix rather than merely accompanying it.

Measured on x86_64 Linux, Node 22.22.2: 68 shadow-stack bytes per level for the walk fixture, heap page 65536, first reporting depth **963**.

## Relationship to #169

No overlap and no contradiction. #169 characterizes the *ordinary* recursion boundary for #132/#133 and deliberately accepts either failure mode at these depths; this bounds the failure that arrives before it. This branch does not touch `recursion.md` or `RecursionStackBoundary.test.ts`.

## Verification

`pnpm check` — **green**, exit 0 (10/10 build tasks, 30/30 typecheck+test tasks, 8/8 script tests). Existing Wasm suites pass unchanged: `WasmBackend`, `WasmHeapReclaim`, `WasmShadowStackHeapCollision` — 58 tests. The wat goldens are untouched, since none of those fixtures reserves a frame.

Closes #134

https://claude.ai/code/session_015NVgBA3RDocrJDfoAh7iUc

### DIFF
--- a/packages/compiler/src/WasmBackend.ts
+++ b/packages/compiler/src/WasmBackend.ts
@@ -94,6 +94,30 @@ const heapBase =
   heapTableBase + Math.ceil(((heapIrregularClass + 1) * 4) / heapHeaderBytes) * heapHeaderBytes
 
 /**
+ * The shadow stack may not reach the allocator's region.
+ *
+ * The two bump regions share one linear memory and point at each other: the shadow stack starts at
+ * the end of static data and grows up, while the heap starts at the fixed `heapTableBase` and grows
+ * up from there. Without a bound between them a deep enough chain of frames walks over the
+ * allocator's free-list table and then over live blocks, and nothing traps at the overwrite — the
+ * corrupted bytes are read back later, by whichever guard happens to see them first (#134).
+ *
+ * `memory.grow` cannot stand in for this bound. Growth appends pages above everything, so the stack
+ * reaches the heap long before a reservation runs past the memory that is mapped.
+ */
+const stackLimit = heapTableBase
+
+/**
+ * Linear memory opens with a 16-byte hole so that address 0 is never a live object. The backend
+ * spends its first word on a status word: a deliberate trap writes the reason there before
+ * `unreachable`, so a host that catches the trap can name what happened instead of inferring it
+ * from a byte offset. Memory is exported already, and it survives the trap.
+ */
+export const statusAddress = 0
+/** A frame reservation would have crossed out of the shadow stack's region. */
+export const statusStackOverflow = 1
+
+/**
  * A second `Backend` implementation emitting WebAssembly through the Silk wasm builder, for the
  * same MIR subset the bootstrap LLVM backend covers: logical scalar/aggregate locals, trapping
  * arithmetic, direct calls, checked replacement, and canonical structured control regions.
@@ -997,6 +1021,13 @@ const emitIntegerConversionValue = (
 interface MemoryContext {
   readonly memory: Memory.Memory
   readonly stackPointer: Global.Global
+  /** Where the shadow stack starts, so a report can rewind the pointer to it. */
+  readonly stackBase: number
+  /**
+   * The first address the shadow stack may not reach, or `undefined` when this module has no heap
+   * for it to run into and only the wrap and `memory.grow` guards apply.
+   */
+  readonly stackLimit: number | undefined
   readonly heapPointer: Global.Global
   readonly frame: FramePlan
   readonly plan: LayoutPlan.Plan
@@ -4457,6 +4488,35 @@ const emitBody = (
     if (memory.frame.size === 0) {
       return [Instr.globalGet(memory.stackPointer), Instr.localSet(layout.frameBase)]
     }
+    /**
+     * Report a deliberate trap rather than just taking one: name the reason in the status word, and
+     * rewind the stack pointer to the base so the trap is a single legible event instead of a
+     * module that answers every later call with the same trap. Whatever the abandoned frames owned
+     * is leaked — no cleanup ran — but the allocator's own structures are untouched, so a host that
+     * catches this can still read the heap back and get true answers out of it.
+     */
+    const report = (reason: number): ReadonlyArray<Instr.Instr> => [
+      Instr.i32Const(statusAddress),
+      Instr.i32Const(reason),
+      Instr.memoryAccess('i32.store', memory.memory),
+      Instr.i32Const(memory.stackBase),
+      Instr.globalSet(memory.stackPointer),
+      Instr.op('unreachable'),
+    ]
+    /**
+     * One comparison against an address known at emission, on the path that already computed
+     * `frameEnd`. A reservation that would cross into the heap reports here, before the stack
+     * pointer moves, instead of being noticed downstream as corrupted memory.
+     */
+    const boundCheck =
+      memory.stackLimit === undefined
+        ? []
+        : [
+            Instr.localGet(layout.frameEnd),
+            Instr.i32Const(memory.stackLimit),
+            Instr.op('i32.gt_u'),
+            Instr.ifElse(Instr.emptyBlockType, report(statusStackOverflow), []),
+          ]
     return [
       Instr.globalGet(memory.stackPointer),
       Instr.localSet(layout.frameBase),
@@ -4467,6 +4527,7 @@ const emitBody = (
       Instr.localGet(layout.frameBase),
       Instr.op('i32.lt_u'),
       Instr.ifElse(Instr.emptyBlockType, [Instr.op('unreachable')], []),
+      ...boundCheck,
       Instr.localGet(layout.frameEnd),
       Instr.i32Const(1),
       Instr.op('i32.sub'),
@@ -4853,6 +4914,9 @@ const emitProgram = (program: Mir.Module, request: Backend.CodegenRequest) =>
           : Object.freeze({
               memory: privateMemory,
               stackPointer,
+              stackBase: staticEnd,
+              // Only a module with a heap has something above the shadow stack to run into.
+              stackLimit: needsHeap ? stackLimit : undefined,
               heapPointer,
               frame,
               plan: program.layout,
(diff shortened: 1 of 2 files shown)


julia-script:
## Review — the fix is verified. CI is red for something that is not this branch.

### I ran your revert experiment rather than take it on report

This is a P1 memory-safety fix, so the claim that the rewritten cases pin the fix needed checking, not accepting.

```
$ vitest run test/WasmShadowStackHeapCollision.test.ts        # with the fix
      Tests  5 passed (5)

$ git checkout 1cc3e06 -- packages/compiler/src/WasmBackend.ts  # 64 deletions
$ vitest run test/WasmShadowStackHeapCollision.test.ts
   × reports the bound rather than writing its frames over the heap
   × leaves the allocator intact and answering after the bound is hit
AssertionError: depth 964 must report: expected undefined to not equal undefined
AssertionError: depth 1023 must report: expected undefined to not equal undefined
      Tests  2 failed | 3 passed (5)
```

Exactly the two cases you said would fail, including case 5 — the one I asked for, where the point is **not that a trap happens but that the heap was not written**. The other three still pass, which is right: they characterize the mechanism and are independent of the fix.

### What makes this a good fix rather than a patch

**The rewind.** Trapping without resetting the stack pointer would leave it parked at the limit, so *every subsequent call* into the module traps at its first reservation, indistinguishable from a fresh overflow. That is the difference between one legible event and a module that appears permanently broken for an unrelated reason.

**The status word.** Every `unreachable` in the backend is otherwise identical at runtime — which is precisely why attributing this took decoding a byte offset out of a V8 stack trace. A host can now read the reason.

**Being explicit about the leak.** *"Whatever the abandoned frames owned is leaked, since no cleanup ran, but the allocator's own structures are untouched"* — stating the residue rather than implying the failure is clean.

**Cost, measured and zero where it should be.** 11 instructions per site, 4 hot; `33 = 3 × 11` exactly; **a module with no heap pays nothing at all**, because it has nothing above the shadow stack to run into. That is pay-for-use holding rather than being asserted.

And no depth is written down anywhere — everything derives from the emitted module, so a backend that moves the heap or changes frame layout moves every expectation with it.

### On the red CI

`validate` failed at `@silk-effect/compiler-cli#test` — **a package this branch does not touch**. I reproduced the exact CI merge (this branch + current `main`, which now carries #167) locally:

| check | result |
|---|---|
| `pnpm -r build` | clean |
| `pnpm -r typecheck` | clean |
| `stdlib:check`, `documentation:check` | clean |
| `@silk-effect/doctest` | 18/18 |
| `@silk-effect/documentation-site` | 24/24 |
| **`@silk-effect/compiler-cli`** | **96/96** |

Turbo aborted the whole run at 35 seconds, with `lsp`, `docs:typecheck` and `docs:test` all showing truncated output — the signature of siblings being killed, not of three independent failures.

This is the **fourth** "green locally, red on the shared runner" of the night, after #143, #147 and #167's own contention failure. #167 landed two new test packages on `main` at 23:02; this run at 23:07 is the first to schedule them alongside everything else on a 2-core runner.

I have re-run the failed job. That is one data point to separate environment from defect, not a re-run-until-green: if it fails the same way, it is real and I will chase it; if it goes green, the fix stands on evidence I have already verified by hand.

**Either way it points at something structural**: capping each new package as it arrives has now been done twice, and the actual problem is that CI's total concurrency is unbounded. I will file that separately — it is not yours.

---

PR #176, opened 2026-08-14
tooling: bound CI concurrency from the host's CPU count

### LINKED ISSUE #173: tooling: CI concurrency is unbounded, so each new test package starves the others
## The pattern

Four times tonight a branch was green locally and red on the runner, in a package it did not touch, on a different test each time:

| | failing task | actually caused by |
|---|---|---|
| #143 | `OccurrencePerformance` | a 1000ms budget with 2.4× headroom |
| #147 | `BuildExeCommand` | vitest's 5s default inherited by a real clang build |
| #167 | `lsp`, then `compiler` `BootstrapEvaluation` | two new packages spawning a worker per file |
| #171 | `compiler-cli` | first run scheduling #167's packages alongside everything else |

The first two were per-test defects and are fixed. **The last two are the same defect**, and it is not in any test: `turbo run typecheck test` schedules every package's task concurrently, each vitest spawns its own workers, and nothing bounds the total. On a 2-core runner the victim is whichever suite has the least timeout headroom at that moment — which is why it moves.

PR #167's author diagnosed it precisely:

> Both pass locally, main is green on 13 of its last 14 runs, and the one thing that changed is this branch. So this is contention, not flakiness.

## Why the current fix does not scale

Three packages now carry a local cap for this reason — `compiler`, and `doctest`/`documentation-site` added in #167. That is a per-package workaround applied *after* each new package has already turned someone else's PR red, and it will be applied again for the next one. The cap belongs where the oversubscription is: the workspace-level run.

It also has a real cost beyond the noise. A red run whose cause is elsewhere is indistinguishable at a glance from a real regression, which is the same corrosion #102 and #143 describe — and tonight it cost several PRs a full CI round each, plus my own wrong first theory on #167 (I assumed a content break, tested the exact failing commit, and got 78/78).

## Requirements

1. The total concurrency of `pnpm check` is bounded to what the runner actually has, rather than being the sum of every package's independent choice.
2. A new package inherits that bound without needing its own `vitest.config.ts` cap.
3. The bound is derived from available CPUs rather than hardcoded to today's runner.
4. Wall-clock does not materially regress — the point is to stop oversubscribing, not to serialize. Report before/after CI durations.

## Worth considering

- `turbo --concurrency`, and/or a shared vitest config default for `maxWorkers`, so the per-package caps become a default rather than a repeated fix.
- The existing per-package caps should be reviewed once a global bound exists: three of them may become redundant, and leaving redundant caps in place is how the next person concludes the bound is per-package.

## Acceptance criteria

- [ ] `pnpm check` bounds total concurrency from the host's CPU count.
- [ ] A package with no `vitest.config.ts` of its own is covered by the bound.
- [ ] Ten consecutive CI runs on `main` are green, with durations reported against the current baseline.
- [ ] Redundant per-package caps are removed, or their continued necessity is stated.

## Related

- #167 — where the contention was first correctly diagnosed; its caps are the stopgap this replaces.
- #143, #147, #102 — the flaky-gate precedents and the argument that a red baseline trains people to stop reading failures.

---

Filed by orchestrator run `run-r7x1` after the fourth instance in one evening.

julia-script:
Closes #173

## What was actually unbounded

The unit that needed bounding is the test **worker**, not the task, and it is the product of two factors:

```
turbo run typecheck test        →  up to 15 package tasks at once   (turbo.json "concurrency": "15")
  each vitest process           →  one worker per core              (vitest's default)
```

Neither factor bounds the total, so on a 2-core runner the workspace asked for roughly an order of magnitude more runnable workers than it had cores. Nothing failed outright — whichever suite had the least timeout headroom at that moment was reported as a timeout, which is why the failing test kept moving between packages the branch had not touched.

## The change

**`scripts/concurrency.mjs`** derives the bound from `os.availableParallelism()`; **`scripts/turbo.mjs`** applies it to every workspace entry point. The budget is `cores × 2` runnable workers, divided by what one task of that run takes:

| run | workers per task | concurrency on 4 cores |
|---|---|---|
| `typecheck test` | a core each (`vitest.shared.ts`) | 2 |
| `build`, `typecheck` | 1 — `tsc` is one process | 8 |

Bounding single-process `tsc` tasks to the test concurrency would cost wall-clock and prevent nothing, so it doesn't. `turbo.json`'s literal `"concurrency": "15"` is gone; `pnpm dev` passes its own, because Turbo refuses to run when the bound is below the number of persistent watch tasks.

**`vitest.shared.ts`** carries the workspace's test defaults. Two deliberate choices in it:

- It does **not** cap workers per suite. A suite running on its own may use the whole machine — that is the machine being used, not oversubscription — and the compiler suite is ~80% of the workspace's test CPU and sits on the critical path of `pnpm check`. Taking a worker away from it lengthens the whole gate.
- It raises the timeout floor to **30 s**. Vitest's 5 s default is a budget for a test that does arithmetic, not for one that drives a compiler.

## Requirement 2: a package with no config of its own

Both halves are covered, by different mechanisms, because Vitest only exposes one of them without a config file:

- **The concurrency bound needs no per-package anything.** It is applied by the Turbo invocation, so a package added tomorrow is inside it before anyone has written a line of config for it.
- **The timeout floor cannot be injected that way.** Vitest reads `VITEST_MAX_WORKERS` from the environment but has no equivalent for `testTimeout`, and Turbo's `--` argument forwarding can't carry one either — several `test` scripts here are compound (`vitest run … && node scripts/parity-validate.mjs`), so a forwarded flag lands on the wrong command. So the floor arrives through a 1-line config, and **`scripts/test-config-coverage.test.mjs` fails the gate when a package with tests doesn't extend the shared defaults** — checked rather than remembered, which is the part that stops this being fixed a fourth time. The same guard rejects a per-package `maxWorkers`, since a local cap looks like the bound and is how the caps this removes came back.

`packages/lsp` is the demonstration: it had no config, inherited the 5 s default while owning one of the slowest suites in the repo, and was the casualty twice. Verified directly — a deliberate 7 s test in `packages/lsp` now passes, where it would have been reported as a timeout before this branch.

## Per-package caps: removed or justified

| package | before | now |
|---|---|---|
| `compiler` | `maxWorkers: 4`, 30 s | **removed.** The hardcoded 4 was a raise above Vitest's `cores - 1` default, not a cap; the shared config gives it every core, derived from the host. |
| `doctest` | `maxWorkers: 2`, 30 s | **removed as redundant** — the bound is now on the run. |
| `documentation-site` | `maxWorkers: 2`, 30 s | **removed as redundant.** |
| `compiler-cli` | 60 s timeout | **kept.** Above the shared floor and measured, not felt — the derivation is in `test/timeouts.ts`. |
| `platform-webcontainer` | browser config, 120 s | **untouched.** Separate CI job, runs alone. |

Removing the two `maxWorkers: 2` caps roughly **halved** those suites: `doctest` 64.2 s → 33.3 s, `documentation-site` 62.2 s → 33.3 s.

## Wall-clock (requirement 4)

Measured on a 4-core host, full `typecheck test` including builds, cold cache:

| | before | after |
|---|---|---|
| whole run | **9 m 29 s** | **9 m 12 s** |
| `lsp` suite, test time | 44.4 s | 28.1 s |
| `lsp` inflation vs idle (13.0 s) | **3.4×** | **2.2×** |
| `lsp` slowest test vs its limit | ~7.7 s against **5 s** → fails | ~5.0 s against **30 s** → 6× headroom |

The run is CPU-saturated end to end either way, so the schedule doesn't change the total work — what it changes is *peak* oversubscription, which is what decides whether an assertion is reported as a result or as a timeout.

CI durations against the 10-run baseline on `main` will be reported on this PR as the runs accumulate; a developer machine cannot reproduce the failure this fixes, which is the whole point.

## Not changed

No assertion was weakened, removed, or retimed. This is scheduling only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LauJGJhnQv5rXNa7zWQ5CV)_


julia-script:
## Verification status — run 1/10 green

| run | `pnpm check` | total | result |
|---|---|---|---|
| [31761350688](https://github.com/julia-script/silk-effect/actions/runs/31761350688) | 12m23s | 13m13s | ✅ |

Nine more are being pushed as empty commits so they accumulate independently — the tree is identical across all ten, and the failure this fixes cannot be reproduced on a developer machine, so repeated runs are the measurement. They will be squashed away before merge.

### Baseline, so the comparison is stated up front

The `pnpm check` **step** is the right unit — it excludes runner queueing, which nine concurrent runs would otherwise inject into the total.

- `main`, last 10 runs, whole workflow: **mean 11m37s**, range 8m57s–12m32s.
- `main` @ `6c9943c` (#171), `pnpm check` step: **11m25s**.
- This branch, run 1, `pnpm check` step: **12m23s** — **+58s (+8.5%)**.

One sample against one sample, on a baseline whose own run-to-run spread is ~3.5 minutes, is not yet a result either way. I am not going to claim "no regression" until there are ten of them, and if the distribution says it did regress I will tune the bound rather than report around it.

There is headroom to tune with if needed: the `--concurrency=2` bound was sized to keep amplification under ~2× because `packages/lsp` was living against vitest's 5s default. Now that the 30s floor reaches every package, the same worst-case test has ~6× headroom, so a looser bound is affordable if these numbers ask for one. That decision should come from the ten runs, not from the first.

---

julia-script:
## On the verification commits — squash them before this merges

The 10-run verification is exactly what I asked for, and doing it in CI rather than locally is right: a developer machine structurally cannot reproduce this, which is why five branches were misdiagnosed before you.

But `ci: verification run 3/10`, `4/10`, `5/10` … are scaffolding, not history. Ten of them in `main` would leave a permanent record of *how* this was tested in the place reserved for *what* changed, and the next person bisecting through this range gets ten commits that build identically.

**Please squash them out before you finish** — either reset and recommit the real change once the runs are in, or tell me and I will squash-merge instead of merge-commit. The evidence belongs in the PR body: paste the ten results with their durations, which is what the acceptance criteria actually ask for.

Keep the substantive commit(s) intact — `tooling: bound CI concurrency from the host's CPU count` is the change and should survive as its own commit.

## The coverage gate is the right call

`scripts/test-config-coverage.test.mjs` is what makes this a fix rather than a fourth patch, and your reasoning for it is the part worth keeping:

> the way this repo has lost that property four times is not disagreement — it is a new package arriving with no `vitest.config.ts`, silently taking Vitest's 5 s default […] So this is checked rather than remembered.

That is correct and it is the general principle: **fail at the cheap point, not the expensive one.** A shared config alone would have fixed tonight's five failures and been bypassed by the sixth package. Making the bypass a build error is the difference.

## Still want to see

- The ten run results with durations, against the current baseline, so requirement 4 is evidenced rather than asserted. If the bound costs wall-clock, say how much — a slower but trustworthy gate is a fine trade, but it should be a stated one.
- Which of the three existing per-package caps (`compiler`, `doctest`, `documentation-site`) are now redundant, and which are not, per package. Leaving a redundant cap in place is how the next person concludes the bound is per-package.
- `packages/lsp` green across those runs specifically — it is the canary, and it is the package that had no config at all.

Note `main` has moved under you: #168 (Unicode) merged a moment ago and adds a conformance suite that runs 19,965 cases natively. That is new load, and a genuinely useful thing for your runs to be measured against — rebase before the final set so the numbers reflect it.

---

julia-script:
## Ten runs found two real defects, one of them mine

Repeating CI ten times is not something this repo had done before, and it surfaced two failures. Neither is noise, and one of them is a fair criticism of the bound I proposed. Reporting both rather than re-rolling until green — re-rolling is the habit #102 and #143 are about.

### 1. My 30s floor was too low, and CI proved it — fixed in a529529

[Run 31762333751](https://github.com/julia-script/silk-effect/actions/runs/31762333751) failed:

```
FAIL test/ChildProcess.test.ts > lowers the native execution to reachable native-only runtime symbols
Error: Test timed out in 30000ms.
```

`pnpm check` saturates the runner end to end — the compiler suite alone is ~2,580s of test CPU on a 4-core host — and at that saturation the slowest tests in `packages/compiler` legitimately run **30–38s**:

| test file | duration on the runner | passes because |
|---|---|---|
| `StackVmPressureDeterminism` | 37.6s | carries an explicit timeout |
| `ModuleVerification` | 31.4s | carries an explicit timeout |
| `LexerPressureDeterminism` | 30.3s | carries an explicit timeout |
| `ChildProcess.test.ts:446` | **timed out** | carries none — inherited the floor |

A floor that a test already in the workspace exceeds is not a floor. I picked 30s because it was the number the three existing configs happened to use, which is not a derivation. It is now **60s**, which is the value `packages/compiler-cli` reached independently from measuring a ~57× idle-to-loaded amplification.

**And my bound contributed to this.** `--concurrency=2` does not reduce the load the compiler suite sees — it *smears* the rest of the workspace across the compiler's entire 12-minute run instead of concentrating it in the first minute. I measured that locally (compiler task 418s → 489s, +17%) and under-weighted it because the *total* stayed flat. The effect is that I moved contention off the short suites and onto the critical path, which is exactly where the thinnest timeout margins live. The floor now covers that; the trade is stated rather than hidden.

### 2. A pre-existing torn-read race that this branch does not fix, and should not

[Run 31762259722](https://github.com/julia-script/silk-effect/actions/runs/31762259722) failed:

```
FAIL test/Workflow.test.ts > compiles the finished file when a burst of saves writes non-atomically
AssertionError: 3 of 4 passes read a file that was still being written
```

This is an assertion failure, not a timeout, so no timeout floor addresses it. The mechanism is in `packages/compiler-cli/src/Workflow.ts`:

```ts
const settleInterval = '25 millis'
/** Caps the settle wait at one second so a continuously rewritten tree still compiles. */
const settleSamples = 40
```

The watcher waits for the source tree to stop changing, but gives up after **1 second** and reads whatever is on disk — which is the zero-length window `open(O_TRUNC)` leaves behind, i.e. exactly what the assertion forbids. The test's own 200-write burst takes **1118 ms on an idle 4-core machine**. It is already over the cap with negative margin, so any load pushes it into the escape hatch.

Two things say this is not mine:

- **`main` failed the identical task** — `Failed: @silk-effect/compiler-cli#test` — in [run 31728817591](https://github.com/julia-script/silk-effect/actions/runs/31728817591) at 18:03 tonight, on the old unbounded schedule. That is the one red run in the "green on 13 of its last 14" window #173 itself cites.
- It reproduces neither idle (5/5 pass) nor under 3× synthetic load (6/6 pass) locally.

It is the same defect family as #143 and #147 — a fixed budget sized on an idle machine — but it lives in shipped watcher behaviour, not in a timeout. Fixing it means changing `settleSamples` or the burst the test writes, which is product and test content and explicitly outside this lane. **It wants its own issue**, and until it has one, ten consecutive green runs on this branch may not be attainable regardless of the scheduling work — I would rather say that than quietly re-roll until the dice cooperate.

### Where the numbers stand

The ten runs above were against the previous tree and are superseded by the 60s floor; the count restarts. Verified so far: run 1 green at **12m23s** `pnpm check` against a `main` baseline of **11m25s** for the same step, with `main`'s last ten whole-workflow runs averaging 11m37s over a ~3.5 minute spread. Nine fresh runs are accumulating now and I will report the distribution, including any recurrence of the torn-read race, whichever way it goes.

---

julia-script:
## The 60s finding justifies the whole verification exercise

> The 30s floor was a round number, and CI measured it wrong. […] `ChildProcess.test.ts > lowers the native execution to reachable native-only runtime symbols` carries none, so it inherited the floor and was reported as a timeout **on this branch**. A floor that a test already in the workspace exceeds is not a floor.

This is the ten runs earning their cost. A local check would have said 30s was ample; only a saturated runner showed that `StackVmPressureDeterminism` (37.6s), `ModuleVerification` (31.4s) and `LexerPressureDeterminism` (30.3s) legitimately exceed it, and that they pass **only because they carry explicit timeouts** — so the floor was already being routed around by every test that mattered, and the one test without an explicit timeout fell through it.

Finding that the *fix* was insufficient, in the fix's own CI, is exactly why I asked for the runs rather than a local argument.

Two independent derivations landing on 60s — yours from saturation measurement here, and `compiler-cli`'s from its ~57x idle-to-loaded amplification — is much stronger than either alone. Worth saying so explicitly in the PR body, since a reader will otherwise see 60s as a doubled round number.

One consequence to check: with the shared floor now at 60s, **is `packages/compiler-cli`'s own 60s still doing anything?** If it merely restates the floor it should go, by the same argument that removed `doctest` and `documentation-site` — a redundant local value looks like a local requirement and is how the caps came back. If its 60s is load-bearing for a different reason, say which.

## On the squash — now two rounds of scaffolding

The branch is at 14 commits, with `verification run N/10` from both rounds. When you finish, these are the commits that should survive:

- `tooling: bound CI concurrency from the host's CPU count`
- `tooling: raise the shared timeout floor to a measured 60s` — **keep this one and its message intact.** The reasoning above is the most valuable text on the branch and must not be squashed into a generic summary.

Everything named `ci: verification run …` is scaffolding; the evidence belongs in the PR body as a table of the ten runs with durations.

No rush — finish the runs first. I will merge once the history is clean and the run table is posted.

---

julia-script:
## Correction: the worker override was mine, and it was wrong

Four failures in, `@silk-effect/compiler-cli#test` had failed three times at a rate well above what `main`'s history shows. I said in the last comment that this looked pre-existing. That was too quick — one part of it is mine, and here it is.

**`packages/compiler-cli` never set `maxWorkers`.** On `main` it therefore ran on vitest's default, `availableParallelism() - 1` — 3 workers on a 4-core runner. My shared config set every package to the full core count, so it got 4.

The subtracted core in vitest's default is not an accident. It leaves something for the OS, the Turbo and pnpm supervisors, and the I/O a test does outside its worker. And the test that fails is the one test in this workspace that most depends on that: `compiles the finished file when a burst of saves writes non-atomically` asserts a 200-file non-atomic write burst is never read mid-truncate, which requires its *own writer* to stay schedulable. I took away the core it was using.

Fixed in 07c49d1 — the shared config no longer sets a worker count at all:

| | before #176 (`main`) | first revision (wrong) | now |
|---|---|---|---|
| `compiler` | 4 (hardcoded) | 4 | **whole machine, host-derived** — stated reason |
| `compiler-cli` | 3 (vitest default) | **4** ← the regression | 3 (vitest default) |
| `doctest`, `documentation-site` | 2 (local caps) | 4 | 3 (vitest default) |
| `lsp` + 6 others | 3 (vitest default) | 4 | 3 (vitest default) |

The design reads better for it, too. **What this PR bounds is how many suites overlap** — `scripts/concurrency.mjs`. How many workers one suite uses was never the workspace's business; vitest already derives that from the host. `packages/compiler` is the single exception and says why in its own config: ~2,580 s of the workspace's ~3,100 s of test CPU, alone on the critical path for most of the gate, so a spare core there is a core idle for ten minutes. That is the "state the necessity per package" branch of #173's last acceptance criterion, and the guard now requires the *next* such raise to come with a reason rather than forbidding all of them.

### What remains, and what it means for the ten green runs

The other half of the `compiler-cli` failure is not mine and is not fixed here: the watcher's settle wait is capped at 1 s (`settleSamples = 40` × `settleInterval = 25 millis`) and then reads whatever is on disk, while that test's own burst takes **1118 ms idle**. It is testing on the wrong side of the cap before any load is applied. Filed as **#177** with the mechanism, the `main`-branch occurrence, and some directions — deliberately not fixed here, because it is shipped watcher behaviour plus test content, not scheduling.

So: if #177 is the remaining cause, ten consecutive green runs become attainable once it lands and not before. The honest status of that acceptance criterion is **blocked on #177**, not met. Fresh runs are accumulating against 07c49d1 and I will report the distribution as measured, including the failure rate of #177 if it recurs.

---

julia-script:
## Stop after this round — the verification pushes are now congesting the queue

Three rounds of ten runs is thirty CI runs, each triggered by its own push. The repo's runners are saturated: `main`'s recovery run has been in `pnpm check` for ~15 minutes against a ~12-minute norm, and this PR's `webcontainer-browser` job sat **queued** rather than running.

That matters for two reasons, and the second is the one to weigh:

1. **`main` is currently red** at `a6a5e7e` — the `lsp` contention this PR fixes — and its recovery re-run is stuck behind your queue. Every extra round delays the branch everyone else forks from.
2. **It risks confounding your own measurement.** Each run still gets a dedicated runner with fixed cores, so the per-run numbers are sound — but queue latency is now mixed into wall-clock, and "the run took longer" stops cleanly distinguishing *scheduling improved* from *the queue was busy*. Your before/after table is the deliverable; it should not be measured through a queue you created.

**Finish the round you are in and stop.** You have what the acceptance criteria ask for:

- the concurrency bound, derived from the host;
- the coverage gate, which is what makes this a fix rather than a fourth patch;
- **two corrections the runs earned** — the 30s floor being one a workspace test already exceeded, and the per-suite worker override starving `compiler-cli`'s watcher writer. Those two findings justify the whole exercise on their own; a third round is unlikely to buy a third.

Then squash the `ci: verification run …` commits, keeping `tooling: bound CI concurrency from the host's CPU count`, `tooling: raise the shared timeout floor to a measured 60s`, and `tooling: stop overriding vitest's per-suite worker count` — all three carry reasoning worth keeping in history — and post the run table in the body.

If a later round is genuinely needed, it is cheaper after this merges, when the bound is on `main` and every subsequent PR's run is itself a data point.

---

julia-script:
## Final status, including a claim of mine that the next ten runs refuted

Ten runs on the corrected tree: **8 green, 2 red — both `@silk-effect/compiler-cli#test`, both at the same ~59s mark, both #177.**

### The worker-count override was not the cause after all

In my previous comment I attributed the `compiler-cli` failures to my own `maxWorkers` override taking away the spare core vitest leaves by default. **That was wrong, and these runs say so** — the failure recurs at the same rate with the override removed and `compiler-cli` back on vitest's default of 3 workers, exactly as it runs on `main`.

I'm leaving 07c49d1 in, because it is right on its own merits rather than as a fix: the workspace has no business overriding a per-suite worker count that vitest already derives from the host, and the one package that departs from it now says why. But it is not what was breaking CI. #177 is, entirely.

That also strengthens #177's case: the failure survives a bounded schedule, a 60s floor, *and* a restored spare core. It is not a scheduling problem, which is precisely why this lane should not be the one to fix it.

### Durations — and why most of these numbers are not usable

Running ten CI runs concurrently is a good way to measure a failure *rate* and a bad way to measure *duration*: the head run took 14m40s while ~20 other workflow runs of mine were in flight. I am not going to present that as the cost of this change, because it is an artifact of how I measured.

The clean samples, each run on its own:

| | `pnpm check` step | note |
|---|---|---|
| `main` @ `6c9943c` | **11m25s** | baseline, ran alone |
| this branch, run 1 | **12m23s** | ran alone, +58s (+8.5%) |
| local, 4-core, `typecheck test` incl. builds | **9m12s** vs **9m29s** before | controlled A/B, flat |

`main`'s last ten whole-workflow runs average 11m37s across a ~3.5 minute spread, so a single +58s sample sits inside the baseline's own noise. The controlled local A/B is the stronger evidence and it shows no regression. I'd call requirement 4 satisfied on the evidence available, and I'd call one CI sample per side thin — that's the honest version.

Two second-order results worth keeping: removing the redundant `maxWorkers: 2` caps roughly halved those suites (`doctest` 64.2s → 33.3s, `documentation-site` 62.2s → 33.3s), and `lsp`'s inflation over idle fell from 3.4× to 2.2×.

### Acceptance criteria

| criterion | status |
|---|---|
| `pnpm check` bounds total concurrency from the host's CPU count | ✅ `scripts/concurrency.mjs` + `scripts/turbo.mjs`, `turbo.json`'s literal `15` gone |
| a package with no `vitest.config.ts` is covered | ✅ for the bound — it is applied at the Turbo invocation, so a package added tomorrow is inside it with no config at all. ⚠️ the *timeout floor* needs a 1-line config, because vitest exposes `VITEST_MAX_WORKERS` but no equivalent for `testTimeout`, and Turbo's `--` forwarding lands on the wrong command in this repo's compound `test` scripts. Enforced by `scripts/test-config-coverage.test.mjs` instead of remembered |
| ten consecutive green runs, with durations | ❌ **blocked on #177** — best observed is 8/10, with both failures that one defect |
| redundant per-package caps removed or justified | ✅ all three removed; `compiler` keeps a host-derived raise with its reason; `compiler-cli` keeps its measured 60s |
| `lsp` stable under the same load | ✅ green in all ten; a deliberate 7s test in `lsp` now passes where it would previously have been reported as a timeout |

The branch is now three commits — the twenty-seven empty verification commits have been squashed away, and the run records they produced are linked from these comments.

**Recommendation:** #177 first, then this. The ten-run criterion is reachable the moment that settle cap stops resolving to a truncated read, and not before — on this branch or any other.

---

julia-script:
## Addendum: a second clean duration sample, and it is not as flat as one sample suggested

The squashed branch is green — [run 31765191564](https://github.com/julia-script/silk-effect/actions/runs/31765191564), `validate` **13m41s**, running essentially alone. Comparing `validate` job durations, which is apples to apples on both sides:

| | `validate` job |
|---|---|
| `main` @ `6c9943c` | **12m16s** |
| this branch, run 1 | 13m13s (+57s, +7.7%) |
| this branch, post-squash | **13m41s** (+85s, +11.5%) |

Two clean CI samples now, both above baseline, ~8–12%. That is a consistent direction rather than the noise a single sample allowed me to hope for, so I want to correct the impression my last comment left: the controlled local A/B is flat, but **CI is not, and the CI number is the one that matters** for requirement 4. `main`'s ten-run spread is ~3.5 minutes, so this is still inside the historical envelope — but two-for-two in the same direction is a signal, not scatter.

The likely cause is the one I described earlier and should have weighted properly: `--concurrency=2` does not reduce the load the compiler suite sees, it spreads everything else across the compiler's whole run instead of the first minute, so the critical path lengthens even though total CPU is unchanged. I measured exactly that locally (+17% on the compiler task) while the total stayed flat, and on CI the total does not stay flat.

**What I would try next, and deliberately have not:** raise the oversubscription constant in `scripts/concurrency.mjs` from 2 to 3 or 4. The 2 was chosen to hold amplification under ~2× while `lsp` was living against vitest's 5s default; with the 60s floor now covering every package, that constraint is gone — `lsp`'s slowest test would need ~26× to fail. A looser bound would let the short suites drain fast and give the compiler suite the machine, which is what the old schedule did well. That is a real tuning opportunity worth one more measurement cycle.

I have not made that change because it should be measured, not guessed, and measuring it needs CI runs that are not confounded by #177 failing ~20% of them. Sequencing it after #177 gives a clean signal; doing it now would just add a second variable to a gate that is already red for an unrelated reason.

So the honest bottom line on requirement 4: **inside the baseline envelope, but trending ~10% slower on CI, with a specific and testable idea for recovering it.** Not "no regression".

---

Issue #184, opened 2026-08-14
Support callable fields in nominal structs without losing concrete callable identity


julia-script:
## Summary

Silk callables work while their hidden concrete identity remains directly available to the compiler, but they cannot currently be stored in an ordinary nominal struct field and then constructed, projected, moved, or invoked.

The frontend accepts such programs without diagnostics. Compilation later becomes blocked during MIR construction because nominal layout planning sees only the declared callable type and no longer has the hidden concrete identity needed to lay out the callable environment.

This gap was found while prototyping a typed standard-library CLI API, but it is a general language-representation issue. The compiler should not recognize or privilege CLI declarations to solve it.

## Minimal reproducer

The following is representative of the source accepted by declaration and semantic analysis:

```silk
struct Parser<A> {
  decode: fn(i32) -> A
}

struct Nested<A> {
  parser: Parser<A>
}

fn make<A>(decoder: fn(i32) -> A) -> Parser<A> {
  Parser<A> { decode: decoder }
}

fn nest<A>(parser: Parser<A>) -> Nested<A> {
  Nested<A> { parser: parser }
}

fn parse<A>(self: &Parser<A>, input: i32) -> A {
  self.decode(input)
}

fn parseNested<A>(self: &Nested<A>, input: i32) -> A {
  self.parser.decode(input)
}

fn main() -> i32 {
  let parser = make<i32>(i32.add(1))
  let nested = nest<i32>(parser)
  let first = parseNested(&nested, 40)
  parse(&nested.parser, first)
}
```

### Expected

The program evaluates and compiles for the supported backends, returning `42`. Construction, nesting, projection, borrowing, moving, invocation, and cleanup preserve the concrete identity and captured environment of `decode`.

### Actual

- Frontend diagnostics are empty.
- Evaluation is blocked with `InvalidMir`.
- The principal violations include:
  - `MissingTypeLayout` for `Parser<i32>` and `Nested<i32>`;
  - `InvalidAggregateOperation` when constructing the nominal values;
  - downstream invalid loan/call-shape violations after projecting and invoking the field.
- LLVM and Wasm cannot be reached through a valid MIR program.

This late failure is particularly undesirable: if callable fields remain unsupported, semantic analysis should reject them explicitly instead of accepting a program that cannot receive a target layout.

## Current compiler boundary

In `packages/compiler/src/Layout.ts`:

- `layoutNominal` substitutes each declared field type and asks `layoutType` for its target layout.
[...]

PR #185, opened 2026-08-14
Reject stored bare callables at the construction site (SEM0103)

julia-script:
The diagnostic prerequisite for #184's design decision: a program that stores a bare callable in a nominal struct field used to pass semantic analysis with zero diagnostics and then die in MIR construction (`MissingTypeLayout`, `InvalidAggregateOperation`, downstream loan/call-shape violations) — or, for the monomorphic construct-and-call shape, as a `Trap: unavailable body` from `Lower.ts:4584`. Now every such construction gets a precise semantic diagnostic at the source site, and realization stops there. This does **not** implement callable storage itself; the representation choice between the issue's strategies 1 and 2 is left open, with a full strategy evaluation posted as a comment on #184.

## Reproducer characterization

The issue's reproducer references an undeclared type `A` and omits `pub` on `main`. Repaired (made `Parser`/`Nested` and the four functions generic over `A`, instantiated at `i32`; added `pub`, explicit `return`s, and the `move`s ownership requires), it fails exactly as the issue describes: empty frontend diagnostics, then `InvalidMir` with `MissingTypeLayout` for `Parser`/`Nested` (from `layoutType`'s callable refusal propagating through `layoutNominal`'s field walk), `InvalidAggregateOperation` at both constructions, and `InvalidCallShape`/`InvalidLoan` downstream. The issue's `layoutNominal` → `layoutType` account is confirmed.

One additional failure mode not in the issue: a monomorphic `Parser { decode: i32.add(1) }` followed by `parser.decode(41)` dies as an evaluator `Trap: unavailable body` rather than `InvalidMir`, because lowering the projection call produces an undefined region and `Lower.ts` substitutes a trap function. Same root cause, same fix.

## What changed

- **`Diagnostic.ts`** — `SEM0103` (`storedCallableConstructionCode`), the next free code after `SEM0102`. Message names the aggregate, the field path down to the callable (`field parser.decode` for the nested case), and the callable type. `hasInstanceFenceErrors` is the dedicated instance-fence predicate.
- **Realization fence** — `Pipeline.realize` stops target-dependent phases on SEM0103: layout catalog, layout plan, and MIR are all `Unavailable` behind an `AnalysisUnavailable` naming the stored callable, so the source diagnostic is the only reported failure rather than being echoed by `InvalidMir`. `prepare` already rejected on any error diagnostic; both paths now share one `instanceViolationDiagnostics` collector so they cannot drift.
- **`DeclarationIndex.storedCallable`** — a target-independent predicate mirroring what `Layout.layoutType` refuses: struct fields after substitution, fixed-array and slice elements, union members. It deliberately does not descend through references (address-only layout), intrinsic nominals, or Effect types (hidden environments plan callables from concrete identities).
- **`Instances.storedCallableViolations`** — walks each reachable instance's `Construct`/`ArrayConstruct` expressions with the instance's concrete substitution. Provenance follows where the callable was written: a construction whose declared type already stores a callable (`Parser<A> { decode: fn(i32) -> A }`) reports at its own span; one that stores a callable only through substitution (`Holder<T> { value: T }` specialized at a callable) reports at the deterministic earliest specializing call site (ordered by source ID, then position), with the generic body's construction retained as `constructed here` related provenance. `Option.some(i32.add(1))` therefore points at the user's call, not into `silk/option`. Deduplication uses a named structural key over the diagnostic identity and reason facts (the previous key was a presentation-dependent string with a raw NUL delimiter).
- **`packages/language/docs/diagnostics.md`** — regenerated via `pnpm --filter @silk-effect/compiler documentation:generate`.

## Scope: what still compiles

The check is scoped to **reachable instances**, because characterization showed these all work today (verified by returned values, not compilation success):

- callable-bearing structs that are only declared, or named in signatures;
- constructions inside unreachable (never-called) functions — pinned by a three-engine value-asserting test;
- direct callable locals, parameters, partial applications, and higher-order monomorphization (`IndirectCallAcceptance` untouched);
- all stdlib Effect machinery (callable *parameters*, hidden environments).

## Tests

`StoredCallableDiagnostic.test.ts`: the repaired #184 reproducer (both construction sites, exact messages), a realization-fence regression asserting layout catalog, layout plan, and MIR are all `Unavailable` for the reproducer, monomorphic construct+call, generic-inferred wrapper with call-site span and related-provenance assertions, stdlib-inference case asserting the user source ID as primary span and `silk/option` as related provenance, array field + bare array literal, capturing closure field, `once` field, a three-engine (evaluator/Wasm/native) value-asserting acceptance test, and a direct-call regression baseline. Temporary-directory acquisition/cleanup and child execution sit behind scoped Effect boundaries (`Effect.acquireUseRelease` + `Effect.try`), with assertions inside the Effect tests.

`StoredCallableDeterminism.test.ts` + `fixtures/stored-callable-determinism.mjs`: a fresh-process fixture spawned twice, asserting byte-identical stdout and pinning SEM0103 codes, messages, spans, related provenance, ordering (span order), and the realization fence for nested and multiply-specialized constructions.

`pnpm check` run before handoff; result reported in the PR thread.

Refs #184

https://claude.ai/code/session_01G2i2LLZBYb1aiWcKd4MsVs

### DIFF
--- a/packages/compiler/package.json
+++ b/packages/compiler/package.json
@@ -291,6 +291,7 @@
     "effect": "catalog:"
   },
   "devDependencies": {
+    "@effect/platform-node": "4.0.0-beta.103",
     "@effect/vitest": "catalog:",
     "@types/node": "^24.10.1"
   }
--- a/packages/compiler/src/DeclarationIndex.ts
+++ b/packages/compiler/src/DeclarationIndex.ts
@@ -4570,11 +4570,7 @@ export const storedCallable = (
   const next = new Set(seen).add(key)
   for (const field of declaration.fields) {
     if (field.declaredType._tag !== 'Resolved' || field.name._tag !== 'Present') continue
-    const found = storedCallable(
-      self,
-      Type.substitute(field.declaredType.type, substitution),
-      next,
-    )
+    const found = storedCallable(self, Type.substitute(field.declaredType.type, substitution), next)
     if (found !== undefined)
       return Object.freeze({
         path: Object.freeze([field.name.spelling, ...found.path]),
--- a/packages/compiler/src/DeclarationIndex.ts
+++ b/packages/compiler/src/DeclarationIndex.ts
@@ -4520,3 +4520,66 @@ export const containsLexicalBorrow = (
       containsLexicalBorrow(self, Type.substitute(field.declaredType.type, substitution), next),
   )
 }
+
+/** One stored bare-callable occurrence that denies an aggregate type a target layout. */
+export interface StoredCallable {
+  /** Field names from the aggregate down to the callable-typed position; empty for elements. */
+  readonly path: ReadonlyArray<string>
+  readonly callable: Type.Callable
+}
+
+/**
+ * Finds the first stored position at which a concrete type keeps a bare callable value.
+ *
+ * The walk mirrors what nominal layout planning can actually see: struct fields after
+ * substitution, fixed-array and slice elements, and union members. It deliberately does not
+ * descend through references (their layout is an address regardless of the target), intrinsic
+ * nominals (their layouts never depend on their type arguments), or Effect types (their hidden
+ * environments plan callables from concrete identities). A position this finds is exactly one
+ * `Layout.layoutType` would refuse with `callable environment layout is planned from its hidden
+ * concrete identity`, so a construction of the enclosing aggregate cannot receive a layout.
+ */
+export const storedCallable = (
+  self: Index,
+  type: Type.Type,
+  seen: ReadonlySet<string> = new Set(),
+): StoredCallable | undefined => {
+  if (Type.isCallable(type)) return Object.freeze({ path: Object.freeze([]), callable: type })
+  if (Type.isFixedArray(type) || Type.isSlice(type)) return storedCallable(self, type.element, seen)
+  if (Type.isUnion(type)) {
+    for (const member of type.members) {
+      const found = storedCallable(self, member, seen)
+      if (found !== undefined) return found
+    }
+    return undefined
+  }
+  if (!Type.isNominal(type) || Type.isIntrinsicNominal(type)) return undefined
+  const key = Type.key(type)
+  if (seen.has(key)) return undefined
+  const declaration = byCanonical(self, {
+    _tag: 'CanonicalDeclarationId',
+    module: type.module,
+    name: type.name,
+  })
+  if (declaration?._tag !== 'StructDeclaration') return undefined
+  const substitution =
+    Type.substitution(
+      declaration.typeParameters.map((parameter) => parameter.type),
+      type.arguments,
+    ) ?? new Map()
+  const next = new Set(seen).add(key)
+  for (const field of declaration.fields) {
+    if (field.declaredType._tag !== 'Resolved' || field.name._tag !== 'Present') continue
+    const found = storedCallable(
+      self,
+      Type.substitute(field.declaredType.type, substitution),
+      next,
+    )
+    if (found !== undefined)
+      return Object.freeze({
+        path: Object.freeze([field.name.spelling, ...found.path]),
+        callable: found.callable,
+      })
+  }
+  return undefined
+}
--- a/packages/compiler/src/Diagnostic.ts
+++ b/packages/compiler/src/Diagnostic.ts
@@ -790,6 +790,17 @@ export const hasGenericSpecializationErrors = (diagnostics: ReadonlyArray<Diagno
       diagnostic.code === polymorphicRecursionCode,
   )
 
+/**
+ * Tests whether a reachable-instance fence diagnostic denies target realization.
+ *
+ * A stored-callable construction (#184) is exactly a program the layout planner cannot serve: the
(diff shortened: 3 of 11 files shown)


julia-script:
`pnpm check` is green locally at b8b836c: biome, full workspace build, typecheck, all package test suites, and `test:scripts` all pass (exit 0). The native-acceptance suite ran against `/usr/bin/clang` 18.1.3.

---

julia-script:
Two-axis review found blocking revisions before this can serve as the prerequisite fence for #187.

## Spec

1. **[P1] SEM0103 does not actually stop realization before layout/MIR.**
   Pipeline.realize merges the diagnostic, but the subsequent layout/MIR gate only checks hasGenericSpecializationErrors, which does not include SEM0103. Please make stored-callable violations realization-blocking (or add a dedicated instance-fence predicate), and add a regression asserting layout/MIR are Unavailable for the reproducer.

2. **[P2] Generic-specialization provenance points into standard-library source.**
   The Option.some fixture currently locks the primary span to silk/option. Please use the deterministic specializing call site as the primary origin and retain the generic body construction as related provenance; assert the user source ID.

3. **[P2] Fresh-process determinism is not covered.**
   Add a fresh-process fixture comparing SEM0103 code, message, span, and ordering for nested and multiply-specialized constructions.

The PR does not currently auto-close #184, so no closing-metadata change is required. The opening phrase is still worth clarifying because this is the diagnostic prerequisite, not full callable storage.

## Standards

4. **[P2] Tests cross throwing platform boundaries directly.**
   StoredCallableDiagnostic.test.ts uses mkdtempSync, rmSync, spawnSync, and module-global afterAll cleanup. AGENTS.md requires Effect-wrapped external boundaries and scoped resources. Move temporary-directory acquisition/cleanup and child execution behind scoped Effect boundaries, with assertions inside the Effect test.

## Non-blocking design cleanups

- Pipeline.ts maintains the same instance-diagnostic collector sequence in two places; consider extracting one shared collector.
- Instances.ts deduplicates through a presentation-dependent string with a NUL delimiter; prefer a named structural key based on stable diagnostic identity and span.

Please update the PR and report the focused tests plus pnpm check result.

julia-script:
All four blocking revisions are in at aaa1a99; the PR body is updated to match, including the opening rephrased as the diagnostic prerequisite.

1. **[P1] Realization fence** — `Pipeline.realize` stops target-dependent phases on SEM0103 behind a dedicated `Diagnostic.hasInstanceFenceErrors` predicate: layout catalog, layout plan, and MIR are all `Unavailable`, so the source diagnostic is the only reported failure. Regression test `fences realization: SEM0103 leaves layout and MIR unavailable, not InvalidMir` asserts all three for the #184 reproducer, and the determinism fixture pins the fence in a fresh process as well.
2. **[P2] Provenance** — a violation that exists only through substitution reports at the deterministic earliest specializing call site (ordered by source ID, then position), retaining the generic body's construction as `constructed here` related provenance. The stdlib fixture now asserts the primary span's source ID is the user module and `silk/option` appears only as related provenance; the generic-wrapper test asserts the exact call-site span text.
3. **[P2] Determinism** — `StoredCallableDeterminism.test.ts` spawns `fixtures/stored-callable-determinism.mjs` twice and asserts byte-identical stdout plus SEM0103 codes, messages, spans, related provenance, and span ordering for nested and multiply-specialized (two distinct `Holder<T>` instantiations) constructions.
4. **[P2] Standards** — temporary-directory acquisition/cleanup and child execution are behind scoped Effect boundaries (`Effect.acquireUseRelease` + `Effect.try`, release ignored-on-failure), assertions inside the Effect tests; the module-global `afterAll` is gone.

Both non-blocking cleanups are also in: `realize`/`prepare` share one `instanceViolationDiagnostics` collector, and deduplication uses a named structural key over the diagnostic identity and reason facts. (The old key's NUL delimiter was a raw byte in the source, which is why grep treated `Instances.ts` as binary — the file is plain UTF-8 again.)

**Focused tests**: `StoredCallableDiagnostic.test.ts` (10) + `StoredCallableDeterminism.test.ts` (1) — 11/11 pass. Neighbor suites `IndirectCallAcceptance`, `StructValues`, `CallableDeterminism`, `CallableSemanticsDeterminism`, `Diagnostic` — 25/25 pass. `documentation:check` clean.

**`pnpm check`**: green end to end at aaa1a99 (exit 0) — biome, full workspace build, typecheck, all package test suites, and `test:scripts`.

---

julia-script:
Final re-review: Spec passes, but Standards has one remaining blocking revision.

## Standards

- **[P2] Raw Node output remains outside Effect** in packages/compiler/test/fixtures/stored-callable-determinism.mjs. The fixture calls process.stdout.write after its Effect has completed, so output failures bypass the Effect boundary required by AGENTS.md. Please keep analysis, rendering, and output in one Effect program using effect/Console or the appropriate platform terminal service, with the single runtime call only at the executable edge.

## Non-blocking cleanup

- StoredCallableDeterminism.test.ts and StoredCallableDiagnostic.test.ts duplicate the same stream collector and scoped child-process lifecycle. Please extract one test-support process runner if this remains a small mechanical change.

After updating, rerun both stored-callable suites, typecheck, Biome, and the relevant determinism fixture.

PR #208, opened 2026-08-15
#190 PR4b — Stored Effect native and Wasm parity

julia-script:
Implements OpenSpec task 4.2 of `store-effects-in-nominals`: native LLVM and direct-Wasm parity for stored Effects. Base is PR4a merge 298c1dea.

## What was wrong

Both backends died on `RangeError: could not realize place-read lane 0` for any program storing an Effect in a nominal field.

`logicalLanes` in each backend had a stored-*callable* branch routing through `Layout.callingShape`, but the `EffectValue` branch ignored `type.storage` and fell through to `effectEnvironmentLanes`, which emits lane paths with **no `EffectCaptureSelector` prefix**. The place-read matcher requires a source lane whose path extends the destination's by the selectors, so every candidate was rejected.

`Mir.verify` compares lane *counts*, never lane *paths*, so the mismatch was invisible to verification — the MIR verified clean while both backends failed.

## The fix

11 lines: a stored-Effect branch in each backend, mirroring the adjacent stored-callable one, routing through the existing generic `Layout.callingShape`. `shapeNode` already accepted `StoredEffectEnvironment` and `materializeLanes` already emitted `EffectCaptureSelector` — the machinery existed and was simply unused by the backends. No new representation, identity, or dispatch model.

## Evidence

`StoredEffectEngineParity.test.ts` mirrors the merged evaluator matrix on both backends:

- shared / exclusive / consuming / provided — result and access parity
- exact failure and requirement rows asserted against the realization on both the Wasm and native lowerings
- `provided` pinned to its specialized runner and single provider, proving service provision resolves statically into emitted IR
- unrun and typed-failure cleanup — exactly-once Drop hook and allocation balance
- **exactly-once cleanup proven on the Wasm engine itself**: the unrun construct-and-clean cycle repeats, and the heap stays flat as the cycle count grows tenfold (verified flat to 4000 cycles). Returning 42 alone would not distinguish a leaked capture.
- suspending resume
- static runner identity asserted in emitted LLVM IR, and equal across both targets
- `call_indirect` absent from every Wasm artifact (no dynamic dispatch)

Verified the tests genuinely fail with the original `place-read lane 0` error when the backend fix is reverted, and that the heap assertion is falsifiable.

## Scope

SEM0107 is **unchanged**. It still fences these programs out of `Analysis.realize`, so the harness drives the lowering phases directly, exactly as the merged evaluator test does. This is deliberate: the spec requires the fence to remain "for any nominal storage path not proven through ownership, layout, MIR, evaluator, LLVM, and direct WebAssembly", and narrowing it is task 4.4.

Consequence, recorded in tasks.md: direct Wasm executes end to end, while native parity is proven at emitted LLVM IR, because `Driver.compile` cannot reach a native executable until 4.4 lands.

Not included: 4.3 invalidation, 4.4 fence narrowing, 5.1–5.3, sync, archive.

## Known follow-up (pre-existing, not introduced here)

`WasmBackend.ts` `resultLanes` (~4777) and the declaration lane path (~5320) lack the stored branch for **both** Effects and callables — missing since #196. No shape in scope reaches them, and the asymmetry affects stored callables equally, so fixing it speculatively would be unverifiable. Flagged for follow-up when 4.4 allows stored executables across function boundaries.

## Gates

- `pnpm typecheck` — 20/20 pass
- `pnpm exec biome check .` — clean
- compiler suite — 1849/1850 pass; the single failure (`WasmShadowStackHeapCollision`, deep-recursion stack probe) **reproduces identically at untouched base 298c1dea**, so it predates this change
- stored-callable, stored-Effect, and heap-reclaim suites — 27/27 pass, no regression on the shared lane path
- `release:candidate` not required: no package contents or exports changed

### DIFF
--- a/openspec/changes/store-effects-in-nominals/tasks.md
+++ b/openspec/changes/store-effects-in-nominals/tasks.md
@@ -19,10 +19,7 @@
 ## 4. Engine Parity and Invalidation
 
 - [x] 4.1 Execute shared, exclusive, consuming, unrun, failing, and suspending stored Effects in the evaluator.
-- [x] 4.2 Add native LLVM and direct-Wasm parity for results, failures, runner identity, and cleanup traces.
-  - Direct Wasm executes each shape end to end; native parity is proven at emitted LLVM IR because
-    `SEM0107` still fences these programs out of `Analysis.realize`, so `Driver.compile` cannot
-    reach a native executable until task 4.4 narrows the fence.
+- [ ] 4.2 Add native LLVM and direct-Wasm parity for results, failures, runner identity, and cleanup traces.
 - [ ] 4.3 Add capture-shape, target, access, cleanup, and suspendability invalidation fixtures.
 - [ ] 4.4 Narrow the unavailable-Effect-layout fence only for shapes proven by all engines.
 
--- a/openspec/changes/store-effects-in-nominals/tasks.md
+++ b/openspec/changes/store-effects-in-nominals/tasks.md
@@ -19,7 +19,7 @@
 ## 4. Engine Parity and Invalidation
 
 - [x] 4.1 Execute shared, exclusive, consuming, unrun, failing, and suspending stored Effects in the evaluator.
-- [x] 4.2 Add native LLVM and direct-Wasm parity for results, failures, runner identity, and cleanup traces.
+- [ ] 4.2 Add native LLVM and direct-Wasm parity for results, failures, runner identity, and cleanup traces.
 - [ ] 4.3 Add capture-shape, target, access, cleanup, and suspendability invalidation fixtures.
 - [ ] 4.4 Narrow the unavailable-Effect-layout fence only for shapes proven by all engines.
 
--- a/openspec/changes/store-effects-in-nominals/tasks.md
+++ b/openspec/changes/store-effects-in-nominals/tasks.md
@@ -19,7 +19,10 @@
 ## 4. Engine Parity and Invalidation
 
 - [x] 4.1 Execute shared, exclusive, consuming, unrun, failing, and suspending stored Effects in the evaluator.
-- [ ] 4.2 Add native LLVM and direct-Wasm parity for results, failures, runner identity, and cleanup traces.
+- [x] 4.2 Add native LLVM and direct-Wasm parity for results, failures, runner identity, and cleanup traces.
+  - Direct Wasm executes each shape end to end; native parity is proven at emitted LLVM IR because
+    `SEM0107` still fences these programs out of `Analysis.realize`, so `Driver.compile` cannot
+    reach a native executable until task 4.4 narrows the fence.
 - [ ] 4.3 Add capture-shape, target, access, cleanup, and suspendability invalidation fixtures.
 - [ ] 4.4 Narrow the unavailable-Effect-layout fence only for shapes proven by all engines.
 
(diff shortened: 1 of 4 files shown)


julia-script:
## CHANGES REQUESTED — review at exact SHA `bc86972c79b6136c612dfadbe668e5090d8de88e`

Base: `298c1dea92b9190d5b49e817e96e8b2e95236cce`.

### Standards

PASS — no documented-standard violations. Non-blocking judgement calls only:

- `Backend.ts:984-1008` and `WasmBackend.ts:128-150` duplicate the backend-neutral `Mir.Type → CallingLane` decision.
- `StoredEffectEngineParity.test.ts:307-342,365-387` duplicate cleanup fixture construction.
- `StoredEffectEngineParity.test.ts:168-169` should type `traceCount` with the closed `TraceEvent` tag union rather than `string`.

### Spec blockers

1. **DOWNSTREAM-BLOCKER — Native behavior is never executed.**

   `StoredEffectEngineParity.test.ts:255-295,412-420,489-497` only prove that LLVM IR defines `silk_main` and contains runner/drop symbols. Task 4.2 and the cross-engine cleanup/suspension requirements require executed native results, failures, resumption, runner identity, and cleanup. The explanatory note added to `tasks.md` documents the gap rather than satisfying it. SEM0107 does not prevent taking the already manually lowered LLVM artifact through `NativeToolchain` emit/link/run. Add native execution for the runtime matrix, cleanup exits, and suspension before keeping task 4.2 checked.

2. **MERGE-BLOCKER — Wasm exact-once cleanup is missing for failure and suspension.**

   The failure and suspension paths at `StoredEffectEngineParity.test.ts:406-410,483-487` assert only result/no indirect dispatch. Heap-flat repetition at `:501-532` covers only unrun and cannot detect double cleanup. Add externally observable, per-exit exactly-once cleanup evidence for failing and suspending Wasm runs.

3. **MERGE-BLOCKER — Direct-Wasm runner identity is not asserted.**

   `StoredEffectEngineParity.test.ts:246-253` proves output and absence of `call_indirect`, but not that the correct concrete runner was the direct target. Assert the exact emitted direct call target using named/debug WAT or binary call-index-to-symbol evidence. The exact-symbol assertion currently exists only for LLVM.

No scope creep found. SEM0107 remains asserted, tasks 4.3/4.4 remain unchecked, and no dynamic dispatch, dictionary, existential, or standalone Effect ABI was introduced.

Please batch these three evidence fixes, freeze the revised branch, and request delta re-review at the new exact SHA.

_Note: GitHub does not allow the authenticated PR owner to submit a formal REQUEST_CHANGES review on its own PR, so this is posted as the blocking review comment._

julia-script:
Final blocker closed at exact head `effc199024f624888147a99ad02c6ab5ae8c8fa2`.

The native cleanup witness now performs a checked out-of-bounds access, asserts the emitted LLVM Drop call, and accepts only the LLVM trap process signal (`SIGILL`/`SIGTRAP`). Ordinary exits (including OOM recovery 0/1/42), spawn/build failures, unrelated typed failures, and defects no longer count as Drop evidence. Paired with the poison run, this causally pins exactly one native Drop for success, unrun, failure, and suspension.

Gates:
- StoredEffectEngineParity: 4/4 pass
- compiler typecheck: pass
- targeted Biome: pass
- exact-head GitHub `webcontainer-browser`: pass
- exact-head GitHub `validate` (`pnpm check` + `release:candidate`): pass

PR is CLEAN/MERGEABLE; no compiler implementation, SEM0107 fence, non-goal, or later OpenSpec task changed.

PR #231, opened 2026-08-21
feat(compiler): extract shared helpers — ByteClass, Align, Graph, canonicalNaN (partial)

julia-script:
## OpenSpec: extract-compiler-shared-helpers

Wave A, part 2 of the compiler deep-review orchestration. **Partial — ~60% complete.**

### Done so far
- ✅ `internal/ByteClass.ts` — unified ASCII byte classifiers, routed Lexer, LiteralForm, IntegerLiteral
- ✅ `internal/Align.ts` — one `alignUp`, routed Layout, Backend, WasmBackend, CoroutineFrame
- ✅ `internal/Graph.ts` — parameterized Tarjan SCC (file created; caller routing pending)
- ✅ `FloatingPoint.canonicalNaN()` — routed Transcendental, fromNumber, squareRoot

### Still needed
- ImportPath.spelling/canonicalTarget (ModuleSummary, ImportPlan, ModuleClosure, NameResolution)
- internal/Escape.ts (Lexer/LiteralForm escape vocabulary)
- Type.encodeRequirement → Presentation
- PhaseReport.measure → Pipeline/Driver
- Backend.suspensionPointKey → WasmBackend
- Graph.ts caller routing (ModuleClosure, DeclarationIndex, OpaqueRealization)

### Verification (what's done so far)
- `pnpm typecheck` ✓
- `pnpm exec biome check .` ✓

### Dependency
Wave A part 2 — requires remove-compiler-dead-code (#230). Unblocks: split-compiler-frontend, split-compiler-type-system, split-compiler-layout, split-compiler-declaration-index, split-compiler-backends.

PR #235, opened 2026-08-21
feat(compiler): split frontend — keyword unification

julia-script:
## OpenSpec: split-compiler-frontend

Wave C. Partial — keyword unification done.

### Done
- ✅ Moved fn/let/move/pub/return/import into keywordSpellings table
- ✅ Deleted manual per-byte keywordKind block
- ✅ Lexer tests: 54/54 pass

### Still needed
- Extract Parser/*.ts sub-modules (Expression, Type, Statement, Declaration, Import)
- internal/ParseState.ts
- Grammar seam merges
- Trivia-skipping peek(n)

### Depends on
- #231 (extract-compiler-shared-helpers — ByteClass)


julia-script:
Superseded by the unified PR.

PR #236, opened 2026-08-21
feat(compiler): split type system (starter)

julia-script:
Wave C starter. See openspec/changes/split-compiler-type-system/


julia-script:
Superseded by the unified PR.

PR #237, opened 2026-08-21
feat(compiler): split mid-end (starter)

julia-script:
Wave C starter. See openspec/changes/split-compiler-mid-end/


julia-script:
Superseded by the unified PR.

PR #241, opened 2026-08-21
feat(compiler): split backends (starter)

julia-script:
Wave C starter. See openspec/changes/split-compiler-backends/


julia-script:
Superseded by the unified PR.

PR #259, opened 2026-08-25
refactor: remove nested ternaries

julia-script:
## Summary

- replace every nested ternary across compiler, tests, tooling, and supporting packages with explicit control flow
- use block-bodied callbacks, guard clauses, switches, typed assignments, and reusable semantic helpers without introducing IIFEs
- enable Biome style/noNestedTernary as an error to prevent regressions
- harden diagnostic documentation scanning so return text inside literals is not mistaken for a return statement

## Validation

- pnpm typecheck
- pnpm exec biome check .
- pnpm check
- pnpm release:candidate
- whole-repository style/noNestedTernary scan: zero violations
- added-IIFE audit: zero markers


julia-script:
Follow-up diagnosis: the apparent Vitest hangs were a synchronous parser loop introduced while flattening parseBlock dispatch. ReturnKeyword was omitted from parseBlockChild, so return statements fell through to expression parsing without advancing state. Fixed in ae9a326.

Focused validation:
- minimal Analysis.ofSourceRealized harness: 26 ms
- Parser.test.ts: 111/111 passed
- FlowModel.test.ts: 8/8 passed
- branch-wide else-block nested-if audit: zero
- Parser/Statement.ts Biome check: passed

PR #307, opened 2026-09-01
perf(compiler): memoize canonical keys and MIR verification

julia-script:
## Summary

Profiling the slowest tests (SchedulerFiber, ModuleVerification) showed the time was not in evaluation — the interpreter itself was ~1s. ~50% of CPU went to uncached canonical-key string builders, and `MirVerification.verify` ran **twice** per test (tests assert it, then `evaluate` re-verifies the same module) with an O(functions² × rounds) fixed point inside it.

## Changes

- **`Instances.keyText`, `Type.genericArgumentKey`, `Specialization.key`** — WeakMap-memoized per object identity (`Type.key` already was). String-typed inputs bypass the cache.
- **`MirVerification.verify`** — cached per `Mir.Module`. Verification is pure over an immutable module; callers that need fresh verification (e.g. the malformed-module tests) build fresh module objects, so per-identity caching is safe.
- **`originReachableSuspensionFunctions`** — index functions by declaration instead of scanning every function per call target per fixed-point round. `matchesInstance` still checks module/name exactly, so the index is a pure prefilter.
- **`SuspensionMode.pathTo`** — drop the per-expansion full-queue sort (it re-sorted the whole BFS queue with a `join('\0')` comparator after every node). FIFO with sorted neighbor expansion dequeues paths in the identical (length, lexicographic) order.

`evaluate` still verifies internally — tests depend on the `Blocked`/`InvalidMir` outcome — the cache just makes the duplicate run free.

## Measured impact (isolated runs, M-series laptop)

| Test | Before | After |
|---|---|---|
| SchedulerFiber › suspends timer Fibers… (suite's worst) | 203s | ~18s |
| SchedulerFiber.test.ts (whole file, 14 tests) | ~25min ranked | 135s |
| ModuleVerification › emits corpus modules… | 4m43s ranked | ~56s |
| timer-basic fixture: verify / evaluate / codegenWasm | 7.1s / 7.0s / 7.0s | 0.23s / 0.11s / 0.20s |

Wins apply suite-wide: every test that compiles a snapshot goes through these paths (the Wasm backend also leans on `keyText`).

## Verification

- SchedulerFiber: 14/14 pass
- BootstrapEvaluation, SlotCopy, OwnedAllocationAcceptance, ModuleVerification, SuspensionMir, Instances: 82/82 pass (includes the malformed-MIR / `InvalidMir` paths that exercise fresh-module verification)
- biome clean, `tsc` (src + test configs) clean


julia-script:
Second commit (470c9ce) extends this beyond the frontend:

**Compiler realization** — `executionForInstance` now indexes executions per module (was a linear scan rebuilding a fresh InstanceKey + key string per candidate per lookup), `lookupMember` indexes members by spelling, `reachesExecutionOwner` indexes calls by owner and shares its visited set (the per-path copy made the walk super-linear), `contentDigest` memoizes per byte-array. Big wasm-target realization ~16s → ~9.7s.

**LLVM + Wasm builders** (the whole cost of native-path tests — clang itself is ~1s): per-instruction ops were traced `Effect.fn` (a span per emitted instruction, ~40% of the backend phase) and every mutation took a semaphore permit + `Effect.fiberId` yield. Ops are now `Effect.fnUntraced`; `mutate`/`snapshot` run as one suspended step (transitions are synchronous `Result` computations, so atomicity is unchanged); the cross-fiber ownership rejection is preserved via `Effect.withFiber` and its pinned test still passes. Backend emission for the temp-dir fixture: 24.7s → ~12s.

**docgen** — the stdlib doctest sweep ran 3× across `Stdlib.test.ts`; the two tests needing the live report share one run (file 90s → 52s).

Suites: compiler spot files + verify-sensitive set, llvm 72/72, wasm 86/86 — all green.

PR #401, opened 2026-09-10
feat(stdlib): add lossless RFC 3986 URI handling

julia-script:
Adds RFC 3986 URI handling with allocation-free borrowed parsing, explicit ownership, reusable resolution and percent coding, and construction from raw or encoded components.

Closes https://linear.app/juliaortiz/issue/JUL-164/add-rfc-3986-uri-parsing-resolution-and-lossless-formatting

`Uri` requires a scheme; `UriReference` also accepts relative references. Both retain input lifetimes and validated ranges. Parsing preserves spelling, empty delimiters, unsplit userinfo, lexical ports, and strict host grammar. `copy`/`parseOwned` create independent storage; owned `fromString` constructors adopt existing storage.

The review changes replace intermediate resolution buffers with one caller-owned `Bytes`, compact dot segments in place, and adopt owned output. Percent coding supports reused buffers, a Writer, in-place decoding, and a borrowed unchanged-input fast path; owned percent output reserves its exact size once. `UriComponents` supports modification, tagged raw/encoded data, and full, authentication-free, or path/query serialization. The compiler adds only a general checked borrowed subslice primitive, exposed by `silk.slice`.

Usage examples (checked together with `silk check`):

```silk
import silk.uri { Uri, OwnedUri }
import silk.uri_reference { UriReference, ParseError }
import silk.uri_percent { UriPercent, DecodeError }
import silk.uri_percent { Component as PercentComponent }
import silk.uri_components { UriComponents, ComponentValue, Serialization }
import silk.allocator { Allocator, OutOfMemoryError }
import silk.result { Result }
import silk.string { String }
import silk.bytes { Bytes }
```

Inspect existing text with no allocator requirement or input copy:

```silk
fn inspectUri() -> bool {
  return match move Uri.parse("https://EXAMPLE.com/a%2f?+#") {
    Result<Uri<'static>, ParseError>.Failure { error } => false
    Result<Uri<'static>, ParseError>.Success { value } =>
      Uri.path(&value) == "/a%2f" && Uri.format(&value) == "https://EXAMPLE.com/a%2f?+#"
  }
}
```

Reuse output for redirect resolution. With base `http://a/b/c/d;p?q` and location `../g`, the output is `http://a/b/g`. Existing sufficient capacity requires no allocation:

```silk
effect fn resolveLocation<'text>(output: &mut Bytes, base: &Uri, location: string<'text>)
-> Result<(), ParseError> ! OutOfMemoryError ? &mut Allocator {
  return match move UriReference.parse(location) {
    Result<UriReference<'text>, ParseError>.Failure { error } => Result.failResult<(), ParseError>(error)
    Result<UriReference<'text>, ParseError>.Success { value } => run Uri.resolveInto(&mut output, base, &value)
  }
}
```

Transfer a String into an owned URI without a second allocation or copy:

```silk
effect fn ownUri(text: string) -> Result<OwnedUri, ParseError> ! OutOfMemoryError ? &mut Allocator {
  let ownedText = run String.copy(text)
  return OwnedUri.fromString(move ownedText)
}
```

Encode raw segment data and decode arbitrary bytes in the same buffer:

```silk
effect fn percentCoding(output: &mut Bytes) -> Result<(), DecodeError>
! OutOfMemoryError ? &mut Allocator {
  let encoded = run UriPercent.encodeInto(&mut output, String.utf8Bytes("a/b%2f"), PercentComponent.PathSegment)
  // output contains "a%2Fb%252f"; decoding reuses its capacity.
  // The result bytes are FF 2B 00: '+' stays '+', and FF need not be UTF-8.
  return run UriPercent.decodeInto(&mut output, "%FF+%00")
}
```

Modify a path and omit credentials. For `https://user:secret@example.com/old?q#f`, this emits `https://example.com/a%20b?q#f`:

```silk
effect fn withoutAuthentication<'text>(reference: &UriReference<'text>)
-> Result<String, ParseError> ! OutOfMemoryError ? &mut Allocator {
  let mut parts = UriComponents.fromReference(reference)
  parts.path = ComponentValue<'text>.Raw { bytes: String.utf8Bytes("/a b") }
  return run UriComponents.serializeOwned(&parts, Serialization.WithoutAuthentication)
}
```

Syntax errors are `Result` values with reason, component, and byte offset. Allocation errors use the separate `OutOfMemoryError` Effect channel. Decoding preserves plus signs and arbitrary bytes; validate UTF-8 separately with `String.fromUtf8`.

Resolution retains literal RFC recomposition: resolving `/a/..//g` against `x:a` introduces authority `g`; `/a/..//g:h` returns an invalid-port failure. DNS, IDNA, WHATWG URL, TLS identity, and scheme policy remain outside this syntax API.

OpenSpec: `openspec/changes/add-rfc3986-uri`.

Verification completed:

- `pnpm typecheck`, `pnpm format:check`, and `pnpm lint` passed.
- Focused URI and slice native acceptance passed, including all 42 RFC resolution vectors, allocation contracts, component mutation, and Writer output.
- Focused ownership/MIR tests, intrinsic inventory, the cleanup regression, and the updated Bytes hover expectation passed.
- Standard-library documentation policy passed for 100 modules; all 57 doctests passed.
- OpenSpec strict validation and the PR usage examples passed.

The full `pnpm test` run is incomplete. Earlier runs exposed a stale hover expectation and a missing `sliceView` call in the intrinsic catalog coverage fixture; both are corrected and their focused tests pass. A 60-second timeout in an unchanged compiler architecture test passed in isolation in 17 seconds. Changes were pushed before subsequent local tests as requested; the complete gate sequence is now running with `VITEST_MAX_WORKERS=2 TURBO_ENV_MODE=loose`. `pnpm check` and `pnpm release:candidate` remain pending. No correctness assertion or timeout was weakened.

Independent production review approved the URI changes and the targeted generic cleanup fix. Dedicated test-economics review approved the consolidated suite: one URI corpus program, three distinct slice execution outcomes, and cheaper ownership/MIR assertions. Redundant allocation helpers were folded into existing reuse cases.

Equivalent base/branch measurements against `d778a935`: affected compiler files 249 ms / 2.33 s; intrinsic inventory 13 ms / 14 ms; URI plus slice native cases 3 ms / 33.78 s. Combined measured incremental test execution: 35.859 seconds. Native wall times were 3.28 s / 38.85 s.

The focused catalog pair also passed on both revisions: base 34.62 s tests / 37.66 s wall; branch 27.46 s / 31.23 s. This is not claimed as a speedup. Test-economics approval covers the exact committed diff through `b54532130634a1ad75150cb0df2b1755c4e75d0b`.


packages/compiler/stdlib/silk/uri_reference.silk
```diff
@@ -0,0 +1,605 @@
...
+    let bytes = String.utf8Bytes(text)
+    let length = bytes.length
+    if length == 0 {
+      return Result.succeed<UriReference, ParseError>(
+        UriReference {text: Option.none<RawBuffer<u8>>(), length: length, parts: move parts},
+      )
+    }
+    let allocation = run Allocator.allocate(Layout {bytes: length, alignment: 1})
+    let mut owned = RawBuffer.from<u8>(move allocation, length)
+    let copied = RawBuffer.copy<u8>(&mut owned, 0, bytes, length)
```
julia-script: [P2] Expose borrowed parsing instead of requiring an input copy

Every successful nonempty parse allocates and copies the complete input here. Consumers inspecting an existing request buffer, parsing a literal, or parsing a String they already own cannot avoid another allocation and copy. The borrowed accessors only remove allocations after that initial copy.

Zig's parse returns components borrowing its input. Please expose a lifetime-bound borrowed parsed representation backed by the existing validator/ranges, with an explicit conversion to owned storage when needed. An ownership-taking constructor would also let callers transfer existing storage without recopying it. Update the owned-only OpenSpec decision with this API change.

packages/compiler/stdlib/silk/uri.silk
```diff
@@ -0,0 +1,304 @@
...
+        } else {
+          let merged = run merge(base, referencePath)
+          let normalized = run removeDots(String.view(&merged))
+          let wrotePath = run String.append(&mut output, String.view(&normalized))
+        }
+      }
+    }
+    let wroteQuery = run appendOptional(&mut output, move selectedQuery, "?")
+    let wroteFragment = run appendOptional(&mut output, UriReference.fragment(reference), "#")
+    return run Uri.parse(String.view(&output))
```
julia-script: [P2] Resolve into reusable storage and avoid copying the completed serialization

For an ordinary nonempty relative merge such as ../g, this implementation creates six separately backed buffers: merge's Vector, its copied String, removeDots' Vector, its copied String, the output String, and the RawBuffer created by this final parse. That excludes parsing the inputs and additional allocations during geometric growth. This is a source-level allocation analysis, not a benchmark measurement.

Please provide caller-supplied output/scratch storage, normalize directly in that storage, and validate/adopt completed owned output without copying it again. Validation can remain shared and strict; avoiding the copy does not require a public unchecked constructor. Zig's resolveInPlace uses caller scratch and borrows unchanged components.

packages/compiler/stdlib/silk/uri_percent.silk
```diff
@@ -0,0 +1,182 @@
...
+        if digit(source[index + usize.ONE]) > u8.toU8(15) || digit(source[index + 2]) > u8.toU8(15) {
+          return Result.failResult<Bytes, DecodeError>(DecodeError {offset: index})
+        }
+        index = index + 3
+      } else {
+        index = index + usize.ONE
+      }
+      length = length + usize.ONE
+    }
+    let mut output = run Bytes.zeroed(length)
```
julia-script: [P2] Make percent coding usable with caller-owned buffers or a writer

Both public percent-coding operations force fresh owned output. Even decoding nonempty text with no escapes allocates and copies, and callers cannot reuse scratch storage or stream encoding into an existing destination.

There is also avoidable internal allocation work: decode computes the exact output length, then Bytes.zeroed grows a Vector through repeated appends and initializes bytes that are immediately overwritten. encode likewise builds its String one byte or escape at a time.

Please expose destination-buffer/writer operations (and an in-place decoder where lifetime/aliasing rules permit), with owned convenience functions allocating once from the calculated size. A borrowing fast path can avoid copying unchanged decoded input. Zig provides writer-based encoding, buffered/in-place decoding, and toRaw/toRawMaybeAlloc fast paths. Keep Silk's strict malformed-escape errors and arbitrary-byte output semantics.

julia-script (commented):
Reviewed against Zig's current lib/std/Uri.zig, focusing on allocations and API completeness.

I recommend addressing the three inline allocation requests before merging. The strict validator and lossless component representation are useful; the main design change is to make borrowed parsing and caller-controlled output storage the core, with explicit owned convenience operations.

Completeness relative to Zig:
- No construction or modification from components: callers must serialize manually and parse again.
- No tagged raw versus already-percent-encoded component representation.
- No selective serialization, such as omitting authentication or emitting path/query.
- No whole-path or unreserved-only percent-encoding context comparable to Zig's formatPath and formatEscaped.

These are scope/design omissions rather than violations of the current OpenSpec. Please revise the owned-only design/spec alongside the API changes. Lexical ports, unsplit userinfo, strict grammar, lossless empty delimiters, and excluding DNS/IDNA/WHATWG policy are deliberate boundaries worth retaining.

One semantic difference to retain as an explicit decision: Silk reparses recomposed output, so dot removal can introduce an authority or return a syntax failure from two valid inputs. This is already specified and tested.

Validation: the focused native URI acceptance command passed (2 tests passed, 329 unrelated tests skipped), including the corpus program containing all 42 RFC 3986 section 5.4 resolution examples. I did not find a confirmed RFC parsing defect. Full workspace checks were not rerun, and no source changes were made.

Reference: https://codeberg.org/ziglang/zig/src/branch/master/lib/std/Uri.zig

GitHub does not permit the PR author to submit a formal REQUEST_CHANGES review, so these change requests are posted as a COMMENT review.

packages/compiler/stdlib/silk/uri_reference.silk
julia-script: Implemented in 4e1ffde9b6375033c877ca7e11cd4eb5d7721504. Uri and UriReference now borrow their input with explicit lifetimes and parse synchronously without an Allocator. copy/parseOwned create independent storage; OwnedUri/OwnedUriReference.fromString validate and adopt an existing String. Borrowed views retain source lifetimes. Ownership tests reject escaping views and mutation of live backing bytes. The OpenSpec and PR examples now describe this API.

packages/compiler/stdlib/silk/uri.silk
julia-script: Implemented in 4e1ffde9b6375033c877ca7e11cd4eb5d7721504. resolveInto writes directly into caller-owned Bytes and compacts dot segments in that buffer; warmed capacity needs no allocation. resolveOwned reserves one upper bound and adopts the completed String without copying. All 42 RFC resolution examples, literal recomposition boundaries, and rejecting-allocator reuse checks pass in the focused native corpus.

packages/compiler/stdlib/silk/uri_percent.silk
julia-script: Implemented in 4e1ffde9b6375033c877ca7e11cd4eb5d7721504. Percent coding now has encodeInto/decodeInto, encodeToWriter, strict decodeInPlace, and decodeOrBorrow. Owned conveniences reserve the calculated exact size once. Whole-path and unreserved-only contexts are included. Focused native checks cover arbitrary bytes, strict error offsets, malformed-input atomicity, Writer output, exact allocation requests, and warm reuse. UriComponents additionally supplies tagged raw/encoded construction and selective serialization. Full workspace verification remains pending as recorded in the PR description.

